Issue #410, opened 2016-05-11
Copy not Working


BinAccel:
I tried to click copy to copy the sample data, but there is nothing on my clipboard. Using Firefox 46.0.1 on Windows 10 x64.

Xyene:
Thanks for reporting; it's good to have an issue about this.

The cause of the issue is the library we use for clipboard managing, which [has a number of open issues regarding Firefox usage](https://github.com/zeroclipboard/zeroclipboard/issues?utf8=✓&q=is%3Aissue+is%3Aopen+firefox). We might be able to switch to something like [clipboard.js](https://github.com/zenorocha/clipboard.js/) so long as it doesn't have issues as well.

Xyene:
@faiyaz26 has fixed this.

PR #419, opened 2016-07-02
fixed issue #410

faiyaz26:
Just fixed the issue, Tooltip will be done later, I have tried many ways to show tooltip but the style breaks so it needs time.


Xyene:
This looks good, thanks - I'll deploy it so we can get some more testing.

PR #489, opened 2016-10-12
fix _update_points

parkjs814:
Using `WHERE sub.problem_id = %s` and `GROUP BY sub.problem_id` together doesn't make sense.
Submissions should be grouped by profile id so that it can properly get maximum points for each user.
Also, `joining inside joining` may not be necessary.


quantum5:
There were more problems in that code that listed in this PR, but I think I've nailed them all in fec54203b8ed941f712d83ccced53f2e1b2399fb...30da9d8d16f043072d3028b8c1dbc3912e8370af. Good job on finding it though.

PR #511, opened 2016-12-23
Add PP support in the backend

WallE256:
This PR adds PP calculation and storage in the DB, only leaving the front-end part to be implemented.


quantum5:
I am sorry but you will have to redo the migration since I added another migration in the meantime. Hopefully all you have to do is renumber and change the previous migration name.

WallE256:
Yea, and took longer than expected

quantum5:
Actually, you might want to make your migration calculate performance points for everyone, instead of waiting for everyone to submit. Keyword: `RunPython`.

judge/models/profile.py
```diff
@@ -99,16 +100,20 @@ def organization(self):
...
-        problems = (self.submission_set.filter(points__gt=0, problem__is_public=True)
-                    .values('problem').distinct().count())
-        if self.points != points or problems != self.problem_count:
+        qdata = (Submission.objects.filter(user=self, points__isnull=False, problem__is_public=True)
+                 .values('problem_id').distinct().annotate(points=Max('points')).order_by('-points')
+                 .values_list('points', flat=True))
+        points = sum(qdata)
+        problems = len(qdata)
+        qdata = qdata[:min(len(qdata), len(table))]
+        table = table[:min(len(qdata), len(table))]
```
Xyene: This `min` statement can nicely be extracted to a variable, e.g. `entries`.

judge/views/user.py
```diff
@@ -103,7 +103,7 @@ def get_context_data(self, **kwargs):
         rating = self.object.ratings.order_by('-contest__end_time')[:1]
         context['rating'] = rating[0] if rating else None
         context['rank'] = Profile.objects.filter(points__gt=self.object.points).count() + 1
-        context['users'] = Profile.objects.filter(points__gt=0).count()
+        context['pp_rank'] = Profile.objects.filter(performance_points__gt=self.object.points).count() + 1
```
Xyene: I think `self.object.points` is the wrong object to use here.

PR #514, opened 2016-12-30
Make pp visible if using nightly

BinAccel:
Plz don't break stuff.

### DIFF
--- a/templates/user/user-base.jade
+++ b/templates/user/user-base.jade
@@ -49,7 +49,7 @@ block body
                 b {% trans "Total points:" %}
                 |  #{user.points}
                 br
-                .pp-feature(style="display: none")
+                .pp-feature(style=(None if perms.judge.test_site else 'display: none'))
                     br
                     b {% trans "Rank by PP:" %}
                     |  ##{pp_rank}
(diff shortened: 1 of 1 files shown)


templates/user/user-base.jade
```diff
@@ -49,7 +49,7 @@ block body
                 b {% trans "Total points:" %}
                 |  #{user.points}
                 br
-                .pp-feature(style="display: none")
+                .pp-feature(style=(None if perms.judge.test_site else 'display: none'))
```
quantum5: I do not think that's valid. These are dictionary lookups, and when PyJade handles them, you are to write `perms['judge']['test_site']`.

BinAccel:
@quantum5 fixed

templates/user/user-base.jade
```diff
@@ -49,7 +49,7 @@ block body
                 b {% trans "Total points:" %}
                 |  #{user.points}
                 br
-                .pp-feature(style=(None if perms.judge.test_site else 'display: none'))
+                .pp-feature(style=(None if perms['judge']['test_site else'] 'display: none'))
```
quantum5: I do not know why `else` is in the string.

BinAccel:
Take 3.

PR #525, opened 2017-02-03
Attempt to allow curators to edit submissions

BinAccel:
Someone check `queryset = queryset.filter(problem__authors__id=request.user.profile.id, problem__curators__id=request.user.profile.id)`. I'm not sure if that's valid.

Can someone also check that there's nothing else that needs to be changed?

### DIFF
--- a/judge/admin/submission.py
+++ b/judge/admin/submission.py
@@ -5,6 +5,7 @@ from django.conf.urls import url
 from django.contrib import admin, messages
 from django.core.cache import cache
 from django.core.exceptions import PermissionDenied
+from django.db.models import Q
 from django.http import HttpResponseRedirect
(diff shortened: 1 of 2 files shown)


judge/admin/submission.py
```diff
@@ -106,7 +106,8 @@ def get_queryset(self, request):
             'time', 'memory', 'points', 'status', 'result'
         )
         if not request.user.has_perm('judge.edit_all_problem'):
-            queryset = queryset.filter(problem__authors__id=request.user.profile.id)
+            id = request.user.profile.id
+            queryset = queryset.filter(Q(problem__authors__id=id) | Q(problem__curators__id=id))
```
quantum5: It helps if you import the `Q` object somewhere.

PR #527, opened 2017-02-05
Add new rank

BinAccel:
Colour and name are negotiable


Xyene:
And @FatalEagle said, "LET THERE BE RANKS!"

And there were ranks---

Issue #531, opened 2017-02-08
Add PP to user table


BinAccel:
Picking up where #511 left off. Will make PR soon:tm:.

Also, where should this be added? After the `points` column seems reasonable, but I am far from a reasonable person.

BinAccel:
Implemented, users can make visible with either a user script or by enabling experimental features

PR #532, opened 2017-02-08
Add PP to front end

BinAccel:
Implement #531 

I tested it on my own site so nothing should suddenly go up in flames. I think.


templates/user/users-table.jade
```diff
@@ -19,6 +19,15 @@ block after_rank
...
                 span.rating(class='{{ user.rating|rating_class }}') #{user.rating}
 
+block before_point_head
+    if sort_links
+       th.pp: a(href=sort_links['performance_points']) {% trans "PP" %}#{sort_order.performance_points}
+    else
+       th.pp {% trans "PP" %}
+
+block before_point
+    td #{user.performance_points|floatformat:2}
```
quantum5: You need a CSS class here.

BinAccel: Will do at home

BinAccel: Excuse the stupid question @quantum5, but why does L38 not need one?

quantum5:
Notice how there is a `.only('display_rank', 'user__username', 'name', 'points', 'rating')` in the query. You must add pp or it will run one query per row.

BinAccel:
Also going to make PP visible by default on the user's profile page

PR #538, opened 2017-02-13
Add no_points field

BinAccel:

PR #539, opened 2017-02-14
Implement unlisted users

kiritofeng:
I toasted #538 , so I'm gonna use this now.


judge/bridge/judgecallback.py
```diff
@@ -178,7 +178,7 @@ def on_grading_end(self, packet):
         submission.case_total = total
 
         sub_points = round(points / total * submission.problem.points if total > 0 else 0, 3)
-        if not submission.problem.partial and sub_points != submission.problem.points:
+        if submission.user.is_unlisted or (not submission.problem.partial and sub_points != submission.problem.points):
```
quantum5: Is that really the best way to do it?

judge/bridge/judgecallback.py
```diff
@@ -178,6 +178,8 @@ def on_grading_end(self, packet):
         submission.case_total = total
 
         sub_points = round(points / total * submission.problem.points if total > 0 else 0, 3)
+        if submission.user.is_unlisted:
```
quantum5: Not what I meant. I meant whether it's actually a good idea to do it here, basically giving a checkbox to make a users submission worth 0 point for the entire duration of the box being checked. Other methods may allow such users to regain points, or get no points for free by just checking the box, while this requires rescoring, which would not have worked anyways since rescore doesn't use this code.

Xyene: I agree with this. I think the best implementation would allow such users to gain points, but exclude them from all rankings. This is a bit tricky, since there are a lot of places that enumerate users. Off the top of my head;

* users list
* organization users list
* API
* contest ranking list

While these changes are being made, it would also be a good idea to filter out people who have `is_active` toggled off &mdash; we don't want banned users to appear in any form of rankings.

BinAccel:
Note: remove users from API, best submissions list, users table, contest rankings

judge/views/contests.py
```diff
@@ -149,7 +149,7 @@ def get_object(self, queryset=None):
         user = self.request.user
         profile = self.request.user.profile if user.is_authenticated else None
 
-        if (profile is not None and
+        if (profile is not None and not profile.is_unlisted and
```
quantum5: Unlisted users cannot exist contests that have been privated after their joined?

PR #547, opened 2017-02-19
Add solved state icon to problem pages

Phoenix1369:
Adding the problem solved icon


Xyene:
Basically this:

image

Issue #553, opened 2017-02-24
Some race condition will cause ping not to display properly on status page


guhenry3:
![Bug](https://i.imgur.com/6BvvXbR.png)

> nah, more like when it loads the first time there's a race
> but it's like
> such a low value fix
> because it only happens for the first ping
> so < 10s from when the judge connects

BinAccel:
This has been fixed.

PR #554, opened 2017-02-24
Show N/A if judge ping is None

BinAccel:
Address #553 .

### DIFF
--- a/templates/status/judge-status-table.jade
+++ b/templates/status/judge-status-table.jade
@@ -24,17 +24,17 @@ for judge in judges
             if judge.online
                 | #{judge.uptime|timedelta:"simple"}
             else
-                | N/A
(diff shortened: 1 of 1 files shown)


templates/status/judge-status-table.jade
```diff
@@ -24,17 +24,17 @@ for judge in judges
             if judge.online
                 | #{judge.uptime|timedelta:"simple"}
             else
-                | N/A
+                | {% trans "N/A" %}
         td.ping
-            if judge.online
+            if judge.online and judge.ping_ms
                 | #{judge.ping_ms|floatformat:3} ms
```
quantum5: Ideally this will also be translated, but we'll just say ms is the SI unit and must therefore be international.

Issue #562, opened 2017-02-26
Stats table does not add up


Xyene:
Since it does not list RTE. Perhaps RTE and IR should be grouped under "Execution Error" in the table.

I'm hesitant to call this a bug, but something should be done about it.

BinAccel:
Also missing from the table is:
- Internal Error
- Aborted
- Output Limit Exceeded

Should these be added as well?

BinAccel:
This has been addressed.

PR #563, opened 2017-02-26
Submission Table Fixes

BinAccel:
Address #562 .

Please **do not** merge until @Xyene or any one else addresses [this](https://github.com/DMOJ/site/issues/562#issuecomment-282566471).


judge/utils/problems.py
```diff
@@ -80,7 +80,7 @@ def get_result_table(*args, **kwargs):
             (_('Compile Error'), 'CE', results['CE']),
             (_('Time Limit Exceeded'), 'TLE', results['TLE']),
             (_('Memory Limit Exceeded'), 'MLE', results['MLE']),
-            (_('Invalid Return'), 'IR', results['IR']),
```
Xyene: Unless I'm wrong, that `'IR'` is used to determine a CSS class for coloring the table. You'll need to change that CSS class to "O".

PR #596, opened 2017-05-19
Fix mistune.escape_link() call to match the latest version of mistune

vdanciu:


quantum5:
I think this just breaks escaping of `"`  in the `href` attribute.

quantum5:
Never mind it seems `mistune` actually made links always escape quotes.

PR #609, opened 2017-08-29
fix bug: dates['min'] or dates['max'] may be None

wynterwang:
In `get_context_data` of `ContestCalendar`, year and month fields in result of `Contest.objects.aggregate` may be None.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -317,8 +317,12 @@ class ContestCalendar(TitleMixin, ContestListMixin, TemplateView):
             raise Http404()
 
         dates = Contest.objects.aggregate(min=Min('start_time'), max=Max('end_time'))
-        min_month = dates['min'].year, dates['min'].month
-        max_month = max((dates['max'].year, dates['max'].month), (self.today.year, self.today.month))
+        min_month = self.today.year, self.today.month
+        max_month = self.today.year, self.today.month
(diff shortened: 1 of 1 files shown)


judge/views/contests.py
```diff
@@ -317,8 +317,12 @@ def get_context_data(self, **kwargs):
             raise Http404()
 
         dates = Contest.objects.aggregate(min=Min('start_time'), max=Max('end_time'))
-        min_month = dates['min'].year, dates['min'].month
-        max_month = max((dates['max'].year, dates['max'].month), (self.today.year, self.today.month))
+        min_month = self.today.year, self.today.month
+        max_month = self.today.year, self.today.month
+        if dates['min'].year != None:
```
quantum5: I thought it's `dates['min']` that is supposed to be `None`? Wouldn't this be `TypeError: NoneType has no attribute "year"`?

Also, do not do `!= None`, use `is None` when checking for `None`.

judge/views/contests.py
```diff
@@ -317,8 +317,12 @@ def get_context_data(self, **kwargs):
...
+            min_month = dates['min'].year, dates['min'].month
+        if dates['max'].year != None:
```
quantum5: Same thing here.

quantum5:
Superseded by #686.

PR #610, opened 2017-09-16
Update api.py

airob0t:
Fix bug with non-ascii


quantum5:
Can you not use some commit from a year ago as the base?

PR #611, opened 2017-09-16
Fix bug with non-ascii

airob0t:
Fix bug with non-ascii


quantum5:
Can we not have the translation stuff in here?

quantum5:
@Xyene why was `str` used in the first place?

In any case, the fix is to drop `str` not to use `unicode`...

PR #651, opened 2017-12-23
Lower priority rejudges and judging priority system

quantum5:
Using a linked list as our submission queue. This allows submissions to be inserted into the queue, right before the priority marker for its priority. Higher number means lower priority.

Rejudges now have priority 1, compared to first submissions which have priority 0.

### DIFF
--- a/judge/admin/submission.py
+++ b/judge/admin/submission.py
@@ -138,7 +138,7 @@ class SubmissionAdmin(admin.ModelAdmin):
             queryset = queryset.filter(Q(problem__authors__id=id) | Q(problem__curators__id=id))
         judged = len(queryset)
         for model in queryset:
-            model.judge(was_rejudged=True)
+            model.judge(rejudged=True)
         self.message_user(request, ungettext('%d submission were successfully scheduled for rejudging.',
                                              '%d submissions were successfully scheduled for rejudging.',
                                              judged) % judged)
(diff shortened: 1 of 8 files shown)


judge/admin/submission.py
```diff
@@ -138,7 +138,7 @@ def judge(self, request, queryset):
             queryset = queryset.filter(Q(problem__authors__id=id) | Q(problem__curators__id=id))
         judged = len(queryset)
         for model in queryset:
-            model.judge(was_rejudged=True)
```
WallE256: `rejudge` instead of `rejudged`?

quantum5: Fixed.

quantum5:
Yes, the current priorities are temporary and more priorities will be added eventually. The priority system should be good enough to stay.

judge/judgeapi.py
```diff
@@ -39,11 +39,11 @@ def judge_request(packet, reply=True):
         return result
 
 
-def judge_submission(submission, **kwargs):
+def judge_submission(submission, rejudge):
```
kiritofeng: https://github.com/DMOJ/site/blob/low-priority-rejudge/judge/models/submission.py#L109 calls `judge_submission` with `self, rejudge, **kwargs`

PR #660, opened 2018-01-02
Remove horizontal scrollbar on ranking page

Injust:


Xyene:
I think there was some reason that was `scroll` and not `auto` but I can't remember why and it seems to be fine now anyways... so this looks good to merge, thanks.

PR #661, opened 2018-01-02
Improve viewport usage for larger resolutions

Injust:


Xyene:
This looks good, but could you describe what this affects, usability-wise (screenshots?)

I don't think I have an appropriate device or setup to really see what this changes.

quantum5:
Most pages on DMOJ look awkward on large screens without there being a width cap, since then the content expands indefinitely, to bad results. This is why those `max-width`s were added originally.

Injust:
@Xyene The change lets the page content fill (slightly) more of the viewport, rather than leaving much of the sides blank. Screenshots at https://imgur.com/a/LTxxg.
I think the most important change is to the contest rankings table, which can now display all data up to 8 problems, up from the previous (barely) 6.

@quantum5 The `max-width` on `body` prevents that from occurring, even when the others are removed.

Xyene:
This looks good to me. The contest ranking could also be helped further by changing the minimum width of the user column to ~20em, since there's less data displayed there since the removal of display names.

PR #664, opened 2018-01-15
Don't show editorial link for private editorials

kiritofeng:
Renamed `has_editorial` to `has_public_editorial`, which is more accurate, and implemented the functionality. @quantum5 @Xyene please check the efficiency of the queries.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -12,7 +12,7 @@ from django.core.cache import cache
 from django.core.exceptions import ObjectDoesNotExist, ImproperlyConfigured
 from django.core.urlresolvers import reverse
 from django.db import connection, IntegrityError
-from django.db.models import Q, Min, Max, Count
+from django.db.models import Q, Min, Max, Count, Sum, Case, When, IntegerField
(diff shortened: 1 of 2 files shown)


judge/views/contests.py
```diff
@@ -199,7 +199,8 @@ def get_context_data(self, **kwargs):
         context = super(ContestDetail, self).get_context_data(**kwargs)
         context['contest_problems'] = Problem.objects.filter(contests__contest=self.object) \
             .order_by('contests__order').defer('description') \
-            .annotate(has_editorial=Count('solution')) \
+            .annotate(has_public_editorial=Sum(Case(When(solution__is_public=True, then=1), \
```
quantum5: The trailing backslash on this line is unnecessary.

judge/views/contests.py
```diff
@@ -199,7 +199,8 @@ def get_context_data(self, **kwargs):
...
+                                             default=0, output_field=IntegerField()))) \
```
quantum5: The start of this line is not aligned with the bracket after `Case`.

PR #674, opened 2018-03-13
Add Language for Korean

joonas-yoon:
I am trying to translate them.

I think I did about 60~70%, but need to fill out more.

want your help :)


quantum5:
I already updated this myself a while ago.

PR #698, opened 2018-09-12
Implement contest time penalty support

Xyene:
WIP


Xyene:
This is actually a pretty big PR as-is. The current implementation works for all but rated contests, which is still a sizable use-case, so I'm considering splitting the two goals up as separate PRs (so that we can make use of unrated time penalty contests in the interim).

judge/views/contests.py
```diff
@@ -446,39 +450,59 @@ def best_solution_state(points, total):
...
-            if (part, prob) in data and data[part, prob][1] is not None else None
-            for prob, points, is_pretested in problems])
+        contest_progress = []
+        for prob, points, is_pretested in problems:
+            if (part, prob) in data and data[part, prob][1] is not None:
+                best_points = data[part, prob][1]
+                last_attempt_time = data[part, prob][2] - participation.start
+                time_penalty = timedelta(microseconds=data[part, prob][3])
+
+                sol_data =  BestSolutionData(code=data[part, prob][0],
```
quantum5: Two spaces.

quantum5: Can we also put all the contents on a new line so they are only indented with 4 spaces?

judge/views/contests.py
```diff
@@ -446,39 +450,59 @@ def best_solution_state(points, total):
...
+                sol_data = None
+
+            contest_progress.append(sol_data)
+
+        return make_contest_ranking_profile(participation, contest_progress)
 
-    return map(make_ranking_profile, queryset.select_related('user__user', 'rating')
-               .defer('user__about', 'user__organizations__about'))
+    rankings = map(make_ranking_profile, queryset.select_related('user__user', 'rating')
+                                                  .defer('user__about', 'user__organizations__about'))
```
quantum5: . is not aligned

judge/migrations/0069_time_penalty.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
```
quantum5: You should probably rebase this migration.

Xyene:
This is no longer current, and cannot be salvaged. The ECOO branch contains a somewhat working implementation of this.

Issue #696, opened 2018-09-12
Extra judge controls in admin


Xyene:
There are occasionally misbehaving judges that need to be dealt with at inconvenient times. SSHing into a particular judge from e.g. a phone to turn it off is not ideal, and it would be straightforward to add some basic level of control in the judge admin.

In particular, I'm referring to:

* A button to send a new `disconnect` packet to the judge, which performs graceful cleanup
* A button to forcefully drop the connection
* A checkbox to block authentication of the judge, even if the key is correct (potentially useful with connection dropping)

quantum5:
Implemented.

PR #699, opened 2018-09-12
Add better controls for judge connections

Xyene:
Addresses #696.


judge/admin/runtime.py
```diff
@@ -77,14 +81,29 @@ class JudgeAdmin(VersionAdmin):
...
         (_('Description'), {'fields': ('description',)}),
         (_('Information'), {'fields': ('created', 'online', 'last_ip', 'start_time', 'ping', 'load')}),
         (_('Capabilities'), {'fields': ('runtimes', 'problems')}),
     )
     list_display = ('name', 'online', 'start_time', 'ping', 'load', 'last_ip')
     ordering = ['-online', 'name']
 
+    def get_urls(self):
+        return [url(r'^(\d+)/disconnect/$', self.disconnect_view, name='judge_judge_disconnect'),
+                url(r'^(\d+)/terminate/$', self.terminate_view, name='judge_judge_terminate')] + super(JudgeAdmin, self).get_urls()
```
quantum5: Line way too long.

judge/bridge/judgehandler.py
```diff
@@ -108,6 +108,13 @@ def working(self):
     def get_related_submission_data(self, submission):
         return 2, 16384, False, False
 
+    def disconnect(self, force=False):
+        if not force:
```
quantum5: `if not` and then `else` is slightly confusing.

judge/bridge/judgelist.py
```diff
@@ -47,6 +47,12 @@ def register(self, judge):
             self.judges.add(judge)
             self._handle_free_judge(judge)
 
+    def disconnect(self, judge_id, force=False):
+        for judge in self.judges:
+            if judge.name == judge_id:
+                judge.disconnect(force=force)
+                return
```
quantum5: Sometimes, a judge might end up connecting multiple times. I think killing that brain damage would be nice to have, so maybe we shouldn't `return`.

judge/models/runtime.py
```diff
@@ -123,6 +128,12 @@ class Judge(models.Model):
     def __unicode__(self):
         return self.name
 
+    def disconnect(self, force=False):
+        if self.online:
```
quantum5: A judge might be online in a phantom state. This is known to occur when a judge disconnects due to network failure, and then the judge reinstates the connection. However, the bridge realizes the judge disconnected afterwards, and the judge becomes marked as offline, despite having another connection. We need to significantly rework the bridge to fix this issue, but in the meantime... we probably shouldn't check `self.online`.

Xyene:
cc @quantum5

PR #708, opened 2018-09-26
Hide private contest comments from front page

Xyene:


Xyene:
LGTM, but I'd like to see it actually running before I merge.

judge/models/contest.py
```diff
@@ -162,6 +162,22 @@ def show_scoreboard(self):
             return False
         return True
 
+    def is_accessible_by(self, user):
+        # All users can see 
+        if self.is_public:
```
quantum5: No. An organization-private contest needs to have `is_public` set to have it published. A contest with `is_public = False` is only accessible by admins and organizers.

A organization-private contest will additionally set `is_private = True`. The goal is to prevent comments on those from being seen by people outside of the organization.

Issue #684, opened 2018-07-06
Allow problems to be visible only to certain organizations


yhylord:
Currently, the only way to do this is to set those problems to hidden, create contests that contain those problems and make contests private to certain organizations. This is a workaround that's kind of troublesome and abusive of private contests. Since we already have problem groups, maybe we can make some problem groups visible only to some organizations?

Xyene:
This has been implemented by @kiritofeng 👍

PR #712, opened 2018-09-27
Implement Organization-Private Problems

kiritofeng:
Addresses #684.


judge/models/problem.py
```diff
@@ -166,9 +181,6 @@ def is_accessible_by(self, user):
         if user.has_perm('judge.edit_own_problem') and self.is_editor(user.profile):
             return True
 
-        if not user.is_authenticated:
```
quantum5: What if it's not public? Then it doesn't run after you move it up.

judge/models/problem.py
```diff
@@ -299,6 +311,7 @@ class Meta:
             ('clone_problem', 'Clone problem'),
             ('change_public_visibility', 'Change is_public field'),
             ('change_manually_managed', 'Change is_manually_managed field'),
+            ('see_organization_problem', 'See organization-private problems'),
```
quantum5: Your migration does not create the permission.

judge/models/problem.py
```diff
@@ -154,9 +158,20 @@ def is_editable_by(self, user):
...
+                return True
+
+            # A non-authenticated user will fail all of the following conditions.
+            # Since several of them need the user to be authenticated, we check now.
+            if not user.is_authenticated:
+                return False
+
+            # User is in the organization
+            if self.organizations.filter(id__in=user.profile.organizations.all()):
+                return True
```
quantum5: Where do you check `see_organization_problem`?

judge/models/problem.py
quantum5: Your migration still doesn't create the permission.

Issue #704, opened 2018-09-25
Use numerical IDs for organization pages


Xyene:
Having to think of a 6-character short ID that never really matches is suboptimal. IDs are far cleaner in the URL, and we could maintain a simple redirect map for the organizations we already have.

quantum5:
How about we do `id-slug` like most websites do? So like `1-don-mills-collegiate-institute`

Xyene:
I'm fine with that approach too. Just not with the current one.

quantum5:
Implemented.

PR #713, opened 2018-09-28
Use id and slug in organization links

quantum5:
Switch organization links from `/organization/<6 character key>/...` to `/organization/<id>-<slug>/...`, for example, `/organization/1-don-mills-collegiate-institute`.

This will complete #704.


judge/models/profile.py
```diff
@@ -18,9 +18,11 @@
 
 
 class Organization(models.Model):
-    name = models.CharField(max_length=50, verbose_name=_('organization title'))
+    name = models.CharField(max_length=128, verbose_name=_('organization title'))
+    slug = models.SlugField(max_length=128, verbose_name=_('organization slug'),
+                            help_text=_('Organization name shows in URL'))
```
Xyene: *shown

judge/views/organization.py
```diff
@@ -312,3 +312,8 @@ def post(self, request, *args, **kwargs):
 
         organization.members.remove(user)
         return HttpResponseRedirect(organization.get_users_url())
+
+
```
Xyene: I think this deserves some explanation as to why we have this, before we forget about this migration in years to come.

judge/views/organization.py
```diff
@@ -10,7 +10,7 @@
 from django.db import transaction
 from django.db.models import Count, Max
 from django.forms import Form, modelformset_factory
-from django.http import HttpResponseRedirect, Http404
+from django.http import HttpResponseRedirect, Http404, HttpResponsePermanentRedirect
```
Xyene: Import order

Issue #642, opened 2017-12-10
edit/profile not working 


thangdnsf:
![image](https://user-images.githubusercontent.com/7885424/33802235-fc3e58f6-dda4-11e7-8044-5563c3e734ac.png)

monouno:
I guess you didn't configure the math engine server, so the submitted form data lack of 'math_engine' causes it to not work.
Annotate this line of code should be ok.
https://github.com/DMOJ/site/blob/d0648dbc791dd99ab1874b4f20f08bc915d248e9/templates/user/edit-profile.html#L115

yhylord:
Hey I got the same problem, and commenting out `{% if has_math_config %}` doesn't work. Any help?

yhylord:
OK found out that I didn't actually comment it out. For anyone encountering this problem, make sure to use `{# COMMENT #}` in jinja2 rather than `# COMMENT`.

Xyene:
@kiritofeng has fixed this.

PR #714, opened 2018-09-28
Remove `math_engine` from `ProfileForm` if there is no `MATHOID_URL`.

kiritofeng:
Addresses #642.

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -27,15 +27,19 @@ class ProfileForm(ModelForm):
 
     class Meta:
         model = Profile
-        fields = ['name', 'about', 'organizations', 'timezone', 'language', 'ace_theme', 'user_script', 'math_engine']
+        fields = ['name', 'about', 'organizations', 'timezone', 'language', 'ace_theme', 'user_script']
         widgets = {
             'name': TextInput(attrs={'style': 'width:100%;box-sizing:border-box'}),
             'user_script': AceWidget(theme='github'),
             'timezone': Select2Widget(attrs={'style': 'width:200px'}),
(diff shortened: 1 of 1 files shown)


judge/forms.py
```diff
@@ -27,16 +27,20 @@ class ProfileForm(ModelForm):
...
-        fields = ['name', 'about', 'organizations', 'timezone', 'language', 'ace_theme', 'user_script', 'math_engine']
+        fields = ['name', 'about', 'organizations', 'timezone', 'language', 'ace_theme', 'user_script']
         widgets = {
             'name': TextInput(attrs={'style': 'width:100%;box-sizing:border-box'}),
             'user_script': AceWidget(theme='github'),
             'timezone': Select2Widget(attrs={'style': 'width:200px'}),
             'language': Select2Widget(attrs={'style': 'width:200px'}),
-            'ace_theme': Select2Widget(attrs={'style': 'width:200px'}),
-            'math_engine': Select2Widget(attrs={'style': 'width:200px'})
+            'ace_theme': Select2Widget(attrs={'style': 'width:200px'})
```
quantum5: Leave the trailing comma. It makes diffs less dumb.

judge/forms.py
```diff
@@ -27,16 +27,20 @@ class ProfileForm(ModelForm):
...
         }
 
+        has_math_config = bool(getattr(settings, 'MATHOID_URL', False))
+        if has_math_config:
+            fields.append('math_engine')
+            widgets.update({'math_engine': Select2Widget(attrs={'style': 'width:200px'})})
```
quantum5: Why can't you just do `widgets['math_engine'] = ...`?

PR #711, opened 2018-09-27
Add organization-private and non-public contest tags

kiritofeng:
Address #703.

Issue #703, opened 2018-09-25
Add badge for organization-private contests


Xyene:
...and merge it with the current lock icon.

![image](https://user-images.githubusercontent.com/1403503/45988371-9252a900-c044-11e8-86f4-a5b15729128f.png)

Xyene:
Let's chump in a badge for private contests in here too. Probably black background with white text would work well.

quantum5:
Implemented.

PR #716, opened 2018-09-28
Patch #711

kiritofeng:
Change the colour of organization tags, and link them to their respective organization.

Change the tag for private contests to lowercase.

This patch fixes the implementation of #703.


templates/contest/list.html
```diff
@@ -21,7 +21,9 @@
                 count_down($(this));
             });
 
-            $('.contest-tag').find('a').featherlight();
+            if($('.contest-tag').find('a').hasOwnProperty('data-featherlight')) {
```
Xyene: I feel like `.find('a[data-featherlight]')` or similar should be sufficient here.

templates/contest/list.html
```diff
@@ -63,8 +65,11 @@
             {% endif %}
             {% if contest.is_private %}
                 {% for org in contest.organizations.all() %}
-                    <span style="background-color: #afafaf; color: #000000;" class="contest-tag">
-                        <i class="fa fa-lock"></i> {{ org.name }}
+                    <span style="background-color: #cccccc;" class="contest-tag">
+                        <a href="{{ org.get_absolute_url() }}"
+                           style="color: #000000">
```
Xyene: Put this on the previous line.

templates/contest/list.html
```diff
@@ -63,8 +65,11 @@
...
+                            <i class="fa fa-lock"></i>{{ org.name }}
```
Xyene: You removed the space before the organization name that used to be here.

templates/contest/list.html
```diff
@@ -21,7 +21,9 @@
                 count_down($(this));
             });
 
-            $('.contest-tag').find('a').featherlight();
+            if($('.contest-tag').find('a').hasOwnProperty('data-featherlight')) {
+                $('.contest-tag').find('a').featherlight();
```
Xyene: I don't think this makes sense, just use the above.

templates/contest/list.html
```diff
@@ -21,7 +21,7 @@
                 count_down($(this));
             });
 
-            $('.contest-tag').find('a').featherlight();
+            $('.contest-tag').find('a[data-featherlight]')
```
quantum5: What in the name of fuck is this?

templates/contest/list.html
quantum5: It's also hilarious how wrong that code is.

templates/contest/list.html
```diff
@@ -63,8 +63,10 @@
             {% endif %}
             {% if contest.is_private %}
                 {% for org in contest.organizations.all() %}
-                    <span style="background-color: #afafaf; color: #000000;" class="contest-tag">
-                        <i class="fa fa-lock"></i> {{ org.name }}
+                    <span style="background-color: #cccccc;" class="contest-tag">
+                        <a href="{{ org.get_absolute_url() }}" style="color: #000000">
```
Xyene: Consistency; added `;` for the `background-color` above but not here.

Issue #560, opened 2017-02-26
Redesign registration email


Xyene:
Needs a whole lotta love.

![image](https://cloud.githubusercontent.com/assets/1403503/23337143/60a54276-fbb2-11e6-9754-7653d3029e46.png)

Xyene:
The TLE email would be a good start.

BinAccel:
Why does it say `on the !`

Xyene:
_Now_ you're asking the right questions.

Xyene:
This is embarrassingly bad. Even a plaintext email would be better, honestly.

PR #719, opened 2018-09-30
Use a simpler activation email template

Xyene:
This is superior to the horrifyingly bad HTML/CSS one currently in use, that doesn't even display the site name correctly. Addresses #560.

### DIFF
--- a/templates/registration/activation_email.html
+++ b/templates/registration/activation_email.html
@@ -1,24 +1,23 @@
-<div style="display:block;margin: 0 2px;padding: 1em;border: 3px solid #FFDE05;background-color: #000;border-radius: 6px;font-size: .95em;color: #444;margin-bottom:0.75em">
-<div style="display:table;margin-left:2em"><img src="https://avatars1.githubusercontent.com/u/6934864?v=3&s=101" style="display:inline;vertical-align: middle">
-<h1 style="font-size:4em;display:inline;vertical-align: middle"><a href="//{{ site.domain }}" style="text-decoration:none;color:gray"><span style="color: #FFDE05">DM::</span>OJ</a>
-      </h1>
-</div>
-</div>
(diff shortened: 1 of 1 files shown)


templates/registration/activation_email.html
```diff
@@ -1,24 +1,23 @@
-<div style="display:block;margin: 0 2px;padding: 1em;border: 3px solid #FFDE05;background-color: #000;border-radius: 6px;font-size: .95em;color: #444;margin-bottom:0.75em">
-<div style="display:table;margin-left:2em"><img src="https://avatars1.githubusercontent.com/u/6934864?v=3&s=101" style="display:inline;vertical-align: middle">
-<h1 style="font-size:4em;display:inline;vertical-align: middle"><a href="//{{ site.domain }}" style="text-decoration:none;color:gray"><span style="color: #FFDE05">DM::</span>OJ</a>
-      </h1>
-</div>
-</div>
-<div style="display:block;margin: 0 2px;padding: 1em;border: 3px solid #2980B9;background-color: #f8f8f8;border-radius: 6px;font-size: .95em;color: #444;">
+<b>Thanks for registering on the {{ site.name }}! We're glad to have you.</b>
+<br><br>
+The last step is activating your account. Please activate your {{ SITE_NAME }} account in the next {{ expiration_days }} days.a
```
quantum5: I highly doubt the trailing `a` is intended.

templates/registration/activation_email.html
```diff
@@ -1,24 +1,23 @@
...
+<br><br>
+Please click on the following link to activate your account:
+<p style="margin-left:1em">
+      <a href="http://{{ site.domain }}/accounts/activate/{{ activation_key }}/">http://{{ site.domain }}/accounts/activate/{{ activation_key }}</a>
```
quantum5: We cannot use `http`. Pass the scheme as a variable. We should have it somewhere.

Issue #730, opened 2018-09-30
Don't show authors in editorials when there are none


quantum5:
![default](https://user-images.githubusercontent.com/461885/46262302-83994580-c4cd-11e8-8333-9f2095a100b8.png)

PR #732, opened 2018-09-30
Hide authors when there are none

Ninjaclasher:
Addresses #730.

### DIFF
--- a/templates/problem/editorial.html
+++ b/templates/problem/editorial.html
@@ -19,8 +19,11 @@
 {% block body %}
     <div class="content-description">
         {% if not has_solved_problem %}
-            <div class="alert alert-danger">{{ _('Remember to use this editorial <b>only</b> when stuck, and <b>not to copy-paste code from it</b>. Please be respectful to the problem author and editorialist. <br><br> <b>Submitting an official solution before solving the problem yourself is a bannable offence.</b>') }}</div>{% endif %}
-        <p>Authors: {{ link_users(solution.authors.all()) }}</p>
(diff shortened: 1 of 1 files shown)


templates/problem/editorial.html
```diff
@@ -19,8 +19,11 @@
 {% block body %}
     <div class="content-description">
         {% if not has_solved_problem %}
-            <div class="alert alert-danger">{{ _('Remember to use this editorial <b>only</b> when stuck, and <b>not to copy-paste code from it</b>. Please be respectful to the problem author and editorialist. <br><br> <b>Submitting an official solution before solving the problem yourself is a bannable offence.</b>') }}</div>{% endif %}
-        <p>Authors: {{ link_users(solution.authors.all()) }}</p>
+            <div class="alert alert-danger">{{ _('Remember to use this editorial <b>only</b> when stuck, and <b>not to copy-paste code from it</b>. Please be respectful to the problem author and editorialist. <br><br> <b>Submitting an official solution before solving the problem yourself is a bannable offence.</b>') }}</div>
+        {% endif %}
+        {% if solution.authors.exists() %}
+            <p>Authors: {{ link_users(solution.authors.all()) }}</p>
```
quantum5: Solution inefficient. This runs two queries. You should just do a with on the `all` queryset, and use `if` on that.

templates/problem/editorial.html
```diff
@@ -21,7 +21,7 @@
         {% if not has_solved_problem %}
             <div class="alert alert-danger">{{ _('Remember to use this editorial <b>only</b> when stuck, and <b>not to copy-paste code from it</b>. Please be respectful to the problem author and editorialist. <br><br> <b>Submitting an official solution before solving the problem yourself is a bannable offence.</b>') }}</div>
         {% endif %}
-        {% if solution.authors.exists() %}
+        {% if solution.authors.all() %}
```
Xyene: What @quantum5 was referring to is something like https://github.com/DMOJ/site/blob/88c1aa319250444230589ade04aef2e8bd68a6b9/templates/problem/problem.html#L175

As-is, this still performs 2 queries.

Issue #726, opened 2018-09-30
Completely remove user display names


quantum5:

quantum5:
I think we have completely wiped it out...

PR #734, opened 2018-10-02
Remove display names from Profile model

Xyene:
Addresses #726.

WIP


judge/models/profile.py
```diff
@@ -125,16 +124,10 @@ def calculate_points(self, table=(lambda x: [pow(x, i) for i in xrange(100)])(ge
 
     @cached_property
     def display_name(self):
-        if self.name:
-            return self.name
         return self.user.username
 
     @cached_property
     def long_display_name(self):
```
quantum5: Hopefully we can get rid of this too.

judge/models/profile.py
```diff
@@ -125,16 +124,10 @@ def calculate_points(self, table=(lambda x: [pow(x, i) for i in xrange(100)])(ge
 
     @cached_property
     def display_name(self):
```
quantum5: Hopefully we can get rid of this too.

judge/models/profile.py
Xyene: > WIP

quantum5: Of course, but this should be part of the work that is currently in progress.

Issue #733, opened 2018-10-01
List "Active contests" on contests page


Xyene:
This would be a list of all the contests the user has joined that are still ongoing (i.e. their participation is not expired). The section should be hidden if there are no such contests.

kiritofeng:
This code is now running in production.

PR #736, opened 2018-10-04
List active contests on contests page

Ninjaclasher:
Addresses #733.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -76,7 +76,7 @@ class ContestList(TitleMixin, ContestListMixin, ListView):
     def get_context_data(self, **kwargs):
         context = super(ContestList, self).get_context_data(**kwargs)
         now = timezone.now()
-        past, present, future = [], [], []
+        past, present, active, future = [], [], [], [] 
         for contest in self.get_queryset():
             if contest.end_time < now:
                 past.append(contest)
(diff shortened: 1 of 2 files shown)


judge/views/contests.py
```diff
@@ -76,17 +76,29 @@ def get_queryset(self):
...
                 past.append(contest)
             elif contest.start_time > now:
                 future.append(contest)
             else:
                 present.append(contest)
+        if self.request.user.is_authenticated:
+            participations = ContestParticipation.objects.filter(user=self.request.user.profile, virtual=0)
+            inactive = []
+            for contest in present:
+                participation = participations.filter(contest_id=contest.id).first()
```
Xyene: This will perform a query for every contest in `present`. A better idea here is to collect all `present` `id`s into a set, and then use the `contest_id__in` argument.

judge/views/contests.py
```diff
@@ -76,17 +76,29 @@ def get_queryset(self):
...
         for contest in self.get_queryset():
             if contest.end_time < now:
                 past.append(contest)
             elif contest.start_time > now:
                 future.append(contest)
             else:
                 present.append(contest)
+        if self.request.user.is_authenticated:
+            participations = ContestParticipation.objects.filter(user=self.request.user.profile, virtual=0)
+            inactive = []
```
Xyene: Why have this as a separate variable if it immediately replaces `present` on L96?

judge/views/contests.py
```diff
@@ -76,17 +76,22 @@ def get_queryset(self):
     def get_context_data(self, **kwargs):
         context = super(ContestList, self).get_context_data(**kwargs)
         now = timezone.now()
-        past, present, future = [], [], []
+        past, present, active, future = [], set(), [], []
```
quantum5: Did you just make present contests ordered arbitrarily?

judge/views/contests.py
```diff
@@ -76,17 +76,22 @@ def get_queryset(self):
...
             if contest.end_time < now:
                 past.append(contest)
             elif contest.start_time > now:
                 future.append(contest)
             else:
-                present.append(contest)
+                present.add(contest)
+        if self.request.user.is_authenticated:
+            active = filter(lambda x: not x.ended, ContestParticipation.objects.filter(contest__in=present, virtual=0))
+            present -= set(map(lambda x: x.contest, active))
```
quantum5: This runs one query per active contest to fetch a full contest object, as well as a query to fetch the participation that you are ignoring. You should just fetch the contest ids and discard contests based on the ids.

Ninjaclasher:
I lost access to the branch when I accidentally deleted my own fork, so there is no way for me to finish this PR. Quick search around and it appears there is no way to reclaim this PR (https://stackoverflow.com/questions/36071272/fix-unknown-repository-of-an-opened-pr-after-deleted-the-fork). Closing it for now.

PR #742, opened 2018-10-12
Add Visual Studio Code folder to gitignore

WallE256:
This is useful for developing in Visual Studio Code.


quantum5:
Honestly I feel like rewriting the entire file to use a standard template, but this is fine for now.

Issue #715, opened 2018-09-28
Lock contest comments


Xyene:
This is a special case of #555 for contests.

Contest comments provide an easy forum to discuss answers during contests, which some organizers have expressed interest in disallowing. Straightforward implementation through an `has_comments_enabled` `BooleanField` defaulted to true.

This doesn't need any fancy implementation where if there are already comments when the field is toggled, only those should be shown, etc. Just remove the entire comment section if this field is unset, and perhaps disallow the programmatic creation of comments too (less important).

Issue #555, opened 2017-02-25
Restrict pages from receiving comments


WallE256:
Due to [recent events](https://dmoj.ca/contest/moi17r1), this is feature is required to stop the spam that might happen on certain subjects. It would lock the comment section, not allowing the regular user to post comments.

PR #746, opened 2018-10-13
Implement Comment Locks

WallE256:
Addresses #555 and #715.


judge/comments.py
```diff
@@ -65,6 +65,10 @@ def post(self, request, *args, **kwargs):
         self.object = self.get_object()
         page = self.get_comment_page()
 
+        if (CommentLock.objects.filter(page=self.get_comment_page()).exists() and
+            not (self.request.user.is_superuser or self.request.user.has_perm('judge.override_comment_lock'))):
```
quantum5: Should probably move the condition to a method called `is_comment_locked` or something similar, since you duplicate it again in `get_context_data`.

judge/models/__init__.py
```diff
@@ -25,4 +25,5 @@
 revisions.register(Judge, fields=['name', 'created', 'auth_key', 'description'])
 revisions.register(Language)
 revisions.register(Comment, fields=['author', 'time', 'page', 'score', 'title', 'body', 'hidden', 'parent'])
+revisions.register(CommentLock)
```
quantum5: Do we really need to track CommentLock versions? We probably shouldn't.

judge/admin/__init__.py
```diff
@@ -12,7 +12,7 @@
 from judge.admin.ticket import TicketAdmin
 from judge.models import Language, Profile, Problem, ProblemGroup, ProblemType, Submission, Comment, \
     MiscConfig, Judge, NavigationBar, Contest, ContestParticipation, Organization, BlogPost, \
-    License, OrganizationRequest, ContestTag, Ticket
+    License, OrganizationRequest, ContestTag, Ticket, CommentLock
```
Xyene: Import order

judge/comments.py
```diff
@@ -6,7 +6,7 @@
 from django.db.models.expressions import Value, F
 from django.db.models.functions import Coalesce
 from django.forms import ModelForm
-from django.http import HttpResponseRedirect
+from django.http import HttpResponseRedirect, HttpResponseForbidden
```
Xyene: Import order

judge/models/comment.py
```diff
@@ -151,3 +151,16 @@ class Meta:
         unique_together = ['voter', 'comment']
         verbose_name = _('comment vote')
         verbose_name_plural = _('comment votes')
+
+class CommentLock(models.Model):
+    page = models.CharField(max_length=30, verbose_name=_('associated Page'), db_index=True,
+                            validators=[RegexValidator('^[pc]:[a-z0-9]+$|^b:\d+$|^s:',
+                                                       _('Page code must be ^[pc]:[a-z0-9]+$|^b:\d+$'))])
```
Xyene: I suppose we might want this for editorial/blog comments too.

judge/models/comment.py
```diff
@@ -151,3 +151,16 @@ class Meta:
         unique_together = ['voter', 'comment']
         verbose_name = _('comment vote')
         verbose_name_plural = _('comment votes')
+
+class CommentLock(models.Model):
+    page = models.CharField(max_length=30, verbose_name=_('associated Page'), db_index=True,
```
Xyene: `verbose_name` should be all lowercase, I think.

judge/comments.py
```diff
@@ -98,6 +105,7 @@ def get_context_data(self, **kwargs):
             unique_together_left_join(queryset, CommentVote, 'comment', 'voter', profile.id)
             context['is_new_user'] = (not self.request.user.is_staff and
                                       not profile.submission_set.filter(points=F('problem__points')).exists())
+            context['comment_lock'] = self.is_comment_locked()
```
Xyene: As I understand it this would cause the UI to show comments as unlocked for users that are not logged in. It wouldn't be an issue since they wouldn't be able to post comments regardless, but I think it would be better to be consistent and tell them that comments are locked anyway.

templates/comments/list.html
```diff
@@ -60,7 +60,7 @@ <h2><i style="padding-right: 0.3em" class="fa fa-comments"></i>{{ _('Comments')
                                     <a href="#comment-{{ node.id }}" title="{{ _('Link') }}" class="comment-link">
                                     <i class="fa fa-link fa-fw"></i>
                                 </a>
-                                    {% if logged_in %}
+                                    {% if logged_in and not comment_lock %}
```
Xyene: Perhaps the list should display a message to the effect of "Comments for this page are disabled." rather than just removing the comment list, so prevent users from thinking their absence is a bug.

judge/models/comment.py
quantum5: It does have that though. Notice the ^b and ^s.

Xyene: Oh yikes

judge/models/comment.py
WallE256: I will change this along with #747 for the sake of both models having the same definition in a separate PR.

templates/comments/list.html
```diff
@@ -141,4 +141,10 @@ <h3>{{ _('New comment') }}</h3>
             {% endif %}
         </div>
     {% endif %}
+
+    {% if comment_lock %}
+        <div style="width: 100%; padding-left: 14px" class="alert alert-warning">
+            {{ _('Comments are disabled on this page.') }}
```
Xyene: I believe if there are no comments and comments are locked, the page will display:

> There are no comments at the moment.

> Comments are disabled on this page.

Which feels fairly redundant to me. The former message should not be shown if the latter is.

templates/comments/list.html
```diff
@@ -141,4 +141,10 @@ <h3>{{ _('New comment') }}</h3>
             {% endif %}
         </div>
     {% endif %}
+
+    {% if comment_lock %}
+        <div style="width: 100%; padding-left: 14px" class="alert alert-warning">
```
Xyene: I would rather not inline this style, we have a comment SCSS for things like this.

Issue #747, opened 2018-10-14
Fix comment regex for editorials


quantum5:
Currently, the regex used for comment pages is `^[pc]:[a-z0-9]+$|^b:\d+$|^s:`, where `^s:` does not care of the solution matches the proper format. The code that follows `s:` is the same as the problem code. The regex should be changed so that it validates the problem code format after `s:`.

PR #752, opened 2018-10-18
Fix the comment regex

WallE256:
Addresses #747.


judge/models/comment.py
```diff
@@ -19,6 +19,8 @@
 
 __all__ = ['Comment', 'CommentLock', 'CommentVote']
 
+comment_validator = RegexValidator('^[pcs]:[a-z0-9]+$|^b:\d+$', 
+                                          _('Page code must be ^[pcs]:[a-z0-9]+$|^b:\d+$'))
```
Xyene: Weird indent

Issue #757, opened 2018-10-20
Checking passwords against Troy Hunt's database


quantum5:
Due to recent events involving password reuse in different sites resulting in account compromise, we should check users for compromised passwords.

We can use the [pwned passwords API](https://haveibeenpwned.com/API/v2#PwnedPasswords) to check passwords before we allow them to be used on registration/password change/password reset. There probably is a library on PyPI that does this that probably also integrates with Django. We should find it, and if it is good, use it.

PR #758, opened 2018-10-20
Check new passwords with pwned passwords API (#757)

PuffyShoggoth:


judge/utils/pwned.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+import hashlib
+import logging
+import sys
+
+import requests
+from django.conf import settings
+from django.utils.six import text_type
+
```
quantum5: Remove this new line. Also, imports should be grouped as follows:

* standard libraries
* third party libraries
* own code

with a new line in between each group.

Within each group, the imports should be sorted.

So, the second part should look like

```python
from django.contrib.auth.password_validation import CommonPasswordValidator
from django.core.exceptions import ValidationError
from django.utils.six import string_types, text_type
from django.utils.translation import ugettext as _, ungettext
import requests
```

judge/utils/pwned.py
```diff
@@ -0,0 +1,145 @@
...
+        raise TypeError('Password values to check must be Unicode strings.')
+    password_hash = hashlib.sha1(password.encode('utf-8')).hexdigest().upper()
+    prefix, suffix = password_hash[:5], password_hash[5:]
+    results = _get_pwned(prefix)
+    if results is None:
+        # Gracefully handle timeouts and HTTP error response codes.
+        return None
+    return results.get(suffix, 0)
+
+
```
quantum5: Too many new lines. Two new line between global functions/classes.

PR #759, opened 2018-10-20
Implement Two Factor Authentication

### LINKED ISSUE #756: Two Factor Authentication Support
Due to recent events, it would be prudent to introduce (optional) two factor authentication, via the time-based one-time password (TOTP) algorithm.

Implementation:

* pull in a django 2FA library
* introduce a new intermediate login state for users with 2FA, and in a middleware, check that if `request.path != reverse('2fa_login')`, `return HttpResponseRedirect(reverse('2fa_login') + '?next=' + quote(request.get_full_path())`
* introduce a new 2FA code entering view, with template

We need to have:

* scratch codes and regenerate code interface
* QR code to display TOTP key

quantum5:
Refs #756. Scratch codes will be introduced later.


templates/user/edit-profile.html
```diff
@@ -149,6 +149,17 @@
                                     </a>
                                 </td>
                             </tr>
+                            <tr>
+                                <td colspan="2"><span class="inline-header">
+                                    {% if profile.is_totp_enabled %}
+                                        Two Factor Authentication enabled.
```
Xyene: Internationalization

templates/user/edit-profile.html
```diff
@@ -149,6 +149,17 @@
...
+                                        <a href="#" class="button inline-button">Disable</a>
+                                    {% else %}
+                                        Two Factor Authentication disabled.
```
Xyene: Internationalization

judge/forms.py
```diff
@@ -121,3 +123,19 @@ def __init__(self, *args, **kwargs):
     def _has_social_auth(self, key):
         return (getattr(settings, 'SOCIAL_AUTH_%s_KEY' % key, None) and
                 getattr(settings, 'SOCIAL_AUTH_%s_SECRET' % key, None))
+
+
+class TOTPForm(Form):
+    TOLERANCE = getattr(settings, 'DMOJ_TOTP_TOLERANCE', 1)
```
Xyene: Setting should be self-documenting by specifying units (I'm assuming it's minutes).

templates/registration/totp_enable.html
```diff
@@ -0,0 +1,52 @@
...
+            {% if form.totp_token.errors %}
+                <div class="form-errors">
+                    {{ form.totp_token.errors }}
+                </div>
+            {% endif %}
+            <div>
+                <label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app:') }}</label>
+                <span class="fullwidth">{{ form.totp_token }}</span>
+            </div>
+            <button style="margin-top: 0.5em" class="button" type="submit">{{ _('Enable TOTP') }}</button>
```
Xyene: "Enable 2-factor authentication" sounds more user-friendly than TOTP.

templates/registration/totp_enable.html
```diff
@@ -41,6 +41,24 @@
     </style>
 {% endblock %}
 
+{% block js_media %}
+    <script>
+        $(function () {
+            if (window.navigator.userAgent.indexOf('Edge') >= 0) {
```
Xyene: Doesn't our caniuse app support this kind of lookup server-side?

quantum5:
Before enabling, edit profile:
default

Enable page:
default

After enabling, edit profile:
default

2FA auth screen:
default

Disable screen:
default

judge/forms.py
quantum5: It's the number of TOTP time units, i.e. 30 seconds by default.

templates/registration/totp_enable.html
quantum5: Yes, it does. But since all the compat work is done in JS, might as well detect it in the JS instead of passing it along.

templates/user/edit-profile.html
kiritofeng: I feel like "Two Factor Authentication _is_ enabled." would sound more grammatically correct.

templates/user/edit-profile.html
```diff
@@ -149,6 +149,17 @@
                                     </a>
                                 </td>
                             </tr>
+                            <tr>
+                                <td colspan="2"><span class="inline-header">
+                                    {% if profile.is_totp_enabled %}
+                                        Two Factor Authentication enabled.
+                                        <a href="{{ url('disable_2fa') }}" class="button inline-button">Disable</a>
+                                    {% else %}
+                                        Two Factor Authentication disabled.
```
kiritofeng: Similarly, "Two Factor Authentication _is_ disabled."

templates/registration/totp_enable.html
quantum5: Also caniuse is more focused on individual features than browsers, and [the feature](https://caniuse.com/#feat=css-crisp-edges) support is quite complicated, and I think my hack only works on Edge, or at least whatever supports an obscure canvas feature that has questionable support. This is mostly a user agent thing. I suppose I could make the js show up conditionally on the server side, instead of doing `window.navigator.userAgent`, but I don't think that's really worth the effort.

Issue #729, opened 2018-09-30
Prepare for 7 digit submission count


quantum5:
Changing the total submission count to 7 digits causes an ugly line wrap:

![default](https://user-images.githubusercontent.com/461885/46262281-14bbec80-c4cd-11e8-8e1d-e18cedb06b50.png)

Since we are hitting 7 digits soon, this should be fixed.

Xyene:
I was actually just thinking of scrapping this and going with an interactive pie chart the other day.

motatoes:
Where does the line wrap occur? I tried to use the dev inspector to add an extra digit to `Total` but the display is still fine for me

![](https://i.imgur.com/1JyAR01.png)

motatoes:
But I am up replacing the stats table with a chart if that's needed

PR #761, opened 2018-10-21
Increase submission list infobox size; #729

Xyene:
image


quantum5:
Show a screenshot on mobile. Will not approve unless I have that.

Xyene:
??? We entirely remove the infobox on mobile

quantum5:
What about narrow but not screens?

Issue #768, opened 2018-10-28
User Admin Notes


WallE256:
Add a new field on each User which is admin only, where you can store different information about each User, like when and why the particular User got perms or when the User cheated etc.

kiritofeng:
Implemented by #771.

PR #771, opened 2018-10-29
Add user notes

WallE256:
Implements #768


judge/migrations/0075_add_user_notes.py
```diff
@@ -0,0 +1,31 @@
...
+        ('judge', '0074_totp'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='notes',
+            field=models.TextField(blank=True, help_text='Notes for administrators regarding this user.', null=True, verbose_name='internal notes'),
+        ),
+        migrations.AlterField(
```
Xyene: This shouldn't be in this patch

judge/migrations/0075_add_user_notes.py
```diff
@@ -0,0 +1,31 @@
...
+            model_name='comment',
+            name='page',
+            field=models.CharField(db_index=True, max_length=30, validators=[django.core.validators.RegexValidator(b'^[pcs]:[a-z0-9]+$|^b:\\d+$', 'Page code must be ^[pcs]:[a-z0-9]+$|^b:\\d+$')], verbose_name='associated page'),
+        ),
+        migrations.AlterField(
```
Xyene: This shouldn't be in this patch

judge/migrations/0075_add_user_notes.py
quantum5: Frankly I attribute this to django generating `AlterField` non-database changes. But in any case, the migration needs to be synced eventually. The convention, however ugly, is to sneak it in with the next actual database change. We should probably work out a better system.

kiritofeng:
**Previews:**

Admin interface:
image
As superuser:
image
Impersonating a non-staff user:
image

PR #774, opened 2018-10-31
Add Submission.id_secret and use it for live update

quantum5:
I haven't actually tested this since I don't have an environment with live updates setup. QA requested.


kiritofeng:
I think https://github.com/DMOJ/site/blob/sub-id-secret/judge/bridge/judgecallback.py#L133 still exposes `id`, but I could be wrong.

quantum5:
@kiritofeng that is not a problem, since:

1. it's only sent out for public problems
2. we are not trying to hide the ID, we are trying to hide message queue by introducing unpredictable data into the queue name that must be guessed

PR #775, opened 2018-10-31
Miscellaneous permission fixes for contests

Xyene:


judge/views/problem.py
```diff
@@ -620,6 +620,9 @@ def problem_submit(request, problem=None, submission=None):
 @permission_required('judge.clone_problem')
 def clone_problem(request, problem):
     problem = get_object_or_404(Problem, code=problem)
+    if not problem.is_accessible_by(request.user):
+        raise PermissionDenied()
```
quantum5: Everywhere else uses Http404 to pretend the problem doesn't exist. Using 403 in fact leaks the problem existence.

judge/views/select2.py
```diff
@@ -116,6 +116,9 @@ def get_name(self, obj):
 class ContestUserSearchSelect2View(UserSearchSelect2View):
     def get_queryset(self):
         contest = get_object_or_404(Contest, key=self.kwargs['contest'])
+        if not contest.can_see_scoreboard(self.request) or contest.hide_scoreboard and contest.is_in_contest(self.request):
```
quantum5: So organizers in contests cannot search? This could be fine, but requires reflection.

Issue #762, opened 2018-10-21
Switch submission info tables to graphs


Xyene:
After enough thousand submissions, it becomes fairly useless as-is.

![image](https://user-images.githubusercontent.com/1403503/47263658-c7330e00-d4d3-11e8-93a4-4952050c5d07.png)

Xyene:
Done.

PR #776, opened 2018-10-31
Switch submission table to a graph; #762

Xyene:
This improves readability. Currently, live-updating is disabled.

### DIFF
--- a/templates/submission/list.html
+++ b/templates/submission/list.html
@@ -43,6 +43,9 @@
                 };
             </script>
         {% endif %}
+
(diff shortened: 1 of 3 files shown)


templates/submission/list.html
```diff
@@ -63,6 +66,38 @@
                     .css({'visibility': 'visible'});
             });
 
+            // Draw the statistics graph.
```
quantum5: Why not just put this in a function and call it after `statistics.html(data);`?

PR #777, opened 2018-11-01
Miscellaneous UI fixes

Xyene:


templates/problem/problem.html
```diff
@@ -28,6 +28,19 @@
             display: inline-block;
             padding-left: 3em;
         }
+
+        #content-right {
+            max-width: 12.5em;
+            min-width: 12.5em;
```
quantum5: Would this look fine on intermediate sized screens that are too wide to have it not float, but too narrow for it to fit well?

Xyene: The `max-width` was already inherited, I just made it explicit here. The `min-width` used to be 8em which would cause some nasty wrapping in the float, so I made sure the size is static. Looks better than before, especially at narrower screen sizes.

PR #778, opened 2018-11-01
Implement live submission pie chart update

quantum5:
Refs #762.


judge/views/submission.py
```diff
@@ -159,7 +160,7 @@ class SubmissionsListBase(DiggPaginatorMixin, TitleMixin, ListView):
     context_object_name = 'submissions'
     first_page_href = None
 
-    def get_result_table(self):
+    def get_result_data(self):
         return get_result_table(self.get_queryset().order_by())
```
Xyene: `get_results_table`, I suppose if you're changing this better be thorough

templates/submission/list.html
```diff
@@ -67,35 +67,33 @@
...
-                Chart.defaults.global.tooltipFontFamily =
-                  Chart.defaults.global.tooltipTitleFontFamily =
-                    $('body').css('font-family');
+                var ctx = $('#status-graph').find('canvas')[0].getContext('2d');
+                var font = $('body').css('font-family');
+                new Chart(ctx).Pie(raw_data.categories.map(function (entry) {
+                    return {
+                        color: colors[entry.code],
+                        label: entry.name,
+                        value: entry.count
```
Xyene: ,

templates/submission/list.html
```diff
@@ -67,35 +67,33 @@
...
-                  },
-                {% endfor %}
-              ];
+            function stats_graph(raw_data) {
+                var colors = {
+                    TLE: '#a3bcbd',
+                    AC: '#00a92a',
+                    WA: '#ed4420',
+                    CE: '#42586d',
+                    ERR: '#ffa71c'
```
Xyene: ,

templates/submission/list.html
```diff
@@ -67,35 +67,33 @@
...
+                    return {
+                        color: colors[entry.code],
+                        label: entry.name,
+                        value: entry.count
+                    };
+                }), {
+                    animation: false,
+                    scaleFontFamily: font,
+                    tooltipFontFamily: font,
+                    tooltipTitleFontFamily: font
```
Xyene: ,

templates/submission/list.html
```diff
@@ -297,7 +294,17 @@ <h3>{{ _('Filter submissions') }} <i class="fa fa-search"></i></h3>
...
                 <div class="sidebox">
-                    <div id="statistics-table">{% include "problem/statistics-table.html" %}</div>
+                    <div id="statistics-table">
+                        <h3>{{ _('Statistics') }} <i class="fa fa-pie-chart"></i></h3>
+                        <div class="sidebox-content">
+                            <div id="status-graph">
+                                <canvas width="230" height="170"></canvas>
+                            </div>
+                            <div class="total">
+                                {{ _('Total:') }} <span id="total-submission-count">{{ results_total }}</span>
```
Xyene: Does this update? Also I think you do need to have a `trans` block in case there are languages where the total has to come before the label.

quantum5: I don't think we have to worry about this. It is standard practice in to use "X:" to avoid reordering when translating strings, and we have a heavy dependency on stuff like `Time limit:`.

templates/submission/list.html
quantum5: Trailing `,` makes my IDE vomit, but fine.

judge/views/submission.py
quantum5: k

templates/submission/list.html
quantum5: And yes, it does update. See `$('#total-submission-count').text(data.total);`.

judge/views/submission.py
```diff
@@ -17,7 +18,7 @@
 from judge import event_poster as event
 from judge.highlight_code import highlight_code
 from judge.models import Problem, Submission, Profile, Contest, ProblemTranslation, Language
-from judge.utils.problems import user_completed_ids, user_authored_ids, user_editable_ids, get_result_table
+from judge.utils.problems import user_completed_ids, user_authored_ids, user_editable_ids, get_result_data
```
Xyene: Import order

judge/views/submission.py
```diff
@@ -235,7 +236,9 @@ def get_context_data(self, **kwargs):
         context['all_statuses'] = self.get_searchable_status_codes()
         context['selected_statuses'] = self.selected_statuses
 
-        context['results'] = self.get_result_table()
+        results = self.get_result_data()
+        context['results_total'] = results['total']
```
Xyene: Not sure how value this is when you can just look up `results['total']` directly in the template, but I guess it's OK

templates/submission/list.html
```diff
@@ -158,12 +156,11 @@
...
 
                 function update_stats() {
-                    return;  // TODO: add back live-updating with graph
-
                     $.ajax({
                         url: '?results'
                     }).done(function (data) {
-                        statistics.html(data);
+                        stats_graph(data);
+                        $('#total-submission-count').text(data.total);
```
Xyene: I know this is here just to prevent adding the count again on the initial load, but I think just having it as part of `stats_graph` makes more sense here.

judge/views/submission.py
quantum5: You can't actually look it up in the template unless you parse JSON, which is why this is here.

templates/submission/list.html
quantum5: Sure, might as well.

quantum5: Moved it in.

PR #785, opened 2018-11-02
Fix organizer hidden scoreboard visibility

Xyene:
Entirely UI changes to match the actual page visibility logic.

### DIFF
--- a/templates/contest/list.html
+++ b/templates/contest/list.html
@@ -132,7 +132,7 @@
                             </div>
                         </td>
                         <td>
-                            {% if contest.show_scoreboard or perms.judge.see_private_contest %}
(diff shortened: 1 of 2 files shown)


templates/contest/list.html
```diff
@@ -132,7 +132,7 @@ <h4>{{ _('Ongoing Contests') }}</h4>
                             </div>
                         </td>
                         <td>
-                            {% if contest.show_scoreboard or perms.judge.see_private_contest %}
+                            {% if contest.show_scoreboard or contest.can_see_scoreboard(request) %}
```
quantum5: I feel like the performance of this could be really bad when contest pages are unpaginated and a lot of contests with hidden scoreboards show up. But I am going to let you merge this now and deal with it later.

Issue #780, opened 2018-11-01
on /problems, Users should represent number of people with full AC


xiaowuc1:
The other statistic next to it is AC%, which is the percentage of submissions with AC. With a statistic like this, the `Users` statistic should represent the count of users that have full AC.

This is for parity with basically every other online judge that actively maintains a count of unique users that have solved a problem.

quantum5:
PR is merged.

PR #791, opened 2018-11-05
Have `users` count the number of users with AC; #780

kiritofeng:
WIP; I think I'm missing something but can't find it..

### DIFF
--- a/judge/models/problem.py
+++ b/judge/models/problem.py
@@ -124,7 +124,7 @@ class Problem(models.Model):
     summary = models.TextField(blank=True, verbose_name=_('problem summary'),
                                help_text=_('Plain-text, shown in meta description tag, e.g. for social media.'))
     user_count = models.IntegerField(verbose_name=_('amount of users'), default=0,
-                                     help_text=_('The amount of users on the best solutions page.'))
+                                     help_text=_('The amount of users with an AC submission.'))
     ac_rate = models.FloatField(verbose_name=_('rate of AC submissions'), default=0)
 
     objects = TranslatedProblemQuerySet.as_manager()
@@ -247,7 +247,7 @@ class Problem(models.Model):
         return ProblemClarification.objects.filter(problem=self)
 
     def update_stats(self):
-        self.user_count = self.submission_set.filter(points__gt=0).values('user').distinct().count()
+        self.user_count = self.submission_set.filter(result='AC').values('user').distinct().count()
         submissions = self.submission_set.count()
         self.ac_rate = 100.0 * self.submission_set.filter(result='AC').count() / submissions if submissions else 0
         self.save()


kiritofeng:
Edit: the initial code *was* working, I am unsure why the test site was not reflecting this.

And I'm leaving the filter as-is because it's the same one that the AC percentage uses.

judge/migrations/0076_problem_user_count.py
```diff
@@ -0,0 +1,28 @@
...
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='problem',
+            name='user_count',
+            field=models.IntegerField(default=0, help_text='The number of users with an AC submission.', verbose_name='amount of users'),
+        ),
+
+        migrations.RunPython(compute_user_count, migrations.RunPython.noop),
```
quantum5: How can the reverse be noop if we are changing the meaning of the field?

judge/models/problem.py
```diff
@@ -124,7 +124,7 @@ class Problem(models.Model):
     summary = models.TextField(blank=True, verbose_name=_('problem summary'),
                                help_text=_('Plain-text, shown in meta description tag, e.g. for social media.'))
     user_count = models.IntegerField(verbose_name=_('amount of users'), default=0,
-                                     help_text=_('The amount of users on the best solutions page.'))
+                                     help_text=_('The number of users with an AC submission.'))
```
quantum5: #780 asked specifically for **full AC**.

judge/models/problem.py
```diff
@@ -247,7 +247,7 @@ def clarifications(self):
         return ProblemClarification.objects.filter(problem=self)
 
     def update_stats(self):
-        self.user_count = self.submission_set.filter(points__gt=0).values('user').distinct().count()
+        self.user_count = self.submission_set.filter(result='AC').values('user').distinct().count()
```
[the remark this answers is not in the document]
quantum5: #780 asked specifically for **full AC**. If this is not what we actually want, fix #780.

judge/migrations/0076_problem_statistics.py
```diff
@@ -0,0 +1,41 @@
...
+from __future__ import unicode_literals
+
+import django.core.validators
+from django.db import migrations, models
+from django.db.models import F
+
+def compute_user_count(apps, schema_editor):
+    Problem = apps.get_model('judge', 'Problem')
+    for problem in Problem.objects.using(schema_editor.connection.alias):
+        problem.user_count = problem.submission_set.filter(points__gte=F('problem__points')).values('user').distinct().count()
```
quantum5: This is all the same problem, so you might as well just use `problem.points` instead of that `F` non-sense?

judge/migrations/0076_problem_statistics.py
```diff
@@ -0,0 +1,41 @@
...
+        problem.ac_rate = 100.0 * self.submission_set.filter(points__gte=F('problem__points')).count() / submissions if submissions else 0
```
quantum5: This is all the same problem, so you might as well just use `problem.points` instead of that `F` non-sense?

judge/models/problem.py
```diff
@@ -247,9 +247,9 @@ def clarifications(self):
         return ProblemClarification.objects.filter(problem=self)
 
     def update_stats(self):
-        self.user_count = self.submission_set.filter(points__gt=0).values('user').distinct().count()
+        self.user_count = self.submission_set.filter(points__gte=F('problem__points')).values('user').distinct().count()
```
quantum5: This is all the same problem, so you might as well just use `self.points` instead of that `F` non-sense?

judge/models/problem.py
```diff
@@ -247,9 +247,9 @@ def clarifications(self):
...
         submissions = self.submission_set.count()
-        self.ac_rate = 100.0 * self.submission_set.filter(result='AC').count() / submissions if submissions else 0
+        self.ac_rate = 100.0 * self.submission_set.filter(points__get=F('problem__points')).count() / submissions if submissions else 0
```
quantum5: This is all the same problem, so you might as well just use `self.points` instead of that `F` non-sense?

quantum5: Also I think you should check `result='AC'` while you are at it, since 5.0/5 WA has happened. That is in fact how users who can see submissions are calculated.

judge/models/problem.py
quantum5: Also I think you should check `result='AC'` while you are at it, since 5.0/5 WA has happened. That is in fact how users who can see submissions are calculated.

judge/migrations/0076_problem_statistics.py
quantum5: Also I think you should check `result='AC'` while you are at it, since 5.0/5 WA has happened. That is in fact how users who can see submissions are calculated.

judge/migrations/0076_problem_statistics.py
quantum5: Also I think you should check `result='AC'` while you are at it, since 5.0/5 WA has happened. That is in fact how users who can see submissions are calculated.

PR #801, opened 2018-11-20
Add `rating`, `volatility`, and contest history

Ninjaclasher:
Adds `rating`, `volatility`, and contest history to the user API.

### DIFF
--- a/judge/views/api/api_v1.py
+++ b/judge/views/api/api_v1.py
@@ -5,7 +5,7 @@ from django.http import JsonResponse, Http404
 from django.shortcuts import get_object_or_404
 
 from dmoj import settings
-from judge.models import Contest, Problem, Profile, Submission, ContestTag
+from judge.models import Contest, ContestParticipation, Problem, Profile, Submission, ContestTag
 from judge.views.contests import base_contest_ranking_list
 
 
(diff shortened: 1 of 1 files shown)


judge/views/api/api_v1.py
```diff
@@ -109,18 +109,34 @@ def api_v1_user_list(request):
...
-    })
+    }
+
+    last_rating = list(profile.ratings.order_by('-contest__end_time'))
+
+    contest_history = {}
+    for participation in (ContestParticipation.objects.filter(user=profile, virtual=0, contest__is_public=True, contest__is_private=False)
+                                  .order_by('-contest__end_time')):
+        contest_history[participation.contest.key] = {
+            'rating': participation.rating.rating if hasattr(participation, 'rating') else None,
```
Xyene: This and the line above will execute a query for every contest the user has participated in (= potentially many queries). You probably want to use something like `values_list` here instead.

judge/views/api/api_v1.py
```diff
@@ -109,18 +109,34 @@ def api_v1_user_list(request):
         'rank': rank
     } for username, name, points, rank in queryset})
 
-
```
Xyene: Functions in global scope should leave 2 newlines between themselves.

judge/views/api/api_v1.py
```diff
@@ -136,3 +152,4 @@ def api_v1_user_submissions(request, user):
         'status': sub['status'],
         'result': sub['result'],
     } for sub in subs.values('id', 'problem__code', 'time', 'memory', 'points', 'language__key', 'status', 'result')})
+
```
Xyene: Extra newline.

judge/views/api/api_v1.py
```diff
@@ -5,7 +5,7 @@
 from django.shortcuts import get_object_or_404
 
 from dmoj import settings
-from judge.models import Contest, Problem, Profile, Submission, ContestTag
+from judge.models import Contest, ContestParticipation, Problem, Profile, Submission, ContestTag
```
Xyene: Sort imports.

judge/views/api/api_v1.py
```diff
@@ -114,13 +114,30 @@ def api_v1_user_info(request, user):
...
+    resp = {
         'display_name': profile.name,
         'points': profile.points,
         'rank': profile.display_rank,
         'solved_problems': submissions,
         'organizations': list(profile.organizations.values_list('key', flat=True)),
-    })
+    }
+
+    last_rating = profile.ratings.order_by('-contest__end_time').first()
```
quantum5: Can you just reuse the query you are using below? Otherwise, at least get rid of the sort on that query.

Ninjaclasher:
Updated, although in my opinion this looks less clean. Another possibility would be to remove the `order_by` sort and use `.last()` to get the last rating, but I'm not sure if it will handle well with rerating contests (`rating` `id`s will be changed).

Ninjaclasher:
That's an interesting feature that I was not aware of. However, I don't think that would work in this case, since the  last `rating` or `volatility` might be `None` (if the contest was not rated). I've also removed the sort query, and reverted how `last_rating` is retrieved.

PR #812, opened 2018-11-25
[RFC] add standalone ratings computation based on Codeforces

xiaowuc1:
This implementation is heavily based off of https://github.com/WslF/CF-rating-prediction/blob/master/CF-PredictorBackEnd/src/main/java/com/wslfinc/cf/sdk/rating/RatingCalculatorTeam.java.

This implementation plugs in current CF ratings and titles, and attempts to ignore volatility as CF has no notion of it.

This request does not currently make any effort to add support for multiple rating systems, or to actually use this rating computation live anywhere. Ideally, the rating computation would be tested against production data to ensure the sanity of the resulting ratings before attempting to proceed with possibly moving forward with adding support for this rating system.

### DIFF
--- a/judge/ratings_cf.py
+++ b/judge/ratings_cf.py
@@ -1,0 +1,234 @@
+import math
+from bisect import bisect
+from itertools import izip
+from operator import itemgetter
+
+from django.db import connection, transaction
+from django.db.models import Count
+from django.utils import timezone
+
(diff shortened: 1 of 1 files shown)


xiaowuc1:
closing this for now - with #840 in flight, we may not need this anymore

(we can revisit CF ratings if something else look terribly broken)

PR #815, opened 2018-11-25
Add slimerjs PDF renderer

### LINKED ISSUE #814: Use SlimerJS for PDF rendering
In #442, we introduced PhantomJS. In the intervening two years, PhantomJS has not received a single update, and its development has suspended until further notice. There are various-issues with PhantomJS like kerning, and the ancientness of the browser is alarming.

At the same time, [SlimerJS](https://slimerjs.org/) has appeared, and it uses Gecko. This vastly simplifies things for us, since we can now render math as MathML, instead of using MathJax and coaxing various chrome-based tools (i.e. PhantomJS, `wkhtmltopdf`) to take a snapshot after the math finishes rendering

SlimerJS is in no way perfect: for example, it doesn't support Firefox 60 ESR out of the box, and I have a patched version that supports it, with some bugs related to keyboard input, but it should remain fine for the foreseeable future, and is a strict improvement over PhantomJS.

quantum5:
Fixes #814.

Also, slimerjs patched to support Firefox 60 ESR (at least for PDF rendering):

* [slimerjs-1.0.0-firefox60.zip](https://github.com/DMOJ/site/files/2613909/slimerjs-1.0.0-firefox60.zip)

### DIFF
--- a/judge/management/commands/render_pdf.py
+++ b/judge/management/commands/render_pdf.py
@@ -8,7 +8,7 @@ from django.template.loader import get_template
 from django.utils import translation
 
 from judge.models import Problem, ProblemTranslation
-from judge.pdf_problems import WebKitPdfMaker, PhantomJSPdfMaker, DefaultPdfMaker
+from judge.pdf_problems import DefaultPdfMaker, PhantomJSPdfMaker, SlimerJSPdfMaker
 
 
 class Command(BaseCommand):
@@ -21,7 +21,7 @@ class Command(BaseCommand):
                             help='language to render PDF in')
         parser.add_argument('-p', '--phantomjs', action='store_const', const=PhantomJSPdfMaker,
                             default=DefaultPdfMaker, dest='engine')
-        parser.add_argument('-w', '--wkhtmltopdf', action='store_const', const=WebKitPdfMaker, dest='engine')
+        parser.add_argument('-s', '--slimerjs', action='store_const', const=SlimerJSPdfMaker, dest='engine')
 
     def handle(self, *args, **options):
         try:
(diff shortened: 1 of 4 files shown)


judge/pdf_problems.py
```diff
@@ -114,12 +117,74 @@ def get_render_script(self):
...
+            console.log('Unable to load the address!');
+            phantom.exit(1);
+        } else {
+            page.render(param.output, { ratio: param.zoom });
+            phantom.exit();
+        }
+    });
+} catch (e) {
+    console.error(e);
+    phantom.exit(1);
```
Xyene: I feel like you should use `slimmer` here if we're not on phantom anymore.

judge/pdf_problems.py
```diff
@@ -114,12 +117,74 @@ def get_render_script(self):
...
+            'footer': ugettext('Page [page] of [topage]').replace('[page]', '&P')
+                .replace('[topage]', '&L').encode('utf-8'),
+        }))
+
+    def make(self, debug=False):
+        with io.open(os.path.join(self.dir, '_render.js'), 'w', encoding='utf-8') as f:
+            f.write(self.get_render_script())
+
+        env = None
+        firefox = getattr(settings, 'FIREFOX', '')
```
Xyene: `SLIMMERJS_FIREFOX_PATH` or something more descriptive.

judge/views/problem.py
```diff
@@ -241,7 +241,8 @@ def get(self, request, *args, **kwargs):
                     'problem': problem,
                     'problem_name': problem.name if trans is None else trans.name,
                     'description': problem.description if trans is None else trans.description,
-                    'url': request.build_absolute_uri()
+                    'url': request.build_absolute_uri(),
+                    'math_engine': maker.math_engine
```
Xyene: ,

judge/management/commands/render_pdf.py
```diff
@@ -42,7 +42,8 @@ def handle(self, *args, **options):
                 'problem': problem,
                 'problem_name': problem.name if trans is None else trans.name,
                 'description': problem.description if trans is None else trans.description,
-                'url': ''
+                'url': '',
+                'math_engine': maker.math_engine
```
Xyene: ,

PR #819, opened 2018-12-03
Preserve tree structure of comments

kiritofeng:


judge/comments.py
```diff
@@ -72,6 +72,17 @@ def post(self, request, *args, **kwargs):
         if self.is_comment_locked():
             return HttpResponseForbidden()
 
+        parent = request.POST.get('parent')
+        if parent:
+            try:
+                parent = Comment.objects.get(id=int(parent))
+            except ValueError:
+                return HttpResponseForbidden()
+            if not parent:
```
Xyene: This will actually never be executed; if there is nothing in the set, get will raise `Comment.DoesNotExist`.

judge/comments.py
```diff
@@ -72,6 +72,16 @@ def post(self, request, *args, **kwargs):
         if self.is_comment_locked():
             return HttpResponseForbidden()
 
+        parent = request.POST.get('parent')
+        if parent:
+            try:
+                parent = Comment.objects.filter(hidden=False).get(id=int(parent))
```
quantum5: Why are you doing any of that? Just do something like: `Comment.objects.filter(hidden=False, id=int(parent), page=page).exists()`, and freak out when it's `False`. Also remember to catch `ValueError` from `int(parent)`.

judge/comments.py
```diff
@@ -72,6 +72,14 @@ def post(self, request, *args, **kwargs):
         if self.is_comment_locked():
             return HttpResponseForbidden()
 
+        parent = request.POST.get('parent')
+        if parent:
+            try:
+                if not Comment.objects.filter(hidden=False, id=int(parent), page=page).exists():
+                    return HttpResponseNotFound()
+            except (ValueError, Comment.DoesNotExist):
```
quantum5: `Comment.DoesNotExist` is no longer a possible exception.

Also suggest that you do `parent = int(parent)` separately, only put that in `try:`, and put the actual `exists` call in the `else` branch of the `try`, for proper coding practice.

PR #822, opened 2018-12-04
Remove organization.key

kiritofeng:
Removes `organization.key` from the frontend and the backend.


judge/forms.py
```diff
@@ -106,7 +106,7 @@ class Meta:
 
 class NewOrganizationForm(EditOrganizationForm):
```
quantum5: Is this form even used? If not, can we just kill it?

judge/models/profile.py
```diff
@@ -68,7 +64,7 @@ def get_users_url(self):
         return reverse('organization_users', args=(self.id, self.slug))
 
     class Meta:
-        ordering = ['key']
+        ordering = ['slug']
```
quantum5: Can we just use the name to order? Otherwise people are going to be like "why isn't the organizations sorted properly????"

judge/migrations/0077_remove_organization_key.py
```diff
@@ -0,0 +1,23 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0076_problem_statistics'),
+    ]
+
+    operations = [
+        migrations.AlterModelOptions(
+            name='organization',
+            options={'ordering': ['slug'], 'permissions': (('organization_admin', 'Administer organizations'), ('edit_all_organization', 'Edit all organizations')), 'verbose_name': 'organization', 'verbose_name_plural': 'organizations'},
```
quantum5: This still orders by slug.

Issue #765, opened 2018-10-26
NoReverseMatch at /admin/judge/organization/


syamcode:
Hello, is there still a bug in the organization in the latest version? i dont get this error in the old version (12 september).
i got this error when accessing organization in admin page.
![dmoj](https://user-images.githubusercontent.com/9907952/47569371-a9472c80-d95d-11e8-9985-3105f372d962.PNG)

quantum5:
This is strange. The migration is *supposed* to create slugs for all organizations. Did you run the migrations? If so, what is the organization name?

syamcode:
yes, I run the migration according to the documentation. i use the default organization name, the name is  DMOJ: Modern Online Judge.

this is the screenshot of organization table in database:
![db](https://user-images.githubusercontent.com/9907952/47570613-c7faf280-d960-11e8-9384-6e45766af999.PNG)

quantum5:
Did you load the demo fixture after you migrated? I think the problem is that the demo fixture is outdated. It will be fixed.

For now, run this for a quick fix:

```sql
UPDATE judge_organization SET slug = 'dmoj' WHERE id = 1;
```

syamcode:
thanks, its work now. yes, i load the demo data after migrate.

quantum5:
@kiritofeng @WallE256 the problem is the fixture is outdated and does not contain a slug field. Feel free to fix.

quantum5:
Also we should probably deal with empty slugs too, in case they get accidentally created through some other mechanism.

Fix should be trivial: use `(?P<slug>[\w-]*)` instead of `(?P<slug>[\w-]+)` in the URL pattern.

quantum5:
Fixed with #824.

PR #824, opened 2018-12-05
Update demo to use slugs

kiritofeng:
Addresses #765.


judge/fixtures/demo.json
```diff
@@ -56,7 +56,7 @@
             ],
             "creation_date": "2017-12-02T08:50:25.199Z",
             "is_open": true,
-            "key": "DMOJ",
+            "slug": "1-dmoj",
```
quantum5: No, slug is just `dmoj` here.

PR #831, opened 2018-12-12
Fix contest submission scores not showing up in the API

Ninjaclasher:
Contest submission scores do not show up in the API despite `hide_scoreboard = False`.

### DIFF
--- a/judge/views/api/api_v1.py
+++ b/judge/views/api/api_v1.py
@@ -38,10 +38,14 @@ def api_v1_contest_detail(request, contest):
         can_see_rankings = False
 
     problems = list(contest.contest_problems.select_related('problem')
-                    .defer('problem__description').order_by('order')) if in_contest else []
+                    .defer('problem__description').order_by('order'))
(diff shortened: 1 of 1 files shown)


judge/views/api/api_v1.py
```diff
@@ -38,10 +38,13 @@ def api_v1_contest_detail(request, contest):
         can_see_rankings = False
 
     problems = list(contest.contest_problems.select_related('problem')
-                    .defer('problem__description').order_by('order')) if in_contest else []
+                    .defer('problem__description').order_by('order'))
     users = base_contest_ranking_list(contest, problems, contest.users.filter(virtual=0)
                                       .prefetch_related('user__organizations')
                                       .order_by('-score', 'cumtime')) if can_see_rankings else []
+    if not in_contest and not contest.ended:
```
kiritofeng: This should be reordered so that the code that deals with the problems is all in one place, and the code that deals with users follows it, rather than being inserted right in its midst.

PR #840, opened 2018-12-18
Update new_rating before computing new_volatility.

xiaowuc1:
new_rating is initialized to always be equal to old_rating, which explains why volatility never increased for any users


nealwu:
Nice, thanks for finding this Nick!

PR #841, opened 2018-12-20
Make links in email HTTPS'd

outloudvi:


outloudvi:
It seems that `{{protocol}}` is not substituted automatically. Something else might be added.

Xyene:
Ah, thought it had been tested 😄 

I remember I wanted to do this a while ago and ran into the same issue, which is why I left it as http:// -- thought something had changed in the meantime.

Xyene:
Closing due to inactivity, feel free to reopen.

PR #842, opened 2018-12-22
Add contest scoring/penalty system

### LINKED ISSUE #808: Contest scoring/penalty systems
We want a customizable contest scoring/penalty system. This is an outline of the proposed implementation.

Under the package `judge.contest_format`, we have a bunch of modules implementing the same interface. These modules will implement the actual computation for `ContestParticipation.score` and `ContestParticipation.cumtime`, as well as compute other scoring information, and present it to the user, through classes called `XContestFormat`, which will be registered by `@register_contest_format(name)`. The current contest format will be implemented as the `default` module.

On `Contest`, we will add a new `CharField` called `format`, and the valid choices would be the registered contest formats, as well as a JSON field, probably implemented as a string field (since we aren't on Postgres), which stores the configuration. In addition, we will add `ContestParticipation.format_data`, which is a JSON field, which will store whatever data the `contest_format` module deems fit to store.

The interface of the new `ContestFormat` classes will be as follows, with tentative names:

* [classmethod] `validate`: validates the configuration
* `update_participation`: compute `score` and `cumtime`, as well as populate `format_data`. In `default`, this will sum all the scores and times for `score` and `cumtime`, storing the individual times and points for each problem in `format_data`.
* `display_user_problem`: display the HTML fragment for the cells describing a user's performance on each problem, pulling it from the `ContestParticipation` and `format_data`. In `default`, this will show the points and last submission time.
* `display_participation_result`: display the HTML fragment for the cells describing a user's overall performance, pulling it from the `ContestParticipation` and `format_data`. In `default`, this will show the total points and cumtime.

Changes to the contest system include:

* `ContestParticipation.recalculate_score` and `ContestParticipation.update_cumtime` will be merged into `ContestParticipation.update`. This method will invoke `update_participation` from the correct module
* `base_contest_ranking_list` will now query the participations, and populate the contest table from it. There will be no more raw SQL and other nastiness: all the computation will now be done in `update_participation`, and stored in `format_data`.

Things that will stay the same:

* Ratings and rankings will be implemented by ordering by `-score`, `cumtime`, as before

quantum5:
Task list:
* [x] add base contest format classes
* [x] add migrations
* [x] convert contest ranking and virtual participation list to use new contest format system
* [x] use new method when updating submissions
* [x] remove old update score and cumtime methods
* [x] make new contest format modules generate JSON for APIv1 entry points
* [x] add admin interface to change contest format and edit configuration

This will fix #808.


judge/contest_format/registry.py
```diff
@@ -0,0 +1,16 @@
+from django.utils import six
+
+formats = {}
+
+
+def register_contest_format(name):
```
Xyene: Maybe this should also proactively take a help text, describing the format. Then we could eventually show this in the admin site.

PR #844, opened 2018-12-28
Use new texoid interface

quantum5:
Switch the `texoid` client to use `requests` and the new non-URL-encoded interface for `texoid`.

### DIFF
--- a/judge/utils/texoid.py
+++ b/judge/utils/texoid.py
@@ -1,14 +1,13 @@
 import hashlib
-import logging
-import urllib2
-from contextlib import closing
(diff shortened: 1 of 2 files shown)


judge/utils/texoid.py
```diff
@@ -1,14 +1,13 @@
 import hashlib
-import logging
-import urllib2
-from contextlib import closing
-from urllib import urlencode
 import json
+import logging
```
Xyene: Remove this line.

quantum5: The blank line? Why?

PR #847, opened 2019-01-02
Allow for users to be banned from contests

kiritofeng:
Implements `persona non grata` for contests.


judge/models/problem.py
```diff
@@ -118,7 +118,7 @@ class Problem(models.Model):
     date = models.DateTimeField(verbose_name=_('date of publishing'), null=True, blank=True, db_index=True,
                                 help_text=_("Doesn't have magic ability to auto-publish due to backward compatibility"))
     banned_users = models.ManyToManyField(Profile, verbose_name=_('personae non gratae'), blank=True,
-                                          help_text=_('Bans the selected users from submitting to this problem'))
+                                          help_text=_('Bans the selected users from submitting to this problem. '))
```
Xyene: Extra space at end should not be there. It's present anyway; `sortedm2m` is not hosed.

kiritofeng:
Preview:
image

PR #853, opened 2019-01-04
Remove uses of cdn.mathjax.com

### LINKED ISSUE #838: Remove uses of cdn.mathjax.org
https://cdn.mathjax.org was [retired](https://www.mathjax.org/cdn-shutting-down/) last year, so we should stop using it.

WallE256:
Closes #838

### DIFF
--- a/templates/problem/raw.html
+++ b/templates/problem/raw.html
@@ -46,7 +46,7 @@
 </div>
 {% if math_engine == 'jax' %}
     <script type="text/javascript" src="mathjax_config.js"></script>
-    <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML"></script>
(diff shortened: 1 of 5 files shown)


templates/problem/raw.html
```diff
@@ -46,7 +46,7 @@ <h2 style="color:#393630; display: inline-block">{{ problem_name }}</h2>
 </div>
 {% if math_engine == 'jax' %}
     <script type="text/javascript" src="mathjax_config.js"></script>
-    <script type="text/javascript" src="https://cdn.mathjax.org/mathjax/latest/MathJax.js?config=TeX-AMS_HTML"></script>
+    <script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js?config=TeX-AMS_HTML"></script>
```
quantum5: Why are we using a different version here?

quantum5:
This can be merged now, but we must take into account CSP concerns before deploying on production.

PR #871, opened 2019-01-19
Fix #821

### LINKED ISSUE #821: Submission pie chart expands every update
When using live updates, the submission pie chart becomes larger and larger, pushing everything off the screen.

On dmoj.ca, this bug is mitigated by reverting the pie chart change, but that is not sustainable.

ehhthing:


templates/submission/list.html
```diff
@@ -75,10 +76,13 @@
...
-
-                var ctx = $('#status-graph').find('canvas')[0].getContext('2d');
+                var canvas = $('#status-graph').find('canvas');
+                var ctx = canvas[0].getContext('2d');
                 var font = $('body').css('font-family');
-                new Chart(ctx).Pie(raw_data.categories.map(function (entry) {
+                if (typeof chart != "undefined") {
+                    chart.destroy();
+                }
+                chart = new Chart(ctx).Pie(raw_data.categories.map(function (entry) {
```
Xyene: Newline above.

templates/submission/list.html
```diff
@@ -67,6 +67,7 @@
             });
 
             // Draw the statistics graph.
+            var chart;
```
quantum5: Can we just initialize this to `null` instead of trying to pull nasty stuff to detect `undefined` down there?

PR #882, opened 2019-02-06
support prefix on navbar urls

Maitre-Hiboux:
- Give support to prefix for navbar urls on the site
 - Give opportunity to reverse proxy the whole site on a different location , example : 

Instead of running your site on https://example.com/ you can run it on https://example.com/dmoj/

using simply the "script_name" variable in uwsgi and nginx (or apache if wanted).

I will document more this part when proposing my docker-compose implementation of dmoj-site, dmoj-db and dmoj-judge.

Anyway this works as a standalone in the code but give more flexibility without making any noticeable difference when running the site on "/"

### DIFF
--- a/judge/fixtures/navbar.json
+++ b/judge/fixtures/navbar.json
@@ -7,7 +7,7 @@
             "lft": 1,
             "order": 1,
             "parent": null,
-            "path": "/problems/",
+            "path": "problems/",
             "regex": "^/problem",
             "rght": 2,
             "tree_id": 1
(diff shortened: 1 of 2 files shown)


quantum5:
I do not think DMOJ should support this feature, since no official deployment uses it. We continue to recommend that a separate domain be chosen. This is a good idea for many reasons, including but not limited to being considered as a separate origin and more complete cookie isolation.

This is a scenario limited to unofficial forks, and should be maintained as part of them. I regret to inform you that your change will not be merged into the official DMOJ repository.

PR #883, opened 2019-02-06
Fix ticket notes

outloudvi:
The featherlight in `templates/ticket/ticket.html` POSTed to the `action` of the form[1], which doesn't agree with `urls.py`[2]. This PR modified the `action` to get it working.

[1]. https://github.com/DMOJ/site/blob/master/templates/ticket/ticket.html#L201
[2]. https://github.com/DMOJ/site/blob/master/dmoj/urls.py#L304

### DIFF
--- a/templates/ticket/edit-notes.html
+++ b/templates/ticket/edit-notes.html
@@ -1,4 +1,4 @@
-<form id="edit-notes" action="./notes" method="post">
+<form id="edit-notes" action="{{ url('ticket_notes', ticket.id) }}" method="post">
     {% csrf_token %}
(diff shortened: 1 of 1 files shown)


templates/ticket/edit-notes.html
```diff
@@ -1,5 +1,5 @@
-<form id="edit-notes" action="" method="post">
+<form id="edit-notes" action="./notes" method="post">
```
quantum5: Please use `reverse` instead of relative links. We do not rely on URL hierarchy and always reverse the URL.

PR #890, opened 2019-02-22
Display members' full name for super user.

kenneth-l-i:
As a superuser (or staff), just listing all member's username is not very informative. (My site use the student number as username).
So I show the full name next to the username in the username list.


quantum5:
This change is not useful for the official DMOJ site. You are welcome to make such changes yourself and keep it inside your own fork, but I do not think this is appropriate material for a pull request.

PR #895, opened 2019-02-28
Add "View Submissions" button on Problem admin

### LINKED ISSUE #488: "View submissions" button on problem admin
Would bring up a search filtered by that problem code &mdash; not precisely ideal, but it should work.

Ninjaclasher:
Fixes #488.

### DIFF
--- a/judge/admin/submission.py
+++ b/judge/admin/submission.py
@@ -124,6 +124,9 @@ class SubmissionAdmin(admin.ModelAdmin):
             return True
         return obj.problem.is_editor(request.user.profile)
 
+    def lookup_allowed(self, key, value):
+        return super(SubmissionAdmin, self).lookup_allowed(key, value) or key in ('problem__code',)
+
     def judge(self, request, queryset):
         if not request.user.has_perm('judge.rejudge_submission') or not request.user.has_perm('judge.edit_own_problem'):
             self.message_user(request, ugettext('You do not have the permission to rejudge submissions.'),
--- a/templates/admin/judge/problem/change_form.html
+++ b/templates/admin/judge/problem/change_form.html
@@ -1,0 +1,20 @@
+{% extends "admin/change_form.html" %}
+{% load i18n %}
+
+{% block extrahead %}{{ block.super }}
+    <script>
+        django.jQuery(function ($) {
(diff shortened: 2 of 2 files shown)


judge/admin/submission.py
```diff
@@ -124,6 +124,9 @@ def has_change_permission(self, request, obj=None):
             return True
         return obj.problem.is_editor(request.user.profile)
 
+    def lookup_allowed(self, key, value):
+        return super(SubmissionAdmin, self).lookup_allowed(key, value) or key in ['problem__code']
```
Xyene: Using a tuple is more idiomatic in this case.

quantum5: Tuples are also faster for this kind of thing, but that's not a significant concern. It is more idiomatic and that's what should matter.

templates/admin/judge/problem/change_form.html
```diff
@@ -0,0 +1,20 @@
...
+        django.jQuery(function ($) {
+            $('.submissionslink').appendTo('div#bottombar').show();
+        });
+    </script>
+{% endblock extrahead %}
+
+{% block after_field_sets %}{{ block.super }}
+    {% if original %}
+        <a style="display: none" title="{% trans "View Submissions" %}" href="{% url 'admin:judge_submission_changelist' %}?problem__code={{ original.code }}"
+           class="button submissionslink">
```
quantum5: Can we keep to convention and use hyphenated CSS class names?

templates/admin/judge/problem/change_form.html
```diff
@@ -0,0 +1,20 @@
...
+    <script>
+        django.jQuery(function ($) {
+            $('.submissionslink').appendTo('div#bottombar').show();
+        });
+    </script>
+{% endblock extrahead %}
+
+{% block after_field_sets %}{{ block.super }}
+    {% if original %}
+        <a style="display: none" title="{% trans "View Submissions" %}" href="{% url 'admin:judge_submission_changelist' %}?problem__code={{ original.code }}"
```
quantum5: Line too long, swap `href` with `class` so that `href` gets the whole line to itself.

PR #899, opened 2019-03-02
Paginate contest list

Ninjaclasher:
Not sure if manually slicing the past contests is the best solution. I tried to modify `get_paginator` to include only past contests, but it ended up making the code even more messy, as the past contests and current contests would be processed in 2 different places.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -31,7 +31,7 @@ from judge.models import Problem
 from judge.timezone import from_database_time
 from judge.utils.opengraph import generate_opengraph
 from judge.utils.ranker import ranker
-from judge.utils.views import TitleMixin, generic_message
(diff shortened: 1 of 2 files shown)


judge/views/contests.py
```diff
@@ -93,13 +94,17 @@ def get_context_data(self, **kwargs):
                     active.append(participation)
                     present.remove(participation.contest)
 
+        start = self.paginate_by * (context['page_obj'].number - 1)
+        past = past[start:start+self.paginate_by]
```
quantum5: We definitely cannot do this. Suggestion is to paginate the queryset for past contests.

templates/contest/list.html
```diff
@@ -10,6 +10,9 @@
             color: #777;
             padding-top: 0.5em;
         }
+        .content-description ul {
```
Xyene: Nit: newline above here.

PR #902, opened 2019-03-02
Add "My tickets" on problem pages

Ninjaclasher:
Currently, the only way for normal users to view the tickets they have submitted is on https://dmoj.ca/tickets. However, many users who are just beginning do not realize that the page exists, as there is no link to it anywhere. Many users submit a ticket, and think they lose access to it, and never see the problem setter's response. 

I suggest that a "My tickets" link be added to each problem page, similar to how curators and authors can see the tickets for their problems. The "My tickets" would link to a page which shows the user the tickets they have submitted for the problem.

### DIFF
--- a/templates/problem/problem.html
+++ b/templates/problem/problem.html
@@ -139,6 +139,13 @@
         {% if not problem.is_manually_managed %}
             <div><a href="{{ url('problem_data', problem.code) }}">{{ _('Edit test data') }}</a></div>
         {% endif %}
+    {% elif request.user.is_authenticated and has_tickets %}
+        <hr>
+        <div>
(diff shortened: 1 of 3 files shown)


templates/problem/problem.html
```diff
@@ -139,6 +139,13 @@ <h2 style="color:#393630; display: inline-block">{{ title }}</h2>
         {% if not problem.is_manually_managed %}
             <div><a href="{{ url('problem_data', problem.code) }}">{{ _('Edit test data') }}</a></div>
         {% endif %}
+    {% elif request.user.is_authenticated %}
+        <hr>
+        <div>
+            <a href="{{ url('problem_ticket_list', problem.code) }}">{{ _('My tickets') }}
```
Xyene: I think we should only show this if we have a nonzero number of tickets.

PR #904, opened 2019-03-02
Cache submission statistics table for unfiltered global submissions list

quantum5:
To do:

* [x] make `_get_result_data` not cache language/status filters
* [x] only update table via Ajax after cache clears
* [x] make cache time configurable, or at least have one place to control it (once we add ajax timeout)


quantum5:
Potential future improvement:

* increment relevant field in bridge when finishing a submission and store back into cache
* increment value in graph when a new submission comes in through JavaScript

PR #914, opened 2019-03-15
Add puppeteer PDF renderer

quantum5:


judge/pdf_problems.py
```diff
@@ -32,6 +42,12 @@ def load(self, file, source):
             target.write(source.read())
 
     def make(self, debug=False):
+        self._make(debug)
+
+        if self.title and HAS_EXIFTOOL:
+            subprocess.call([EXIFTOOL, '-Title=%s' % (self.title,), self.pdffile])
```
Xyene: The lack of logging here has me concerned.

quantum5: I was thinking that it's unimportant since nothing really happens if `exiftool` dies, but yea, probably should log it...

quantum5: Added logging for `exiftool` errors.

quantum5:
We need... `libasound2 libxss1 libxtst6`.

Issue #909, opened 2019-03-09
Extended judge feedback


Xyene:
For some problems, the 64-byte string we allow inline in testcases is not enough. For these, I propose the addition of an extended feedback field in the test case model. From a checker's point of view, this would be a new keyword argument in `CheckerResult` (or we could automatically use the extended field when the length is over 64).

![image](https://user-images.githubusercontent.com/1403503/54066778-2fc12400-4204-11e9-8812-ef1464251f63.png)

quantum5:
This is on production now.

PR #916, opened 2019-03-16
Add extended feedback; #909

kiritofeng:
Implements #909.

Preview: (please excuse my choice of outputs)
Screenshot-2019-03-15-20-44-34


judge/bridge/judgecallback.py
```diff
@@ -351,14 +351,15 @@ def on_test_case(self, packet, max_feedback=SubmissionTestCase._meta.get_field('
         test_case.total = packet['total-points']
         test_case.batch = self.batch_id if self.in_batch else None
         test_case.feedback = (packet.get('feedback', None) or '')[:max_feedback]
+        test_case.extended_feedback = packet.get('extended_feedback', None) or ''
```
quantum5: `dict.get(x)` returns `None` when not found, so the second argument is hardly necessary.

PR #917, opened 2019-03-16
Require access code for virtual contest joins; #892

kiritofeng:


judge/views/contests.py
```diff
@@ -259,6 +259,9 @@ def join_contest(self, request, access_code=None):
                                          'You are permanently barred from joining this contest.'))
 
         if contest.ended:
+            if not (request.user.is_superuser or self.is_organizer) \
+                        and contest.access_code and access_code != contest.access_code:
+                raise ContestAccessDenied()
             while True:
```
Xyene: Newline above this.

PR #919, opened 2019-03-16
fix problems solved pluralization on user page

xiaowuc1:


templates/user/user-base.html
```diff
@@ -38,7 +38,12 @@
             </div>
             <br>
 
-            <div><b>{{ _('%(count)d problems solved', count=user.problem_count) }}</b></div>
+            {% if user.problem_count == 1 %}
```
quantum5: Rule 1 of i18n: do not do plural rules yourself. Russian has two plural forms, and in other languages, plural is > 1.

Use, for example:

```jinja2
{% trans trimmed counter=submissions_left %}
    {{ counter }} submission left
{% pluralize %}
    {{ counter }} submissions left
{% endtrans %}
```

PR #928, opened 2019-03-24
Fix placing of `pretest-*` CSS classes

kiritofeng:
If a contest is not marked as `run_pretests_only`, prestests will not be run, even if `is_pretested` is true for a contest problem. However, not all contest setters diable both flags once a problem has been system tested, which gives the false impression that the problem was not system tested yet.

### DIFF
--- a/judge/contest_format/default.py
+++ b/judge/contest_format/default.py
@@ -49,7 +49,7 @@ class DefaultContestFormat(BaseContestFormat):
         if format_data:
             return format_html(
                 u'<td class="{state}"><a href="{url}">{points}<div class="solving-time">{time}</div></a></td>',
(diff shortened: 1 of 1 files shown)


judge/contest_format/default.py
```diff
@@ -49,7 +49,7 @@ def display_user_problem(self, participation, contest_problem):
         if format_data:
             return format_html(
                 u'<td class="{state}"><a href="{url}">{points}<div class="solving-time">{time}</div></a></td>',
-                state=('pretest-' if contest_problem.is_pretested else '') +
+                state=('pretest-' if participation.contest.run_pretests_only and contest_problem.is_pretested else '') +
```
quantum5: `self.contest`. Don't want to run another query now, do you?

PR #930, opened 2019-03-27
Prefix all constants with DMOJ_

kiritofeng:
New naming convention for constants is `DMOJ_MODULE_<NAME>`, as decided in Slack.

### DIFF
--- a/judge/migrations/0055_add_performance_points.py
+++ b/judge/migrations/0055_add_performance_points.py
@@ -11,13 +11,15 @@ from django.db.models import Max
 def gen_pp(apps, schema_editor):
     Profile = apps.get_model('judge', 'Profile')
     Problem = apps.get_model('judge', 'Problem')
-    table = (lambda x: [pow(x, i) for i in xrange(100)])(getattr(settings, 'PP_STEP', 0.95))
-    bonus_function =  getattr(settings, 'PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
+    _pp_step = getattr(settings, 'DMOJ_PP_STEP', 0.95)
+    table = [pow(_pp_step, i) for i in xrange(getattr(settings, 'DMOJ_PP_ENTRIES', 100))]
+    bonus_function =  getattr(settings, 'DMOJ_PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
     for row in Profile.objects.all():
         data = (Problem.objects.filter(submission__user=row, submission__points__isnull=False, is_public=True)
                 .annotate(max_points=Max('submission__points')).order_by('-max_points')
                 .values_list('max_points', flat=True))
-        extradata = Problem.objects.filter(submission__user=row, submission__result='AC', is_public=True).values('id').distinct().count()
+        extradata = Problem.objects.filter(submission__user=row, submission__result='AC', is_public=True) \
+                        .values('id').distinct().count()
         size = min(len(data), len(table))
         row.performance_points = sum(map(mul, table[:size], data[:size])) + bonus_function(extradata)
         row.save()
--- a/judge/models/profile.py
+++ b/judge/models/profile.py
@@ -120,13 +120,17 @@ class Profile(models.Model):
     def username(self):
         return self.user.username
 
-    def calculate_points(self, table=(lambda x: [pow(x, i) for i in xrange(100)])(getattr(settings, 'PP_STEP', 0.95))):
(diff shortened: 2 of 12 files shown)


judge/models/profile.py
```diff
@@ -120,13 +120,17 @@ def organization(self):
...
-                .annotate(max_points=Max('submission__points')).order_by('-max_points')
-                .values_list('max_points', flat=True).filter(max_points__gt=0))
-        extradata = Problem.objects.filter(submission__user=self, submission__result='AC', is_public=True).values('id').distinct().count()
-        bonus_function = getattr(settings, 'PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
+        data = (Problem.objects.filter(submission__user=self, submission__points__isnull=False, is_public=True,
+                                       is_organization_private=False)
+                    .annotate(max_points=Max('submission__points')).order_by('-max_points')
+                    .values_list('max_points', flat=True).filter(max_points__gt=0))
+        extradata = Problem.objects.filter(submission__user=self, submission__result='AC', is_public=True).values('id')
+                        .distinct().count()
```
Xyene: This isn't a syntax error?

judge/migrations/0055_add_performance_points.py
```diff
@@ -11,13 +11,15 @@
 def gen_pp(apps, schema_editor):
     Profile = apps.get_model('judge', 'Profile')
     Problem = apps.get_model('judge', 'Problem')
-    table = (lambda x: [pow(x, i) for i in xrange(100)])(getattr(settings, 'PP_STEP', 0.95))
-    bonus_function =  getattr(settings, 'PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
+    _pp_step = getattr(settings, 'DMOJ_PP_STEP', 0.95)
+    table = (lambda x: [pow(x, i) for i in xrange(getattr(settings, 'DMOJ_PP_ENTRIES', 100))])(_pp_step)
```
quantum5: Why do we still need this `lambda`? It only existed in the old code because whoever wrote it does not want to have a variable.

judge/migrations/0055_add_performance_points.py
```diff
@@ -11,13 +11,15 @@
...
+    bonus_function =  getattr(settings, 'DMOJ_PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
     for row in Profile.objects.all():
         data = (Problem.objects.filter(submission__user=row, submission__points__isnull=False, is_public=True)
                 .annotate(max_points=Max('submission__points')).order_by('-max_points')
                 .values_list('max_points', flat=True))
-        extradata = Problem.objects.filter(submission__user=row, submission__result='AC', is_public=True).values('id').distinct().count()
+        extradata = Problem.objects.filter(submission__user=row, submission__result='AC', is_public=True).values('id')
+                        .distinct().count()
```
quantum5: SyntaxError

judge/models/profile.py
quantum5: It is.

quantum5:
Also provide a list of changes, and preferably, a `sed` or `perl` script to edit `local_settings.py`.

quantum5:
This doesn't look broken, but given that past changes have _syntax errors_, I would like to see this code being run.

quantum5:
Still need a `local_settings.py` update script.

kiritofeng:
Update script: https://gist.github.com/kiritofeng/7e03a9fd77fc46e2e4dd1538a176a5dc

judge/models/profile.py
```diff
@@ -120,13 +120,17 @@ def organization(self):
...
+    _pp_table = [pow(_pp_step, i) for i in xrange(getattr(settings, 'DMOJ_PP_ENTRIES', 100))]
+    def calculate_points(self, table=_pp_table):
         from judge.models import Problem
-        data = (Problem.objects.filter(submission__user=self, submission__points__isnull=False, is_public=True, is_organization_private=False)
-                .annotate(max_points=Max('submission__points')).order_by('-max_points')
-                .values_list('max_points', flat=True).filter(max_points__gt=0))
-        extradata = Problem.objects.filter(submission__user=self, submission__result='AC', is_public=True).values('id').distinct().count()
-        bonus_function = getattr(settings, 'PP_BONUS_FUNCTION', lambda n: 300 * (1 - 0.997 ** n))
+        data = (Problem.objects.filter(submission__user=self, submission__points__isnull=False, is_public=True,
+                                       is_organization_private=False) \
```
quantum5: Why are you using `\` when between `(` and `)`?

Issue #926, opened 2019-03-23
`Output (Clipped)` UI Issues


kiritofeng:
Currently, `Output (Clipped)` is shown whenever `output_prefix_length` is not set to `0` in a problem's `init.yml`.

In contests, this field will still appear, even if `output_prefix_override` is set to 0. To make this less confusing, the field should be hidden if either `output_prefix_length` or the override is set to 0.

kiritofeng:
Code is currently in production.

PR #933, opened 2019-03-29
Fix `Your output` UI Issues; #926

kiritofeng:
Removes the pane if `prefix_length` equals 0, and also changes the heading to `Your output` if there is no `prefix_length` is set to `None`.

### DIFF
--- a/templates/submission/status-testcases.html
+++ b/templates/submission/status-testcases.html
@@ -36,7 +36,7 @@
         <table>{% for case in batch.cases %}
             <tr id="{{ case.id }}" class="case-row toggle closed">
                 <td>
-                    {%- if case.status != 'AC' and case.output -%}
+                    {%- if case.status != 'AC' and case.output and (prefix_length is None or prefix_length > 0) -%}
                         <i class="fa fa-chevron-right fa-fw"></i>
                     {%- endif -%}
                     {%- if batch.id -%}
@@ -72,7 +72,7 @@
                 {% endif %}
             </tr>
 
-            {% if case.status != 'AC' and case.output %}
+            {% if case.status != 'AC' and case.output and (prefix_length is None or prefix_length > 0) %}
                 <tr id="{{ case.id }}-output" style="display:none" class="case-feedback toggled">
                     <td colspan="5">
                         <div class="case-info">


templates/submission/status-testcases.html
```diff
@@ -72,14 +72,15 @@ <h3>{{ _('Execution Results') }}</h3>
                 {% endif %}
             </tr>
 
-            {% if case.status != 'AC' and case.output %}
+            {% if case.status != 'AC' and case.output and prefix_length != 0 %}
```
quantum5: Can you just be like `> 0`?

templates/submission/status-testcases.html
```diff
@@ -36,7 +36,7 @@ <h3>{{ _('Execution Results') }}</h3>
         <table>{% for case in batch.cases %}
             <tr id="{{ case.id }}" class="case-row toggle closed">
                 <td>
-                    {%- if case.status != 'AC' and case.output -%}
+                    {%- if case.status != 'AC' and case.output and prefix_length != 0 -%}
```
quantum5: Can you just be like `> 0`? Makes more sense semantically.

templates/submission/status-testcases.html
```diff
@@ -36,7 +36,7 @@ <h3>{{ _('Execution Results') }}</h3>
         <table>{% for case in batch.cases %}
             <tr id="{{ case.id }}" class="case-row toggle closed">
                 <td>
-                    {%- if case.status != 'AC' and case.output -%}
+                    {%- if case.status != 'AC' and case.output and prefix_length > 0 -%}
```
quantum5: Should actually be `prefix_length is None or prefix_length > 0`, since you can't compare `None` with `0`. We also want to be explicit that it can be `None`.

templates/submission/status-testcases.html
```diff
@@ -72,7 +72,7 @@ <h3>{{ _('Execution Results') }}</h3>
                 {% endif %}
             </tr>
 
-            {% if case.status != 'AC' and case.output %}
+            {% if case.status != 'AC' and case.output and prefix_length > 0 %}
```
quantum5: Ditto.

PR #935, opened 2019-03-29
Create registration_closed.html

kenneth-l-i:
To handle the SIGSEGV: invalid_state error due to REGISTRATION_OPEN = False in local_settings.py

### DIFF
--- a/templates/registration/registration_closed.html
+++ b/templates/registration/registration_closed.html
@@ -1,4 +1,4 @@
 {% extends "base.html" %}
 {% block body %}
-    <p>{{ _('Registration is currently closed. Please contact an administrator') }}</p>
+    <p>{{ _('Registration is currently closed. Please contact an administrator.') }}</p>
 {% endblock %}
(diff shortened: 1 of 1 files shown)


templates/registration/registration_closed.html
```diff
@@ -0,0 +1,3 @@
+{% block body %}
```
kiritofeng: You should extend `base.html`

templates/registration/registration_closed.html
```diff
@@ -0,0 +1,3 @@
...
+    <p>{{ _('We temporarily closes the user registration. For enquiries, please contact the administrator.') }}</p>
```
kiritofeng: Can you change the message to `Registration is currently closed. Please contact an administrator.`

templates/registration/registration_closed.html
```diff
@@ -0,0 +1,4 @@
+{% extends "base.html" %}
+{% block body %}
+    <p>{{ _('Registration is currently closed. Please contact an administrator') }}</p>
```
quantum5: Need a period at the end of the sentence.

PR #948, opened 2019-04-14
Add ability to abort queued submissions; #845

### LINKED ISSUE #845: Add ability to abort queued submissions
We display the _Abort_ button, but it's a no-op until the submission is scheduled on a judge. This is suboptimal when faced with a large backlog of long-running submissions.

Ninjaclasher:
Fixes #845.

### DIFF
--- a/judge/bridge/judgelist.py
+++ b/judge/bridge/judgelist.py
@@ -20,6 +20,7 @@ class JudgeList(object):
         self.queue = dllist()
         self.priority = [self.queue.append(PriorityMarker(i)) for i in range(self.priorities)]
         self.judges = set()
+        self.node_map = {}
         self.submission_map = {}
         self.lock = RLock()
 
@@ -38,6 +39,7 @@ class JudgeList(object):
                             logger.exception('Failed to dispatch %d (%s, %s) to %s', id, problem, language, judge.name)
                             self.judges.remove(judge)
                             return
+                        del self.node_map[id]
                         self.queue.remove(node)
                         break
                 node = node.next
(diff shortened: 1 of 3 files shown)


judge/judgeapi.py
```diff
@@ -97,4 +101,11 @@ def disconnect_judge(judge, force=False):
 
 
 def abort_submission(submission):
-    judge_request({'name': 'terminate-submission', 'submission-id': submission.id}, reply=False)
+    from .models import Submission
+    response = judge_request({'name': 'terminate-submission', 'submission-id': submission.id})
+    # This defaults to true, so that in the case the judgelist fails to remove the submission from the queue,
+    # and returns a bad-request, the submission is not falsely shown as "Aborted" when it will still be judged.
+    if not response.get('judge-aborted', True):
+        Submission.objects.filter(id=submission.id).update(status='AB', result='AB')
```
Xyene: Wait... you have the `Submission` already, why are you filtering for it again?

judge/bridge/judgelist.py
```diff
@@ -78,7 +80,14 @@ def on_judge_free(self, judge, submission):
...
             logger.info('Abort request: %d', submission)
-            self.submission_map[submission].abort()
+            try:
+                self.submission_map[submission].abort()
+                success = True
+            except KeyError:
+                self.queue.remove(self.node_map[submission])
+                del self.node_map[submission]
+                success = False
+            return success
```
Xyene: Return in both branches directly.

judge/judgeapi.py
quantum5: It's not like this runs an extra query. In fact, this generates a smaller update query.

PR #950, opened 2019-04-15
Remove duplicate `get_content_title` in ProblemDataView

Ninjaclasher:


judge/views/problem_data.py
```diff
@@ -144,11 +144,6 @@ def get_content_title(self):
         return format_html(_('Editing data for <a href="{1}">{0}</a>'), self.object.name,
                            reverse('problem_detail', args=[self.object.code]))
 
-    def get_content_title(self):
```
quantum5: No, get rid of the other one. This one clearly is designed to stop bad translators from injecting random HTML.

Issue #437, opened 2016-08-05
Migration squashing


quantum5:
This issue will track migration squashing. But we must first answer the question: how often should we squash all our migrations? Every 50? Every 100? Every 200?

Documentation: https://docs.djangoproject.com/en/stable/topics/migrations/#squashing-migrations

This is obviously low priority.

/cc @Xyene

Xyene:
Every 50 sounds fine, right now we're over 60 and it is noticeably slow even on a fast server.

quantum5:
You realize that _you_ will have to write part of the squashed migration, right? Django isn't smart enough to drop all the transition stuff we have.

Xyene:
Mhm, which is why it remains a low priority.

quantum5:
From Django 2.0 release notes:

> The `on_delete` argument for `ForeignKey` and `OneToOneField` is now required in models and migrations. Consider squashing migrations so that you have fewer of them to update.

We better do this.

Xyene:
We did this. We might do this again someday, but let's close this issue out for now.

PR #957, opened 2019-04-18
Squashed migrations (up to 0084)

quantum5:
Refs #437.


judge/migrations/0001_squashed_0084_contest_formats.py
```diff
@@ -0,0 +1,765 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.20 on 2019-04-18 02:06
+from __future__ import unicode_literals
```
Xyene: Yikes

quantum5: What's so bad about this line? It's Django-generated.

PR #964, opened 2019-04-18
Add on_delete=CASCADE explicitly to ForeignKey/OneToOneField

### LINKED ISSUE #962: Add on_delete=CASCADE to all ForeignKeys
Django 2.0+ requires `on_delete` to always be defined. In the past, the default is `CASCADE`, so we should specify `on_delete=CASCADE` for all `ForeignKey` and `OneToOneField` that do not already have it. This will also need to be done in the migrations.

Do this after #957 is merged so that you only have to deal with the squashed migration.

quantum5:
Fixes #962

### DIFF
--- a/judge/models/problem.py
+++ b/judge/models/problem.py
@@ -5,7 +5,7 @@ from django.contrib.contenttypes.fields import GenericRelation
 from django.core.cache import cache
 from django.core.validators import RegexValidator
 from django.db import models
-from django.db.models import F, QuerySet
+from django.db.models import F, QuerySet, CASCADE, SET_NULL
 from django.db.models.expressions import RawSQL
 from django.db.models.functions import Coalesce
 from django.urls import reverse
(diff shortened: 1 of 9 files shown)


judge/models/problem.py
```diff
@@ -5,7 +5,7 @@
 from django.core.cache import cache
 from django.core.validators import RegexValidator
 from django.db import models
-from django.db.models import F, QuerySet
+from django.db.models import F, QuerySet, CASCADE, SET_NULL
```
kiritofeng: On [line 363](https://github.com/DMOJ/site/blob/a856120187627ce046567f1c8ddbbac0872a189c/judge/models/problem.py#L363), the `models.SET_NULL` should just be `SET_NULL`.

judge/models/problem.py
```diff
@@ -105,7 +105,7 @@ class Problem(models.Model):
                                      help_text=_(
                                          'These users will be able to view a private problem, but not edit it.'))
     types = models.ManyToManyField(ProblemType, verbose_name=_('problem types'))
-    group = models.ForeignKey(ProblemGroup, verbose_name=_('problem group'))
+    group = models.ForeignKey(ProblemGroup, verbose_name=_('problem group'), on_delete=models.CASCADE)
```
kiritofeng: Shouldn't this be `on_delete=CASCADE`, since you explicitly imported `CASCADE`?

quantum5:
By the way don't bother checking for the lack of `CASCADE`. It wouldn't have started on Django 2.0 if it missed any `CASCADE`.

PR #973, opened 2019-04-20
Provide extra submission metadata for judges

### LINKED ISSUE #802: Provide extra submission metadata for judges
Notably, allowing judges access to a hash of the submitter's username as well as attempt number would be light in terms of computation and protocol changes, but allow for some more interesting grading schemes (e.g. running on different test data for different attempts).

Ninjaclasher:
Fixes #802.

### DIFF
--- a/judge/bridge/judgecallback.py
+++ b/judge/bridge/judgecallback.py
@@ -9,7 +9,7 @@ from django.utils import timezone
 from judge import event_poster as event
 from judge.caching import finished_submission
 from judge.models import Submission, SubmissionTestCase, Problem, Judge, Language, LanguageLimit, RuntimeVersion
-from .judgehandler import JudgeHandler
+from .judgehandler import JudgeHandler, SubmissionData
 
 logger = logging.getLogger('judge.bridge')
 json_log = logging.getLogger('judge.json.bridge')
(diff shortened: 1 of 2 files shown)


judge/bridge/judgecallback.py
```diff
@@ -70,12 +71,15 @@ def get_related_submission_data(self, submission):
             ))
             return
 
+        attempt_no = Submission.objects.filter(problem__id=pid, contest__participation__id=part_id,
```
Xyene: Pretty sure this shouldn't include CE/IE.

judge/bridge/judgecallback.py
```diff
@@ -70,12 +71,15 @@ def get_related_submission_data(self, submission):
...
+                                               user__id=uid, date__lt=sub_date).count() + 1
+
         try:
             time, memory = (LanguageLimit.objects.filter(problem__id=pid, language__id=lid)
                                          .values_list('time_limit', 'memory_limit').get())
         except LanguageLimit.DoesNotExist:
             pass
-        return time, memory, short_circuit, is_pretested
+        return time, memory, short_circuit, is_pretested, part_virtual, attempt_no, uid
```
Xyene: Can we define a `namedtuple` for this?

judge/bridge/judgehandler.py
```diff
@@ -104,7 +104,7 @@ def working(self):
         return bool(self._working)
 
     def get_related_submission_data(self, submission):
-        return 2, 16384, False, False
+        return 2, 16384, False, False, None, 1, None
```
Xyene: This tuple is really getting out of hand...

Ninjaclasher:
Turns out I was accidentally filtering _for_ CEs and IEs, instead of excluding them. This should be fixed in the latest commit. Sorry.

PR #980, opened 2019-04-25
Hide organization-private problem submissions when not in organization

Ninjaclasher:


judge/views/submission.py
```diff
@@ -212,8 +212,14 @@ def _get_queryset(self):
...
         queryset = self._get_queryset()
-        if not self.in_contest and not self.request.user.has_perm('judge.see_private_problem'):
-            queryset = queryset.filter(problem__is_public=True)
+        if not self.in_contest:
+            if not self.request.user.has_perm('judge.see_private_problem'):
+                queryset = queryset.filter(problem__is_public=True)
+            if not self.request.user.has_perm('judge.see_organization_problem'):
+                filter = Q(problem__is_organization_private=False)
+                if self.request.user.is_authenticated:
+                    filter |= Q(problem__organizations__id__in=self.request.profile.organizations.all())
```
quantum5: Do not use `problem__organizations__id__in`. This forces the organization table to be joined in as well. You simply need `problem__organizations__in`.

I realize other code does this as well, but they shall all be fixed.

PR #982, opened 2019-04-26
Remove __id__in anti-pattern

quantum5:
Basically, doing `__in` works for a foreign key, and there absolutely no reason to join in the target table just to filter the primary key.


quantum5:
This is already running in production.

Issue #981, opened 2019-04-26
Celery workers and progress reports


quantum5:
Currently, a lot of admin site work takes place in the web worker. This has led to timeouts and uwsgi's `harakiri` triggering, and clearly is unsustainable. A lot of features also can't be implemented because they are too slow to be done in a request, for example, #40.

We should add `celery` to the site, and use it to do background work. We can report progress in the tasks, and have a page that reports the status before redirecting to the page that'll be shown when the task is complete.

This page should take the following GET parameters:

* `task_id`: celery task ID
* `message`: a description of what the task is doing
* `redirect`: the URL to redirect to after the task is done

This page should periodically refresh and display the task's progress, if available, in a progress bar. Once the task is done, it should redirect to the URL specified in the `redirect` GET parameter.

/cc @Xyene @kiritofeng @WallE256

quantum5:
#983 has been merged and deployed to production, and it works well.

PR #983, opened 2019-04-28
Add Celery and job status page to display progress for long tasks

quantum5:
This requires you to have celery beats running so that Celery can cleanup the finished tasks.

We cannot cleanup the finished tasks ourselves because we want refreshing the job status page to continue working after the task finishes. If the result is cleaned up, the task is shown as pending, which is confusing.

Refs #981.


templates/task_status.html
```diff
@@ -0,0 +1,97 @@
...
+        <div class="progress progress-complete">
+            <div class="progress-bar" role="progressbar">Completed!</div>
+        </div>
+        <div class="progress progress-failed">
+            <div class="progress-bar" role="progressbar">Failed!</div>
+        </div>
+    </div>
+</div>
+</body>
+</html>
```
kiritofeng: Missing newline at end of file.

PR #985, opened 2019-04-28
Make TimezoneMiddleware reuse request.profile

quantum5:


quantum5:
Note that this saves one SQL query every request.

PR #986, opened 2019-04-29
Upgrade Django requirement to 2.1

quantum5:
This should only be merged once testing on Django 2.1 concludes.


quantum5:
I think this should be approved after running 2.1 in production for months...

PR #990, opened 2019-05-02
Move submission source out of `Submission` model

kiritofeng:
Tested with the `fakeprod` data, and obtained the following times:
```
$ time python manage.py migrate judge 0085_submission_source
Operations to perform:
  Target specific migration: 0085_submission_source, from judge
Running migrations:
  Applying judge.0085_submission_source... OK

real	10m53.223s
user	0m11.840s
sys	0m1.936s
```

```
$ time python manage.py migrate judge 0001_squashed_0084_contest_formats
Operations to perform:
  Target specific migration: 0001_squashed_0084_contest_formats, from judge
Running migrations:
  Rendering model states... DONE
  Unapplying judge.0085_submission_source... OK

real	12m0.273s
user	0m15.244s
sys	0m2.276s
```


judge/views/problem.py
```diff
@@ -565,7 +565,12 @@ def problem_submit(request, problem=None, submission=None):
             else:
                 model = form.save()
 
+            # Create the SubmissionSource object
+            source = SubmissionSource(submission=model, source=form.cleaned_data['source'])
+            source.save()
```
quantum5: I have a bad feeling about this. If a `ContestSubmission` is lost, nothing happens, but if a `SubmissionSource` is lost, all hell breaks loose. Suggest that you put the and the previous block in `with transaction.atomic():`.

judge/views/problem.py
```diff
@@ -548,25 +548,33 @@ def problem_submit(request, problem=None, submission=None):
...
 
-            profile.update_contest()
-            model.judge(rejudge=False)
+                # Create the SubmissionSource object
+                source = SubmissionSource(submission=model, source=form.cleaned_data['source'])
+                source.save()
+                profile.update_contest()
+
+                # Save a query
+                model.source = source
```
quantum5: No, don't bother doing this inside the transaction. If the submission fails to be judged, it should still be saved.

PR #996, opened 2019-05-05
Add submission management interface

quantum5:
Sample Screenshot

Rejudge Warning

### DIFF
--- a/templates/problem/manage_submission.html
+++ b/templates/problem/manage_submission.html
@@ -83,7 +83,20 @@
                         return;
                     }
                 }
-                $(this).parents('form').submit();
+
+                var $form = $(this).parents('form');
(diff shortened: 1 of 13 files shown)


templates/problem/manage_submission.html
```diff
@@ -0,0 +1,153 @@
...
+                </div>
+                <a id="rejudge-selected" class="unselectable button full" href="#">
+                    {{ _('Rejudge selected submissions') }}
+                </a>
+            </form>
+        </div>
+
+        <div class="pane">
+            <h3>{{ _('Rescore Everything') }}</h3>
+            <p id="rescore-warning">{{ _('This will rescore %(count)d submissions.', count=submission_count) }}</p>
```
kiritofeng: Should we also add a warning for rejudges?

quantum5: This is done.

PR #997, opened 2019-05-05
Add ECOO Contest Format

kiritofeng:
Retains almost the same SQL query as the original written for the ECOO branch, but with new UI and more configurable options.

Preview

### DIFF
--- a/judge/contest_format/ecoo.py
+++ b/judge/contest_format/ecoo.py
@@ -1,0 +1,121 @@
+from datetime import timedelta
+from django.core.exceptions import ValidationError
+from django.db import connection
+from django.template.defaultfilters import floatformat
+from django.urls import reverse
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+from django.utils.translation import gettext_lazy
+
+from judge.contest_format.default import DefaultContestFormat
+from judge.contest_format.registry import register_contest_format
+from judge.timezone import from_database_time
+from judge.utils.timedelta import nice_repr
+
+
+@register_contest_format('ecoo')
+class ECOOContestFormat(DefaultContestFormat):
+    name = gettext_lazy('ECOO')
+    config_defaults = {'cumtime': False, 'first_ac_bonus': 10, 'time_bonus': 5}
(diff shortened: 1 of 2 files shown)


judge/contest_format/ecoo.py
```diff
@@ -0,0 +1,121 @@
...
+                        bonus += self.config['first_ac_bonus']
+                    # Time bonus
+                    if self.config['time_bonus']:
+                        bonus += (participation.end_time - time).total_seconds() // 60 // self.config['time_bonus']
+                    points += bonus
+
+                format_data[str(prob)] = {'time': dt, 'points': score, 'bonus': bonus}
+                points += score
+
+        participation.cumtime = 0
```
quantum5: If you are going to keep track of cumtime for optional time breaking, this should be `cumtime` and not `0`.

judge/contest_format/ecoo.py
```diff
@@ -0,0 +1,121 @@
...
+        participation.score = points
+        participation.format_data = format_data
+        participation.save()
+
+    def display_user_problem(self, participation, contest_problem):
+        format_data = (participation.format_data or {}).get(str(contest_problem.id))
+        if format_data:
+            bonus = format_html('<font style="font-size:10px;"> +{bonus}</font>',
```
quantum5: What? Why don't you just use `<small>` or something? `<font>` is proscribed.

judge/contest_format/ecoo.py
```diff
@@ -0,0 +1,121 @@
...
+
+            return format_html(
+                '<td class="{state}"><a href="{url}">{points}{bonus}<div class="solving-time">{time}</div></a></td>',
+                state=('pretest-' if self.contest.run_pretests_only and contest_problem.is_pretested else '') +
+                       self.best_solution_state(format_data['points'], contest_problem.points),
+                url=reverse('contest_user_submissions',
+                            args=[self.contest.key, participation.user.user.username, contest_problem.problem.code]),
+                points=floatformat(format_data['points']),
+                bonus=bonus,
+                time=nice_repr(timedelta(seconds=format_data['time']), 'noday') if self.config['cumtime'] else '',
```
quantum5: I think time is useful even if `cumtime` is not used for breaking. Just don't show `cumtime` on the result cell if it's not used for tie breaking.

PR #999, opened 2019-05-07
Fix ordering in ProblemTicketListView.

PeiAllen:
Fix ordering from "Oldest to Newest" to "Newest to Oldest", so that it is the same as dmoj.ca/tickets.


Xyene:
Thanks!

PR #1011, opened 2019-05-17
Update problem.py

magicalsoup:
Kirito said this was fine?


judge/views/problem.py
```diff
@@ -536,7 +536,7 @@ def problem_submit(request, problem=None, submission=None):
         if form.is_valid():
             if (not request.user.has_perm('judge.spam_submission') and
                         Submission.objects.filter(user=profile, was_rejudged=False).exclude(
-                            status__in=['D', 'IE', 'CE', 'AB']).count() > 2):
+                            getattr(settings, DMOJ_SUBMISSIONS_MAX_CONCURRENT_SUBS, 2)
```
WallE256: This just gets the particular setting, breaking the query

PR #1020, opened 2019-06-09
Add validators on problem resource limits

kiritofeng:
This is so that we are not hit by a wave of IEs when the judge fails to convert the time/memory limit to `unsigned long`.


judge/models/problem.py
```diff
@@ -3,7 +3,7 @@
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.core.cache import cache
-from django.core.validators import RegexValidator
+from django.core.validators import MinValueValidator, MaxValueValidator, RegexValidator
```
Xyene: Alphabetize

quantum5:
@kiritofeng please rebase this migration in a separate PR, and quickly before this gets deployed.

PR #1028, opened 2019-07-26
Add PP to API; #1026

### LINKED ISSUE #1026: expose pp in API
The original sum of points metric is exposed in the API. Given that one can sort by pp but not by this metric on production, it perhaps makes sense to also expose pp.

kiritofeng:
Closes #1026.


xiaowuc1:
in general, avoid making style changes that are unrelated to the PR

PR #1029, opened 2019-07-26
Fix up APIv1 formatting

kiritofeng:


Xyene:
Fixed when we started enforcing flake.

PR #1039, opened 2019-09-02
Allow for private contests with specific users to be created

kiritofeng:
This pull renames `is_public` to `is_visible`, and `is_private` to `is_organization_private`, and adds a new Boolean field `is_private`.

The new field allows for the creation of contests that only specific users can see, and these contests (as well as organization-private contests) can only be created if the user has a new permission, to prevent abuse.


judge/migrations/0087_problem_resource_limits.py
```diff
@@ -9,7 +9,7 @@
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('judge', '0085_submission_source'),
+        ('judge', '0086_rating_ceiling'),
```
quantum5: ??? Why are we editing existing migrations here?

quantum5: Okay I see this is actually broken. Fix it separately.

PR #1040, opened 2019-09-02
Use Autoprefixer for CSS postprocessing

WallE256:
Change from Pleeease to Autoprefixer as Pleeease has not been updated in at least 2 years.


make_style.sh
```diff
@@ -1,7 +1,26 @@
...
+
 cd `dirname $0`
-sass --update resources:sass_processed
-pleeease compile sass_processed/style.css -t resources/style.css
-mv sass_processed/content-description.css resources/content-description.css
-mv sass_processed/table.css resources/table.css
-mv sass_processed/ranks.css resources/ranks.css
+sass -f --update resources:sass_processed
+echo
+postcss sass_processed/style.css --verbose --use autoprefixer -d resources
```
quantum5: I don't really like this copying. Maybe `for file in a b c; do ... done`?

make_style.sh
```diff
@@ -1,7 +1,23 @@
...
+  echo 'Error: postcss is not installed.' >&2
+  exit 1
+fi
+
+if ! [ -x "$(command -v autoprefixer)" ]; then
+  echo 'Error: autoprefixer is not installed.' >&2
+  exit 1
+fi
+
+FILES="sass_processed/style.css sass_processed/content-description.css sass_processed/table.css sass_processed/ranks.css"
```
quantum5: If you are going to use a variable, use an array... 😭

make_style.sh
```diff
@@ -1,7 +1,23 @@
...
 cd `dirname $0`
-sass --update resources:sass_processed
-pleeease compile sass_processed/style.css -t resources/style.css
-mv sass_processed/content-description.css resources/content-description.css
-mv sass_processed/table.css resources/table.css
-mv sass_processed/ranks.css resources/ranks.css
+sass -f --update resources:sass_processed
+
+echo
+postcss ${FILES[*]} --verbose --use autoprefixer -d resources
```
quantum5: This doesn't actually work...

```sh
$ FILES=(a 'b c')
$ for i in ${FILES[*]}; do echo "$i"; done
a
b
c
$ for i in "${FILES[@]}"; do echo "$i"; done
a
b c
```

PR #1041, opened 2019-09-04
Spectating has `virtual=-1`, not `1`.

kiritofeng:


judge/management/commands/runmoss.py
```diff
@@ -30,7 +30,7 @@ def handle(self, *args, **options):
             for dmoj_lang, moss_lang in self.LANG_MAPPING:
                 print("%s: " % dmoj_lang, end=' ')
                 subs = Submission.objects.filter(
-                    contest__participation__virtual__in=(0, 1),
+                    contest__participation__virtual__in=(0, -1),
```
Xyene: Actually could we introduce a constant somewhere so we don't make this bug in the future?

quantum5: Yes, `ContestParticipation.LIVE`, `ContestParticipation.SPECTATE`.

PR #1048, opened 2019-09-10
Unconditionally cause editorials to 404 in contest; #1035

### LINKED ISSUE #1035: in contest, unconditionally make editorials 404
The crux of this is that there's basically no practical situation where a user should need to access the editorial for a given problem while in a contest. (This issue presupposes the ideology that the platform should inherently support best practices for sane user behavior, and should be marked as `wontfix` in the event that this is not a fair assumption, or if the scenarios are perhaps only applicable on the production instance of DMOJ and therefore not generally applicable for all.)

There are a couple scenarios for which this would be a strict improvement over the current editorial hiding implementation, which is a gate on time. The first scenario is when a contest needs to be extended for unforeseen reasons, where the editorials were originally scheduled to go public at the original contest end.

A more plausible scenario, which probably appears in practice on the production instance of DMOJ with respect to Olympiads homework occasionally, is when a problem with a public editorial is reused in a contest.

The final scenario where this would be helpful is when doing a virtual contest, to better simulate contest conditions.

kiritofeng:
Fixes #1035.

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -120,7 +120,9 @@ class ProblemSolution(SolvedProblemMixin, ProblemMixin, TitleMixin, CommentedDet
         solution = get_object_or_404(Solution, problem=self.object)
 
         if (not solution.is_public or solution.publish_on > timezone.now()) and \
-                not self.request.user.has_perm('judge.see_private_solution'):
+                not self.request.user.has_perm('judge.see_private_solution') or \
+                (self.request.user.is_authenticated and
+                 self.request.profile.current_contest):
             raise Http404()
         context['solution'] = solution
         context['has_solved_problem'] = self.object.id in self.get_completed_problems()
--- a/templates/problem/problem.html
+++ b/templates/problem/problem.html
@@ -126,7 +126,8 @@
     {% endif %}
     <div><a href="{{ url('chronological_submissions', problem.code) }}">{{ _('All submissions') }}</a></div>
     <div><a href="{{ url('ranked_submissions', problem.code) }}">{{ _('Best submissions') }}</a></div>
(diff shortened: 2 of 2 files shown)


judge/views/problem.py
```diff
@@ -120,7 +120,9 @@ def get_context_data(self, **kwargs):
         solution = get_object_or_404(Solution, problem=self.object)
 
         if (not solution.is_public or solution.publish_on > timezone.now()) and \
-                not self.request.user.has_perm('judge.see_private_solution'):
+                not self.request.user.has_perm('judge.see_private_solution') or \
+                (self.request.user.is_authenticated and
+                 self.request.profile.current_contest):
```
Xyene: Can't this be on the same line and still be < 120 chars?

templates/problem/problem.html
```diff
@@ -136,8 +136,10 @@ <h2 style="color:#393630; display: inline-block">{{ title }}</h2>
     <div><a href="{{ url('chronological_submissions', problem.code) }}">{{ _('All submissions') }}</a></div>
     <div><a href="{{ url('ranked_submissions', problem.code) }}">{{ _('Best submissions') }}</a></div>
     {% if editorial and editorial.is_public %}
-        <hr>
-        <div><a href="{{ url('problem_editorial', problem.code) }}">{{ _('Read editorial') }}</a></div>
+        {% if not (request.user.is_authenticated and request.profile.current_contest) %}
```
Xyene: Collapse this if statement into the immediately preceding one.

judge/views/problem.py
kiritofeng: It's 124 characters.

PR #1049, opened 2019-09-12
Add Today link to contest calendar view; #724

### LINKED ISSUE #724: Contest calendar should show today
Instead of just previous and next months, we should also show today, so we can easily get back to the current month:

default

TheAvidDev:
Fixed #724

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -410,6 +410,7 @@ class ContestCalendar(TitleMixin, ContestListMixin, TemplateView):
 
         context['now'] = timezone.now()
         context['calendar'] = self.get_table()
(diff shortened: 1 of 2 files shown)


Xyene:
The code looks good, thanks. What does this look like in a screenshot?

TheAvidDev:
What it looks like on the current month:
today

What it looks like when you are at min_month (no context prior):
today_next
This works with the max_month too.

What it looks like before current month but after an earlier contest.
today_all

TheAvidDev:
Accidentally closed.

Xyene:
Looks good, thanks!

PR #1051, opened 2019-09-17
Move ANSI HTML-ifying to site

Ninjaclasher:
No idea how long it will take or whether it will OOM on production data.

### DIFF
--- a/judge/migrations/0091_compiler_message_ansi2html.py
+++ b/judge/migrations/0091_compiler_message_ansi2html.py
@@ -1,0 +1,23 @@
+# -*- coding: utf-8 -*-
+import lxml.html as lh
+from django.db import migrations
+from lxml.html.clean import clean_html
+
+
+def strip_error_html(apps, schema_editor):
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False).iterator():
+        if sub.error:
+            sub.error = clean_html(lh.fromstring(sub.error)).text_content()
+            sub.save(update_fields=['error'])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0090_fix_contest_visibility'),
+    ]
+
+    operations = [
+        migrations.RunPython(strip_error_html, migrations.RunPython.noop, atomic=True),
+    ]
--- a/judge/jinja2/reference.py
+++ b/judge/jinja2/reference.py
@@ -2,6 +2,7 @@ import re
 from collections import defaultdict
 from urllib.parse import urljoin
 
+from ansi2html import Ansi2HTMLConverter
 from django.contrib.auth.models import AbstractUser
 from django.urls import reverse
 from lxml.html import Element
@@ -178,3 +179,8 @@ def join(first, second, *rest):
     if not rest:
         return urljoin(first, second)
     return urljoin(urljoin(first, second), *rest)
+
+
+@registry.filter(name='ansi2html')
+def ansi2html(s):
+    return Ansi2HTMLConverter(inline=True).convert(s, full=False)
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,3 +28,4 @@ jsonfield
 pymoss
 packaging
 celery
+ansi2html
--- a/templates/submission/status-testcases.html
+++ b/templates/submission/status-testcases.html
@@ -12,11 +12,11 @@
         <h4>{{ _('Your submission is being processed...') }}</h4>
     {% elif submission.status == 'CE' %}
         <h3>{{ _('Compilation Error') }}</h3>
-        <pre>{{ submission.error|safe }}</pre>
+        <pre>{{ submission.error|ansi2html }}</pre>
     {% else %}
         {% if submission.error %}
             <h3>{{ _('Compilation Warnings') }}</h3>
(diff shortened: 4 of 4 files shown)


judge/migrations/0088_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,31 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+import lxml.html as lh
+from lxml.html.clean import clean_html
+
+
+def strip_error_html(apps, schema_editor):
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False):
+        sub.error = clean_html(lh.fromstring(sub.error)).text_content() if sub.error else None
```
quantum5: If you really care about performance, you'd skip the `save` if `not sub.error`.

judge/migrations/0088_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,31 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import lxml.html as lh
+from lxml.html.clean import clean_html
+
+
+def strip_error_html(apps, schema_editor):
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False):
```
quantum5: You would use [`QuerySet.iterator`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iterator) with a large chunk size if you cared about this not being insanely slow and use up all the memory.

judge/migrations/0088_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,31 @@
...
+        ('judge', '0087_problem_resource_limits'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='submission',
+            name='error',
+            field=models.TextField(null=True, verbose_name='compile errors'),
+        ),
+        migrations.RunPython(strip_error_html, migrations.RunPython.noop),
```
quantum5: You would use `atomic=True` to run `strip_error_html`  in a transaction if you cared about performance.

judge/migrations/0088_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,31 @@
...
+import django.db.models.deletion
+import lxml.html as lh
+from lxml.html.clean import clean_html
+
+
+def strip_error_html(apps, schema_editor):
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False):
+        sub.error = clean_html(lh.fromstring(sub.error)).text_content() if sub.error else None
+        sub.save(update_fields=['error'])
```
quantum5: The better thing to do is to use [`Manager.bulk_update`](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#bulk-update) after we migrate to 2.2. And if we are doing that, don't buffer everything in memory, run `bulk_update` every 2k submissions or something.

judge/models/submission.py
```diff
@@ -71,7 +71,7 @@ class Submission(models.Model):
     status = models.CharField(verbose_name=_('status'), max_length=2, choices=STATUS, default='QU', db_index=True)
     result = models.CharField(verbose_name=_('result'), max_length=3, choices=SUBMISSION_RESULT,
                               default=None, null=True, blank=True, db_index=True)
-    error = models.TextField(verbose_name=_('compile errors'), null=True, blank=True)
+    error = models.TextField(verbose_name=_('compile errors'), null=True)
```
Xyene: After further discussion it seems it might be best to leave this as-is, since the alternatives all seem worse.

judge/migrations/0088_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,31 @@
...
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False).iterator():
+        sub.error = clean_html(lh.fromstring(sub.error)).text_content() if sub.error else None
+        sub.save(update_fields=['error'])
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0087_problem_resource_limits'),
```
Xyene: Rebase on top of 0090 :)

judge/migrations/0088_compiler_message_ansi2html.py
Xyene: (Note for @Ninjaclasher: getting this PR in isn't blocked on 2.2.)

judge/migrations/0091_compiler_message_ansi2html.py
```diff
@@ -0,0 +1,23 @@
...
+from django.db import migrations
+import lxml.html as lh
+from lxml.html.clean import clean_html
+
+
+def strip_error_html(apps, schema_editor):
+    Submission = apps.get_model('judge', 'Submission')
+    for sub in Submission.objects.filter(error__isnull=False).iterator():
+        if sub.error:
+            sub.error = clean_html(lh.fromstring(sub.error)).text_content()
```
Xyene: `fromstring` will fail if `sub.error` is an empty string.

PR #1059, opened 2019-09-25
Replaced request.user.profile with user.profile; #782

### LINKED ISSUE #782: Switch to using request.profile
We have a lot of nastiness that involves `request.user.profile` and other more complicated variations which we will be able to replace with `request.profile` after #781.

Behaviour: `request.profile` is a `judge.Profile` model instance when logged in, and `None` when not logged in.

TheAvidDev:
All occurrences of `request.user.profile` replaced with `user.profile` except those in `middleware.py`.

Fixes #782


judge/views/select2.py
```diff
@@ -72,7 +72,10 @@ def get_queryset(self):
             q = Q(is_private=False, is_organization_private=False)
             if self.request.user.is_authenticated:
                 q |= Q(organizations__in=self.request.profile.organizations.all())
+<<<<<<< HEAD
+=======
                 q |= Q(private_contestants=self.request.profile)
+>>>>>>> 4021a4a01bdde1694c551bcb7ea54e5e958bccd6
```
Xyene: Merge conflict

judge/views/submission.py
```diff
@@ -195,7 +195,11 @@ def _get_queryset(self):
                                                               language=self.request.LANGUAGE_CODE), to_attr='_trans'))
         if self.in_contest:
             queryset = queryset.filter(contest__participation__contest_id=self.contest.id)
+<<<<<<< HEAD
+            if self.contest.hide_scoreboard and self.contest.is_in_contest(self.request):
+=======
             if self.contest.hide_scoreboard and self.contest.is_in_contest(self.request.user):
+>>>>>>> 4021a4a01bdde1694c551bcb7ea54e5e958bccd6
```
Xyene: Merge conflict

judge/views/select2.py
TheAvidDev: Fixed

judge/views/submission.py
TheAvidDev: Fixed

PR #1060, opened 2019-09-25
Added searching by status code; #938

### LINKED ISSUE #938: Filter submissions by status code/abbreviation
Add the ability to search for status in the filter submissions box via status code/abbreviation. To elaborate, have the "Wrong Answer" filter appear when user types in "WA", have the "Run Time Error" filter appear when user types in "RTE", etc.

image

TheAvidDev:
Allows searching by status code for submission Status and Language.

Fixes #938

### DIFF
--- a/templates/submission/list.html
+++ b/templates/submission/list.html
@@ -62,12 +62,12 @@
 
                 function matchAll(params, data) {
                     if ($.trim(params.term) === '') {
-                      return data;
+                        return data;
                     }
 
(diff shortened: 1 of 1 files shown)


Xyene:
Closing in favour of #1085, thanks for taking the time to PR, though 😃 One tricky thing about internationalization is that 'toLowerCase' doesn't always do what you'd expect for languages that aren't English, and may produce wrong results.

PR #1066, opened 2019-09-26
Switch to using join for best submissions

quantum5:
Instead of using the old `WHERE id IN ...`.

This required adding a helper `join_sql_subquery`.


quantum5:
Tested on prod with 20× speed improvement on the query.

Issue #1068, opened 2019-09-27
Problem data management page broken on jQuery 3


Xyene:
It seems the upgrade was not backwards-compatible, and all instances of things like `'tr[data-type=S], tr[data-type=E], tr[data-type=]'` need to be changed to `'tr[data-type="S"], tr[data-type="E"], tr[data-type=""]'`.

PR #1069, opened 2019-09-27
Fix problem data management page; #1068

xiaowuc1:


quantum5:
Actually no, it works just fine. The problem is not the quoting around stuff, it's the lack of quotes around the empty string.

PR #1107, opened 2019-10-04
Moves constants to 'settings.py'; #993

### LINKED ISSUE #993: Move constants to `settings.py`
Currently constants are accessed as `getattr(settings, SETTING_NAME, SETTING_DEFAULT)`.

This leads to a lot of hardcoding of `SETTING_DEFAULT`, as well as unnecessarily long code that could be avoided if these constants and their default values were instead added to `settings.py`.

Carson-Tang:
Fixes #993

### DIFF
--- a/dmoj/settings.py
+++ b/dmoj/settings.py
@@ -10,6 +10,7 @@ https://docs.djangoproject.com/en/1.11/ref/settings/
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 import os
+import tempfile
 
 from django.utils.translation import ugettext_lazy as _
 from django_jinja.builtins import DEFAULT_EXTENSIONS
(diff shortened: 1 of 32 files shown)


kiritofeng:
Please make https://github.com/DMOJ/site/blob/master/judge/views/blog.py#L41 a constant too

judge/views/blog.py
```diff
@@ -71,7 +72,8 @@ def get_context_data(self, **kwargs):
                                                       .exclude(problem__in=user_completed_ids(user))
```
Xyene: Hehe, this is all dead code...

dmoj/settings.py
```diff
@@ -31,8 +32,92 @@
 SITE_ID = 1
 SITE_NAME = 'DMOJ'
 SITE_LONG_NAME = 'DMOJ: Modern Online Judge'
+SITE_ADMIN_EMAIL = False
+
+'''
```
quantum5: I think we typically just use `#` for comments.

judge/pdf_problems.py
```diff
@@ -32,7 +31,7 @@ class BasePdfMaker(object):
     title = None
 
     def __init__(self, dir=None, clean_up=True):
-        self.dir = dir or os.path.join(getattr(settings, 'DMOJ_PDF_PROBLEM_TEMP_DIR', tempfile.gettempdir()),
+        self.dir = dir or os.path.join(settings.DMOJ_PDF_PROBLEM_TEMP_DIR,
                                        str(uuid.uuid1()))
```
quantum5: Put on same line as before.

judge/views/register.py
```diff
@@ -46,8 +46,8 @@ def clean_email(self):
                                                 'is allowed per address.') % self.cleaned_data['email'])
         if '@' in self.cleaned_data['email']:
             domain = self.cleaned_data['email'].split('@')[-1].lower()
-            if (domain in getattr(settings, 'BAD_MAIL_PROVIDERS', ()) or
-                    any(regex.match(domain) for regex in bad_mail_regex)):
+            if (domain in settings.BAD_MAIL_PROVIDERS) or \
```
quantum5: Why the backslash? It's supposed be to using brackets...

judge/widgets/mixins.py
```diff
@@ -34,7 +34,7 @@ class CompressorWidgetMixin(object):
     except ImportError:
         pass
     else:
-        if getattr(settings, 'COMPRESS_ENABLED', not getattr(settings, 'DEBUG', False)):
```
quantum5: Semantics actually changed. Before, if you override `DEBUG`, and `COMPRESS_ENABLED` is not set, then `DEBUG` will be used. Now you have to override `COMPERSS_ENABLED` explicitly.

judge/widgets/mixins.py
```diff
@@ -34,7 +34,7 @@ class CompressorWidgetMixin(object):
...
+        if settings.COMPRESS_ENABLED and not settings.DEBUG:
```
quantum5: No, this also changes semantics because if you do `DEBUG = False`, `COMPRESS_ENABLED` is ignored. The correct solution here is `getattr(settings, 'COMPRESS_ENABLED', not settings.DEBUG)` and removing `COMPRESS_ENABLED` from `settings.py`.

PR #1116, opened 2019-10-12
Pass raw source bytes to MOSS

Xyene:


judge/management/commands/runmoss.py
```diff
@@ -48,6 +48,6 @@ def handle(self, *args, **options):
                     if username in users:
                         continue
                     users.add(username)
-                    moss_call.add_file_from_memory(username, source.encode('utf-8'))
+                    moss_call.add_file_from_memory(username, source)
```
quantum5: Huh? It seems like you are passing a `str` instead of `bytes`.

quantum5: After examining the pymoss code, I am still convinced that `source` must be `bytes`, and something else is going on. I will not approve this unless an explanation of exactly why this patch works is given.

quantum5:
Confirm that `runmoss` works without this patch.

PR #1127, opened 2019-10-21
Make manage submissions rejudging idempotent; fixes #1009

### LINKED ISSUE #1009: Manage submissions rejudging is not idempotent
Reproduction steps:

1) Rejudge some submissions via a manage submissions page.
2) Rejudge them again while some submissions are in `QU` state.

Expected result:
1) `QU` subs stay in `QU`
2) non-`QU` subs become `QU`

Actual result:
1) [correct behavior about non-`QU` subs]
2) `QU` subs become `IE`, with a further side-effect that new `QU` subs are automatically `IE`

Xyene:


judge/bridge/judgelist.py
```diff
@@ -99,12 +99,14 @@ def check_priority(self, priority):
     def judge(self, id, problem, language, source, priority):
         with self.lock:
             if id in self.submission_map or id in self.node_map:
-                logger.warning('Already judging? %d', id)
+                # Already judging, don't queue again. This can happen during batch rejudges, rejudges should be
```
Xyene: Removed this warning because despite filtering earlier (which is useful for showing users accurate submission counts), we can still race to this point.

PR #1130, opened 2019-10-22
Add time/memory/point bounds as constants; #1036

### LINKED ISSUE #1036: cap maximum possible TL that users can input
Most problems do not require more than a few seconds per test case.

Another possible mitigation: make regular problemsetters only able to set TLs within a sane range.

Carson-Tang:
Resolves #1036


dmoj/settings.py
```diff
@@ -56,6 +56,11 @@
 DMOJ_CAMO_HTTPS = False
 DMOJ_CAMO_EXCLUDE = ()
 DMOJ_PROBLEM_DATA_ROOT = None
+DMOJ_PROBLEM_MIN_TIME_LIMIT = 0
+DMOJ_PROBLEM_MAX_TIME_LIMIT = 60
+DMOJ_PROBLEM_MIN_MEMORY_LIMIT = 0
+DMOJ_PROBLEM_MAX_MEMORY_LIMIT = 1048576
```
Xyene: Comment that this is in kb

dmoj/settings.py
```diff
@@ -56,6 +56,11 @@
 DMOJ_CAMO_HTTPS = False
 DMOJ_CAMO_EXCLUDE = ()
 DMOJ_PROBLEM_DATA_ROOT = None
+DMOJ_PROBLEM_MIN_TIME_LIMIT = 0
+DMOJ_PROBLEM_MAX_TIME_LIMIT = 60
```
Xyene: Comment that this is in seconds

PR #1132, opened 2019-10-22
i18n: add Japanese and Portuguese

Xyene:


dmoj/settings.py
```diff
@@ -335,7 +335,9 @@
     ('fr', _('French')),
     ('hr', _('Croatian')),
     ('hu', _('Hungarian')),
+    ('ja', _('Japanese')),
     ('ko', _('Korean')),
+    ('pt-BR', _('Portuguese, Brazilian')),
```
quantum5: Is this really the best way to write it? I think the typical style is `Portuguese (Brazil)` or something.

PR #1133, opened 2019-10-23
Implement organization logo override; #803

### LINKED ISSUE #803: Organization logo override
#718, but for organization pages.

Ninjaclasher:
Fixes #803.


judge/models/profile.py
```diff
@@ -44,6 +44,10 @@ class Organization(models.Model):
                                             'only applicable to private organizations'))
     access_code = models.CharField(max_length=7, help_text=_('Student access code'),
                                    verbose_name=_('access code'), null=True, blank=True)
+    logo_override_image = models.CharField(verbose_name=_('Logo override image'), default='', max_length=150,
```
Xyene: I think this should be made editable in the non-admin organization editor.

judge/views/contests.py
```diff
@@ -167,6 +167,7 @@ def get_context_data(self, **kwargs):
         context['meta_description'] = self.object.summary or metadata[0]
         context['og_image'] = self.object.og_image or metadata[1]
         context['has_moss_api_key'] = settings.MOSS_API_KEY is not None
+        context['logo_override_image'] = self.object.logo_override_image
```
Xyene: It might be nice if (a) the contest is private to a single organization and (b) that organization has a logo override and (c) the contest doesn't override it, to use the organization's.

PR #1143, opened 2019-10-26
Fork django-sortedm2m to fix #1064

### LINKED ISSUE #1064: Edit Profile Page Broken
https://dmoj.ca/edit/profile/ is broken because jQuery is not being correctly loaded.

It appears that `jQuery` is defined while `$` is undefined, thus something is probably overriding `$`

Xyene:
A PR has been opened upstream at https://github.com/jazzband/django-sortedm2m/pull/149, but is unmerged currently. If/when it gets merged, we can switch back to upstream.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ django-pagedown<2
 django-registration-redux
 django-reversion
 django-social-share
--e git://github.com/DMOJ/django-sortedm2m#egg=django-sortedm2m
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -5,7 +5,7 @@ django-pagedown<2
 django-registration-redux
 django-reversion
 django-social-share
-django-sortedm2m
+-e git://github.com/DMOJ/django-sortedm2m#egg=django-sortedm2m
```
quantum5: It works without `.git`, but it looks better with it.

PR #1144, opened 2019-10-26
Remove batch deleting models; Fixes #335

### LINKED ISSUE #335: Disallow batch deleting User/Profile/Problem/Contest models
Yes, there is a confirmation dialog, but when "Delete selected users" is right next to "Rescore users", it is perfectly plausible that a tired admin could get past the confirmation view by reflex. There's really no need to batch delete such important models: in almost two years of running DMOJ, I don't think we've ever batch-deleted anything but submissions.

Carson-Tang:


judge/admin/contest.py
```diff
@@ -129,6 +129,12 @@ class ContestAdmin(VersionAdmin):
     filter_horizontal = ['rate_exclude']
     date_hierarchy = 'start_time'
 
+    def get_actions(self, request):
```
quantum5: Can we... make this a mixin?

PR #1147, opened 2019-10-27
Add a bit of padding to the footer.

ehhthing:
This has been bothering me for a while, but there's no padding for the footer, making the text touch the bottom of the window.

Preview:

image

### DIFF
--- a/resources/base.scss
+++ b/resources/base.scss
@@ -102,6 +102,7 @@ footer {
     width: 100%;
     position: absolute;
     bottom: 0;
+    padding-bottom: 10px;
 }
 
 body {


TheAvidDev:
This change is acceptable but requires an additional change within the footer template (in the admin settings). As of now, if this gets merged, it will look like this:
screenshot_20200510_221356

Removing `style="height: 1.5em"` from the `select` element of the localization form is required for better appearance:
screenshot_20200510_221412

Xyene:
This change has stalled, so I'm going to close it to free up the pending PR list. Feel free to reopen when you're ready.

PR #1151, opened 2019-10-27
Only set `is_pretested` if the `ContestProblem` also has pretests set

kiritofeng:
Otherwise, by setting the global pretest flag, problems that have `pretest_test_cases` in their `init.yml` will be run on pretests, regardless of the per-problem flag.

### DIFF
--- a/judge/judgeapi.py
+++ b/judge/judgeapi.py
@@ -61,8 +61,8 @@ def judge_submission(submission, rejudge, batch_rejudge=False):
     try:
         # This is set proactively; it might get unset in judgecallback's on_grading_begin if the problem doesn't
         # actually have pretests stored on the judge.
-        updates['is_pretested'] = ContestSubmission.objects.filter(submission=submission) \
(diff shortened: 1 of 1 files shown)


judge/judgeapi.py
```diff
@@ -61,8 +61,8 @@ def judge_submission(submission, rejudge, batch_rejudge=False):
     try:
         # This is set proactively; it might get unset in judgecallback's on_grading_begin if the problem doesn't
         # actually have pretests stored on the judge.
-        updates['is_pretested'] = ContestSubmission.objects.filter(submission=submission) \
-            .values_list('problem__contest__run_pretests_only', flat=True)[0]
+        updates['is_pretested'] = all(ContestSubmission.objects.filter(submission=submission)
+                                      .values_list('problem__contest__run_pretests_only', 'problem__is_pretested')[0])
```
quantum5: Why are you not using `flat=True`?

PR #1167, opened 2019-11-07
Reword help text in copy_language

Carson-Tang:


quantum5:
More descriptive title. Which help text? 🤔

PR #1173, opened 2019-11-08
Do not cascade when deleting Language; closes #1162

### LINKED ISSUE #1162: Language model deletion should not cascade
```
In [2]: Language.objects.filter(key__in=('FS', 'CS', 'VB', 'ERL', 'R')).delete()         
Out[2]: 
(75056,
 {'judge.OrganizationRequest': 8,
  'judge.SubmissionSource': 4008,
  'judge.SubmissionTestCase': 50362,
  'judge.Rating': 24,
  'judge.CommentVote': 160,
  'reversion.Version': 30,
  'judge.Ticket_assignees': 14,
  'judge.TicketMessage': 13,
  'judge.RuntimeVersion': 0,
  'judge.LanguageLimit': 0,
  'judge.Profile_organizations': 105,
  'judge.Problem_testers': 1,
  'judge.ContestSubmission': 630,
  'judge.ContestParticipation': 128,
  'judge.Comment': 23,
  'judge.Ticket': 9,
  'judge.Judge_runtimes': 4,
  'judge.Problem_allowed_languages': 15224,
  'judge.Submission': 4008,
  'judge.Profile': 300,
  'judge.Language': 5})
```

kiritofeng:


judge/models/runtime.py
```diff
@@ -94,6 +95,10 @@ def get_python2(cls):
     def get_absolute_url(self):
         return reverse('runtime_list') + '#' + self.key
 
+    @classmethod
+    def get_default_language(cls):
+        return Language.objects.get(key=settings.DEFAULT_USER_LANGUAGE).pk
```
quantum5: This function implies that it returns an object. Either rename or make use lambda or do something else.

kiritofeng: After Django 1.7, you cannot use a lambda for `default`, so I guess I'll rename it to `get_default_language_pk`

PR #1176, opened 2019-11-09
Use secure CSPRNG

### LINKED ISSUE #1175: Use a CSPRNG to generate judge keys
On https://dmoj.ca/admin/judge/judge/add/, the "regenerate" button show in the picture uses `Math.random` instead of a CSPRNG:

This disturbs me.

ehhthing:
Fixes #1175

### DIFF
--- a/judge/admin/runtime.py
+++ b/judge/admin/runtime.py
@@ -58,7 +58,7 @@ django.jQuery(document).ready(function ($) {{
         var length = 100,
             charset = "abcdefghijklnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`~!@#$%^&*()_+-=|[]{{}};:,<>./?",
             key = "",
-            random = new Uint8Array(length);
+            random = new Uint16Array(length);
 
         window.crypto.getRandomValues(random);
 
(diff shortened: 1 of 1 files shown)


judge/admin/runtime.py
```diff
@@ -57,9 +57,13 @@ def render(self, name, value, attrs=None, renderer=None):
...
             charset = "abcdefghijklnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789`~!@#$%^&*()_+-=|[]{{}};:,<>./?",
-            key = "";
+            key = "",
+            random = new Uint8Array(length);
+        
+        window.crypto.getRandomValues(random);
+
         for (var i = 0, n = charset.length; i < length; ++i) {{
-            key += charset.charAt(Math.floor(Math.random() * n));
+            key += charset.charAt(random[i] % n);
```
quantum5: This distribution is just wrong. The first 71 characters of the charset are 50% more likely to appear than the last 21 characters.

TheAvidDev:
Shouldn't something more standard like python's [secrets](https://docs.python.org/3/library/secrets.html) library be used?

ehhthing:
> Shouldn't something more standard like python's [secrets](https://docs.python.org/3/library/secrets.html) library be used?

I'm not exactly sure what you are suggesting here. This is fully client side, meaning that we can only use the JavaScript [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API) which is a standard.

quantum5:
Appears to have been superseded by #1757.

PR #1178, opened 2019-11-09
Fix more language stats chart UI issues related to ac8421a

Ninjaclasher:
In particular: 
 - Switch to Chart.js's legend instead of a custom legend, which supports filtering by labels, and is more robust.
 - Fix issue with showing all label names when only some would be present on chart (can be seen on the second and third charts on https://dmoj.ca/stats/language/).
 - Fix issue with the language AC rate chart cutting off if there are only a few languages, as `ctx.canvas.height = 20 * data.labels.length;` did not account for the axis labels.
Screenshot from 2019-11-09 14-29-02

### DIFF
--- a/judge/views/stats.py
+++ b/judge/views/stats.py
@@ -36,12 +36,11 @@ def repeat_chain(iterable):
 
 def language_data(request, language_count=Language.objects.annotate(count=Count('submission'))):
     languages = language_count.filter(count__gt=0).values('key', 'name', 'count').order_by('-count')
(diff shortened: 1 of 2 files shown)


quantum5:
Screenshot of final result?

Ninjaclasher:
Screenshot from 2019-11-09 15-11-10

Ninjaclasher:
First two charts:
Screenshot from 2019-11-09 15-30-10

Last two charts:
Screenshot from 2019-11-09 15-30-18

Ninjaclasher:
Updated the commit message

PR #1183, opened 2019-11-11
Automatically hide all descendant comments; fixes #1182

### LINKED ISSUE #1182: when a comment is hidden, automatically hide all descendant comments
The relevant code is here:
https://github.com/DMOJ/online-judge/blob/18816e798b0d8cb7c90d2fd4a1e06baf6532616b/judge/views/comment.py#L160

Ninjaclasher:
Fixes #1182.

### DIFF
--- a/judge/views/comment.py
+++ b/judge/views/comment.py
@@ -166,6 +166,5 @@ def comment_hide(request):
         return HttpResponseBadRequest()
 
     comment = get_object_or_404(Comment, id=comment_id)
-    comment.hidden = True
-    comment.save(update_fields=['hidden'])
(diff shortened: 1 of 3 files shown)


judge/views/comment.py
```diff
@@ -166,6 +166,7 @@ def comment_hide(request):
         return HttpResponseBadRequest()
 
     comment = get_object_or_404(Comment, id=comment_id)
+    comment.get_descendants().update(hidden=True)
```
Xyene: `get_descendants(include_self=True)` so there's 1 less query

Xyene:
Thanks!

PR #1202, opened 2019-11-29
Rescore contest when parameters change; closes #1157

### LINKED ISSUE #1157: Rescore contest when parameters change
Contest counterpart of #385, can use a similar approach.

kiritofeng:


judge/admin/contest.py
```diff
@@ -153,6 +153,10 @@ def has_change_permission(self, request, obj=None):
             return True
         return obj.organizers.filter(id=request.profile.id).exists()
 
+    def _rescore(self, request, contest_key):
```
Xyene: Probably shouldn't be in a separate fcn if there's a single caller and it's this short

judge/admin/contest.py
```diff
@@ -219,6 +223,11 @@ def get_form(self, *args, **kwargs):
         ).distinct()
         return form
 
+    def save_model(self, request, obj, form, change):
+        super(ContestAdmin, self).save_model(request, obj, form, change)
+        if form.changed_data and any(f in form.changed_data for f in ('format_config', 'format_name')):
```
Xyene: what about when problems or problem weights change? does anything need to be updated then?

PR #1203, opened 2019-12-01
Allow overflow on button row.

ehhthing:
This fixes many issues involving random size changes on mobile.

Before: 
image

After: 
image

(the bottom dark bar is just to illustrate the scroll bar)

### DIFF
--- a/resources/pagedown_widget.css
+++ b/resources/pagedown_widget.css
@@ -27,7 +27,7 @@
     margin: 10px 5px 5px;
     padding: 0;
     height: 20px;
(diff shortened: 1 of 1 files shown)


resources/pagedown_widget.css
```diff
@@ -27,6 +27,7 @@
     margin: 10px 5px 5px;
     padding: 0;
     height: 20px;
+    overflow-x: scroll;
```
quantum5: This looks really bad. Have you tried `overflow-x: auto`?

PR #1204, opened 2019-12-02
Expose `is_disqualified` in the contest API

Ninjaclasher:


judge/views/api/api_v1.py
```diff
@@ -74,6 +74,7 @@ def api_v1_contest_detail(request, contest):
                 'user': participation.username,
                 'points': participation.score,
                 'cumtime': participation.cumtime,
+                'disqualified': participation.is_disqualified,
```
Xyene: `is_disqualified`

PR #1210, opened 2019-12-09
Add DMOJ_SUBMISSION_LIMIT

kiritofeng:


judge/views/problem.py
```diff
@@ -532,7 +532,7 @@ def problem_submit(request, problem=None, submission=None):
         if form.is_valid():
             if (not request.user.has_perm('judge.spam_submission') and
                     Submission.objects.filter(user=profile, was_rejudged=False)
-                              .exclude(status__in=['D', 'IE', 'CE', 'AB']).count() > 2):
+                              .exclude(status__in=['D', 'IE', 'CE', 'AB']).count() > settings.DMOJ_SUBMISSION_LIMIT):
```
Xyene: Wrong, it's actually that - 1

Xyene: Or rather, >=

PR #1223, opened 2020-01-06
Decimal precision for user points in contest; Fixes #675

### LINKED ISSUE #675: contest ranking sort does not appear to sort by the exact number of points earned by contestant
Reproduction steps (copied from a production instance of this issue):

1) Generate a problem with the opportunity to get partial credit.
2) Insert the problem into a contest weighted such that the user can get a fractional number of points less than 1 in the weighting of the contest.
3) In the contest, score the number of points referenced in step 2.

Expected outcome:
This user outperforms other users who have scored zero points on the contest.

Actual outcome:
This user's ranking is strictly worse than other people who have scored zero points.

An anonymized scoreboard from production is provided showing exactly what this looks like.

screen shot 2018-04-15 at 12 12 00 pm

Carson-Tang:


judge/migrations/0109_contest_decimal_points.py
```diff
@@ -0,0 +1,32 @@
+# Generated by Django 2.2.7 on 2020-03-30 01:44
+
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def int_to_float(apps, schema_editor):
+    ContestParticipation = apps.get_model('judge', 'ContestParticipation')
+    for participation in ContestParticipation.objects.all():
```
Ninjaclasher, line 10:
```suggestion
    for participation in ContestParticipation.objects.select_related('contest').iterator():
```
Minor optimization.

judge/migrations/0109_contest_decimal_points.py
```diff
@@ -0,0 +1,32 @@
...
+        participation.recompute_results()
```
Ninjaclasher, line 11: Object methods don't exist in a migration - You'll need to copy the body of `recompute_results` over.

judge/migrations/0109_contest_decimal_points.py
```diff
@@ -0,0 +1,40 @@
+# Generated by Django 2.2.7 on 2020-03-30 01:44
+
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models, transaction
+
+
+def recompute_results(participation):
+    with transaction.atomic():
+        participation.contest.format.update_participation(participation)
```
Ninjaclasher, line 10: `format` is still a "method" (cached property), so it is not available here.

judge/migrations/0109_contest_decimal_points.py
```diff
@@ -0,0 +1,41 @@
...
+
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models, transaction
+from judge import contest_format
+
+
+def recompute_results(participation):
+    with transaction.atomic():
+        contest_format.formats[participation.format_name].update_participation(participation)
```
Ninjaclasher, line 11:
```
AttributeError: 'ContestParticipation' object has no attribute 'format_name'
```

Carson-Tang:
Calling `update_participation` on the existing contests doesn't work very well.

 `AttributeError: 'ContestParticipation' object has no attribute 'start'`, it should probably be done somewhere outside the migration?

Thanks to @Ninjaclasher for testing ❤️

Xyene (approved):
This looks good to me (extra good because it was tested :) -- could you rebase the migration? (Also nit while you're at it, `s/Fixes/fixes` in commit message for consistency)

PR #1225, opened 2020-01-09
dmoj/settings.py: Bump jquery to 3.4.1

outloudvi:
For whoever accidentally toggled `INLINE_JQUERY` off, we can bump jquery to 3.4.1 here as the same as whoever doesn't change the option [using jquery 3.4.1](https://github.com/DMOJ/online-judge/blob/0a724b4c06a274c19463e76e05734255060e797b/templates/base.html#L77).


Xyene:
Thanks!

PR #1228, opened 2020-01-12
Do not crash on errors in `display_user_problem`

kiritofeng:


ehhthing:
I've tested it, and it seems to be not broken.

judge/contest_format/atcoder.py
```diff
@@ -97,17 +97,22 @@ def update_participation(self, participation):
...
+                           self.best_solution_state(format_data['points'], contest_problem.points)),
+                    url=reverse('contest_user_submissions',
+                                args=[self.contest.key, participation.user.user.username,
+                                      contest_problem.problem.code]),
+                    points=floatformat(format_data['points']),
+                    penalty=penalty,
+                    time=nice_repr(timedelta(seconds=format_data['time']), 'noday'),
+                )
+            except (KeyError, TypeError, ValueError):
+                return mark_safe('<td>???</td>')
```
Xyene: A comment here about the bug this fixes would be nice

judge/contest_format/ioi.py
```diff
@@ -79,15 +79,19 @@ def update_participation(self, participation):
...
+                    state=(('pretest-' if self.contest.run_pretests_only and contest_problem.is_pretested else '') +
+                           self.best_solution_state(format_data['points'], contest_problem.points)),
+                    url=reverse('contest_user_submissions',
+                                args=[self.contest.key, participation.user.user.username,
+                                      contest_problem.problem.code]),
+                    points=floatformat(format_data['points']),
+                    time=nice_repr(timedelta(seconds=format_data['time']), 'noday') if self.config['cumtime'] else '',
+                )
+            except (KeyError, TypeError, ValueError):
+                return mark_safe('<td>???</td>')
```
Xyene: Rather than copying this over and over, perhaps it should be up to the caller of `display_user_problem`?

kiritofeng:
#1235 has a better implementation.

PR #1235, opened 2020-02-06
Do not crash on errors in `display_user_problem`

kiritofeng:


judge/views/contests.py
```diff
@@ -575,8 +583,7 @@ def make_contest_ranking_profile(contest, participation, contest_problems):
         cumtime=participation.cumtime,
         organization=user.organization,
         participation_rating=participation.rating.rating if hasattr(participation, 'rating') else None,
-        problem_cells=[contest.format.display_user_problem(participation, contest_problem)
-                       for contest_problem in contest_problems],
+        problem_cells=[display_user_problem(contest_problem) for contest_problem in contest_problems],
```
Xyene: This is just `map`, but ok

PR #1236, opened 2020-02-08
Allow contest organizers to see hidden scoreboard contest submissions

Ninjaclasher:
Users still cannot see their submissions for hidden scoreboard contests, as the join would still be messy.


Xyene:
Thanks!

PR #1240, opened 2020-03-01
Fix contest join tab width

Xyene:
Fixes 

image

to look like

image


quantum5:
How does it look when it's on the same line as the title?

PR #1246, opened 2020-03-14
Add support for sample cases

Ninjaclasher:


judge/contest_format/base.py
```diff
@@ -82,6 +82,17 @@ def get_problem_breakdown(self, participation, contest_problems):
         """
         raise NotImplementedError()
 
+    @abstractmethod
```
kiritofeng: It would be nice if we could configure this on a per-contest basis instead

PR #1247, opened 2020-03-15
view_contest_scoreboard user list on contest admin page

PeiAllen:
Added a view_contest_scoreboard list in contest settings to allow normal users to access private and hidden scoreboards.


judge/models/contest.py
```diff
@@ -68,6 +68,9 @@ class Contest(models.Model):
                                           help_text=_('Whether the scoreboard should remain hidden for the duration '
                                                       'of the contest.'),
                                           default=False)
+    see_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('See Contest Scoreboard'), blank=True,
```
Xyene: Verbose name should be all lowercase.

judge/models/contest.py
```diff
@@ -139,6 +142,8 @@ def can_see_scoreboard(self, user):
             return True
         if user.is_authenticated and self.organizers.filter(id=user.profile.id).exists():
             return True
+        if self.see_contest_scoreboard.filter(id=user.profile.id).exists():
```
Xyene: This needs the `user.is_authenticated` check as the line above, or we'll crash on an anonymous user who has no profile.

PR #1249, opened 2020-03-16
Mark all QU submissions as IE on bridge startup

### LINKED ISSUE #1208: Mark all QU submissions as IE on bridge startup
Brought up by @Ninjaclasher in Slack. The alternative is to re-queue submissions, but that + re-queuing has led to IE loops in the past, when the judge was less stable.

Ninjaclasher:
Fixes #1208.


Xyene:
:+1:

PR #1250, opened 2020-03-17
Use getattr to avoid mis-using Mathoid

outloudvi:
* `MATHOID_URL = False` is set in the default settings: [src](https://github.com/DMOJ/online-judge/blob/master/dmoj/settings.py#L90)
* False is also a valid property value so `hasattr(settings, 'MATHOID_URL')` will be a `True`, even when `MATHOID_URL = False`
* This will cause the MathoidMathParser to be called, even no valid Mathoid server is configured: [src](https://github.com/DMOJ/online-judge/blob/master/judge/jinja2/markdown/math.py#L52)

Using `getattr`, a `False` or `None` will be consider as falsy, thus solves the problem.


Xyene:
Thanks!

PR #1258, opened 2020-03-22
pagedown_math: $.load(handler) is deprecated

outloudvi:
In #1225 I proposed to upgrade the jQuery version of inline jQuery, and I'm experiencing:

```
e.indexOf is not a function
```

in pagedown_math.js. Finally I realized that `$.load(handler)` was [deprecated](https://api.jquery.com/load-event/) in jQuery 3.0. Therefore we also migrate this to a up-to-date jQuery API.


Xyene:
Looks good, thanks.

PR #1260, opened 2020-03-24
Implement ability to customize contest problem labels; #1124

### LINKED ISSUE #1124: Custom contest problem labels
Right now we hardcode "1", "2", "3", etc. Some contests (like ICPC) may wish to use different labels, like "A", "B", or "C". I'm in favour of allowing user to override by a `label` field in `ContestProblem`, and give the contest format the choice when that is null by some sort of `get_label_for_problem(n : int) -> str` function.

Ninjaclasher:
Closes #1124.


judge/models/contest.py
```diff
@@ -113,6 +115,10 @@ class Contest(models.Model):
                               help_text=_('A JSON object to serve as the configuration for the chosen contest format '
                                           'module. Leave empty to use None. Exact format depends on the contest format '
                                           'selected.'))
+    problem_label_script = models.TextField(verbose_name='contest problem label script', blank=True,
+                                            help_text='A custom LUA function to generate problem labels. Requires a '
+                                                      'single function with an integer paramater, the contest problem '
```
Xyene: typo in `parameter`

judge/models/contest.py
```diff
@@ -113,6 +115,10 @@ class Contest(models.Model):
                               help_text=_('A JSON object to serve as the configuration for the chosen contest format '
                                           'module. Leave empty to use None. Exact format depends on the contest format '
                                           'selected.'))
+    problem_label_script = models.TextField(verbose_name='contest problem label script', blank=True,
+                                            help_text='A custom LUA function to generate problem labels. Requires a '
```
Xyene: `Lua`

judge/models/contest.py
```diff
@@ -113,6 +115,10 @@ class Contest(models.Model):
...
+                                                      'single function with an integer paramater, the contest problem '
+                                                      '"order", and returns a string, the label.')
```
Xyene: 0-indexed or 1-indexed?

judge/models/contest.py
```diff
@@ -122,12 +128,29 @@ def format_class(self):
     def format(self):
         return self.format_class(self, self.format_config)
 
+    @cached_property
+    def get_label_for_problem(self):
+        def DENY_ALL(obj, attr_name, is_setting):
+            raise AttributeError()
+        lua = LuaRuntime(attribute_filter=DENY_ALL, register_eval=False, register_builtins=False)
+        return lua.eval(self.problem_label_script or settings.DMOJ_CONTEST_DEFAULT_PROBLEM_LABEL_SCRIPT)
```
Xyene: I think this should be queried from the contest format. All ICPC contests should have letters, for instance.

judge/models/contest.py
Ninjaclasher: This is dependant on the contest setter (if labels are generated with `problem.order`).

PR #1281, opened 2020-03-29
Do submission size check client side; #946

### LINKED ISSUE #946: do submission size check client-side
We can inspect the length of the code being submitted, so we can just perform the check client-side.

Ninjaclasher:
Fixes #946.

### DIFF
--- a/templates/problem/submit.html
+++ b/templates/problem/submit.html
@@ -110,7 +110,7 @@
 
                 $('#problem_submit').submit(function (event) {
                     if ($('#id_source').val().length > 65536) {
-                        alert("{{ _('Your source code must contain less than or equal to 65536 characters.') }}");
+                        alert("{{ _('Your source code must contain at most 65536 characters.') }}");
                         event.preventDefault();
                         $('#problem_submit').find(':submit').attr('disabled', false);
                     }
--- a/templates/problem/submit.html
+++ b/templates/problem/submit.html
@@ -107,6 +107,14 @@
                 $(window).resize(function () {
                     $('#ace_source').height(Math.max($(window).height() - 353, 100));
                 }).resize();
+
+                $('#problem_submit').submit(function (event) {
+                    if ($('#id_source').val().length > 65536) {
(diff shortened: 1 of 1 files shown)


templates/problem/submit.html
```diff
@@ -107,6 +107,14 @@
                 $(window).resize(function () {
                     $('#ace_source').height(Math.max($(window).height() - 353, 100));
                 }).resize();
+
+                $('#problem_submit').submit(function (event) {
+                    if ($('#id_source').val().length > 65536) {
+                        alert("{{ _('Your source code must contain less than or equal to 65536 characters.') }}");
```
Xyene: "at most"

templates/problem/submit.html
```diff
@@ -107,6 +107,14 @@
                 $(window).resize(function () {
                     $('#ace_source').height(Math.max($(window).height() - 353, 100));
                 }).resize();
+
+                $('#problem_submit').submit(function (event) {
+                    if ($('#id_source').val().length > 65536) {
```
quantum5: Hilariously, this doesn't count it the same way as the server. In JavaScript, `'😭'.length == 2`, but we count `😭` as one character on the server.

Ninjaclasher: One solution we can use is to use the code described [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/length#Unicode) with using `[...str].length` instead of `str.length`, though I'm not sure if there are any implications.

From what I can remember, the purpose of making the check client-side is to prevent the site from timing out when submitting a few megabytes of source code (leaving the user clueless as to why their code couldn't be submitted). Another solution could be to increase the client bound to something like 262144 characters, and if the source code is between 65536 and 262144 characters, let the server handle the error message (which the server can handle reasonably quickly).

Xyene: I don't think this matters much, and `[...str].length` is not yet that compatible (we don't transpile JavaScript down, anyway).

Issue #958, opened 2019-04-18
Add ACM contest format


quantum5:
We have wanted to support ACM-style contests for a long time. Now, given that #808 is introduced, we can actually write a module that implements ACM contests.

This supersedes all previous proposals on ACM-style contest implementations.

PR #1283, opened 2020-03-30
Add ACM contest format; #958

kiritofeng:


judge/contest_format/icpc.py
```diff
@@ -0,0 +1,114 @@
...
+                        penalty += prev * self.config['penalty'] * 60
+                    else:
+                        # We should always display the penalty, even if the user has a score of 0
+                        prev = subs.count()
+
+                if score:
+                    cumtime += dt
+                    last = max(last, dt)
+
+                format_data[str(prob)] = {'time': dt, 'points': score, 'penalty': prev}
```
Ninjaclasher, line 89: `prev` is not defined when `self.config['penalty'] = 0`

judge/contest_format/icpc.py
```diff
@@ -0,0 +1,114 @@
...
+from django.utils.translation import gettext_lazy
+
+from judge.contest_format.default import DefaultContestFormat
+from judge.contest_format.registry import register_contest_format
+from judge.timezone import from_database_time
+from judge.utils.timedelta import nice_repr
+
+
+@register_contest_format('icpc')
+class ICPCContestFormat(DefaultContestFormat):
```
Ninjaclasher, line 18: Should probably import `ICPCContestFormat` in `judge/contest_format/__init__.py`. Otherwise, it doesn't seem to show up in the contest admin.

judge/models/contest.py
```diff
@@ -346,6 +346,7 @@ class ContestParticipation(models.Model):
     cumtime = models.PositiveIntegerField(verbose_name=_('cumulative time'), default=0)
     is_disqualified = models.BooleanField(verbose_name=_('is disqualified'), default=False,
                                           help_text=_('Whether this participation is disqualified.'))
+    tiebreaker = models.FloatField(verbose_name=_('tie-breaking field'), default=0.0)
```
Ninjaclasher, line 349: `tiebreaker` should probably be included [here](https://github.com/DMOJ/online-judge/blob/master/judge/views/contests.py#L583) as well to sort the actual rankings. (Unless I'm misunderstanding the ICPC format)

Ninjaclasher:
Everything seems to work as intended aside from those few minor issues.

However, there are a few things that might need to be changed:
 - Should `tiebreaker` be set to 0 for the other contest formats? For example, suppose someone used this contest format and got a nonzero `tiebreaker` value. Then, they decided to switch formats (e.g IOI no cumtime). Their `tiebreaker` value would still be from the ICPC format, and now two users with the same score will be ranked differently. Since `tiebreaker` cannot be changed from the contest participation admin page, there would be no way to reset `tiebreaker` without access to a shell...
 - `update_participation` seems to do O(number of problem) queries to calculate the number of submissions to be included in the time penalty. Perhaps this could be moved into the raw SQL query?
 - Don't forget to rebase the migration.

kiritofeng:
1. Set `tiebreaker` to 0 for other contest formats to prevent oddities from breaking out
2. AtCoder also does O(# of problem) queries. I've been told that as long as it does not exceed O(# of problems it is fine, and I believe the equivalent raw SQL query would be quite disgusting to write
3. Good catch

judge/contest_format/icpc.py
```diff
@@ -0,0 +1,129 @@
...
+                if score:
+                    cumtime += dt
+                    last = max(last, dt)
+
+                format_data[str(prob)] = {'time': dt, 'points': score, 'penalty': prev}
+                points += score
+
+        participation.cumtime = cumtime + penalty
+        participation.score = points
+        participation.tiebreaker = -last  # field is sorted from greatest to least
```
quantum5, line 96: Why is tiebreaker even DESC? Usually ASC seems more natural as a default.

README.md
```diff
@@ -14,7 +14,7 @@ See it live at [dmoj.ca](https://dmoj.ca/)!
    * Specifying **per-language resource limits**
    * Capable of scaling to hundreds of judging servers
 * Extremely configurable contest system:
-   * Supports IOI/AtCoder/ECOO formats out-of-the box
+   * Supports IOI/AtCoder/ECOO/ACM formats out-of-the box
```
Ninjaclasher, line 17: Also need to add `ACM` to line 67 in the README.

README.md
```diff
@@ -64,7 +64,7 @@ Global, per-problem, and per-contest submission lists are live-updating, and can
 
 
 ### Extensible contest system
-Contests feature an optional rating system, and can be configured to run in any timeframe. Users are also able to participate virtually after the contest ends. IOI, AtCoder, and ECOO contest formats are supported out-of-the-box, and new formats can be added with custom code.
+Contests feature an optional rating system, and can be configured to run in any timeframe. Users are also able to participate virtually after the contest ends. IOI, AtCoder, ECOO, and ACM contest formats are supported out-of-the-box, and new formats can be added with custom code.
```
Xyene, line 67:
```suggestion
Contests feature an optional rating system, and can be configured to run in any timeframe. Users are also able to participate virtually after the contest ends. ACM-ICPC, IOI, AtCoder, and ECOO contest formats are supported out-of-the-box, and new formats can be added with custom code.
```

README.md
Xyene, line 17:
```suggestion
   * Supports ACM-ICPC/IOI/AtCoder/ECOO formats out-of-the box
```

PR #1289, opened 2020-03-31
404 contests that the user should not have access to in the API

Ninjaclasher:
Looks like this bug has been a thing since 789f8fcccc103de58c58447a012195459727b622...

Example: `olylvl7hw200328` is an organization private contest but an anonymous user can access https://dmoj.ca/api/contest/info/olylvl7hw200328.


Xyene:
Access control is truly a gift that keeps on giving.

Issue #483, opened 2016-10-07
Problem Types can be picked up by the API


Phoenix1369:
Relevant for problems which are part of a live contest as participants can discover the tags during the competition.

BinAccel:
Users can also leave a contest and look at the problems, which circumvents this.

quantum5:
The change here does not work when the contest is over, since it bans you from seeing the types after the contest is over.

On the other hand, after reworking the upsolving system in #473, leaving the contest would not be a problem.

PR #1291, opened 2020-04-01
Add API V2

Ninjaclasher:
- [x] Implement more filtering options.
- [x] Find a better way of handling `LoginRequiredMixin`. Currently, it redirects to a login page, which is not favourable.
- [x] Add organization list/~detail~ endpoint.
- [x] Add virtual contest participation endpoint.
- [x] Optimize the queries further (especially in `APISubmissionList`).

Additional PRs that need to be made before this PR can be completed:
 - [x] Create `Problem.get_visible_problems` classmethod
 - [x] #483.

### DIFF
--- a/judge/views/api/api_v2.py
+++ b/judge/views/api/api_v2.py
@@ -487,7 +487,7 @@ class APISubmissionList(APIListView):
         use_straight_join(queryset)
         join_sql_subquery(
             queryset,
-            subquery=Problem.get_visible_problems(self.request.user).only('id').query,
+            subquery=Problem.get_visible_problems(self.request.user).distinct().only('id').query,
             params=[],
             join_fields=[('problem_id', 'id')],
             alias='visible_problems',
--- a/judge/views/api/api_v2.py
+++ b/judge/views/api/api_v2.py
@@ -166,18 +166,30 @@ class APIContestDetail(APIDetailView):
         if contest.hide_scoreboard and in_contest:
             can_see_rankings = False
 
-        problems = list(contest.contest_problems.select_related('problem')
-                        .defer('problem__description').order_by('order'))
+        problems = list(
+            contest.contest_problems
+            .select_related('problem')
+            .defer('problem__description')
+            .order_by('order'),
+        )
 
         new_ratings_subquery = Rating.objects.filter(participation=OuterRef('pk'))
-        old_ratings_subquery = (Rating.objects.filter(user=OuterRef('user__pk'),
-                                                      contest__end_time__lt=OuterRef('contest__end_time'))
-                                .order_by('-contest__end_time'))
-        participations = (contest.users.filter(virtual=ContestParticipation.LIVE, user__is_unlisted=False)
-                          .annotate(new_rating=Subquery(new_ratings_subquery.values('rating')[:1]))
-                          .annotate(old_rating=Subquery(old_ratings_subquery.values('rating')[:1]))
-                          .annotate(username=F('user__user__username'))
-                          .order_by('-score', 'cumtime') if can_see_rankings else [])
+        old_ratings_subquery = (
+            Rating.objects
+            .filter(user=OuterRef('user__pk'), contest__end_time__lt=OuterRef('contest__end_time'))
+            .order_by('-contest__end_time')
+        )
+        participations = (
+            contest.users
+            .filter(virtual=ContestParticipation.LIVE, user__is_unlisted=False)
+            .annotate(
+                username=F('user__user__username'),
+                old_rating=Subquery(old_ratings_subquery.values('rating')[:1]),
+                new_rating=Subquery(new_ratings_subquery.values('rating')[:1]),
+            )
+            .order_by('-score', 'cumtime')
+            if can_see_rankings else []
+        )
         can_see_problems = (in_contest or contest.ended or contest.is_editable_by(self.request.user))
 
         return {
--- a/judge/views/api/api_v2.py
+++ b/judge/views/api/api_v2.py
@@ -401,6 +401,7 @@ class APIUserList(APIListView):
 
     def get_object_data(self, profile):
         return {
+            'id': profile.id,
             'username': profile.username,
             'points': profile.points,
             'performance_points': profile.performance_points,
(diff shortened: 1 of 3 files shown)


Ninjaclasher:
Marking this as ready for review because the rest of the TODO list are nice to haves.
 - `LoginRequiredMixin` redirecting is not a big issue because most people won't be accessing the API anonymously, or if they do, they will be redirected once and know.
 - Using `Problem.get_visible_problems` is simply for code cleanup. The places where the current code should be replaced with `Problem.get_visible_problems` once that method is made have been marked clearly.

The PR was tested relatively thoroughly on MCPT data on 4 different users (anonymous, normal user, staff, and superuser), so there should not be any issues with the code itself. However, there may still be some access control issues, which is why I'm marking this PR as review so more people can check.

Finally, I've created a [gist](https://gist.github.com/Ninjaclasher/a3016f46f6174b7f1822f05170fad729) containing what the API is supposed to return at each endpoint.

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,579 @@
...
+                    'max_submissions': problem.max_submissions or None,
+                    'name': problem.problem.name,
+                    'code': problem.problem.code,
+                } for problem in problems
+            ] if can_see_problems else [],
+            'rankings': [
+                {
+                    'user': participation.username,
+                    'score': participation.score,
+                    'cumtime': participation.cumtime,
```
Xyene, line 238: Might as well take the opportunity here to give this a better name than `cumtime`, so that it matches the rest.

Ninjaclasher: This has been fixed.

Ninjaclasher:
TODO:
- [ ] Expose submission batches (`init.yml` style)
- [ ] Filter by `is_active` for users list
- [ ] `submissions` -> `solved_problems`
- [ ] Expose custom runtime limits for problems
- [ ] Submission detail access logic (?)

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,593 @@
...
+        can_see_rankings = contest.can_see_scoreboard(self.request.user)
+        if contest.hide_scoreboard and in_contest:
+            can_see_rankings = False
+        can_see_problems = (in_contest or contest.ended or contest.is_editable_by(self.request.user))
+
+        problems = list(
+            contest.contest_problems
+            .select_related('problem')
+            .defer('problem__description')
+            .order_by('order'),
```
quantum5, line 185: Is the trailing comma necessary?

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,593 @@
...
+                    q |= Q(organizers=self.request.profile)
+                q |= Q(view_contest_scoreboard=self.request.profile)
+            visible_contests = visible_contests.filter(q)
+
+        return (
+            ContestParticipation.objects
+            .filter(virtual__gte=0, contest__in=visible_contests)
+            .annotate(contest_key=F('contest__key'), username=F('user__user__username'))
+            .order_by('id')
+            .only('score', 'cumtime', 'tiebreaker', 'is_disqualified', 'virtual')
```
quantum5, line 284: Might be more efficient to just use `values` and make `get_object_data` identity.

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,593 @@
...
+        return (
+            Profile.objects
+            .filter(is_unlisted=False, user__is_active=True)
+            .annotate(
+                username=F('user__username'),
+                latest_rating=Subquery(latest_rating_subquery.values('rating')[:1]),
+                latest_volatility=Subquery(latest_rating_subquery.values('volatility')[:1]),
+            )
+            .order_by('id')
+            .only('points', 'performance_points', 'problem_count', 'display_rank')
```
quantum5, line 401: Again, maybe values is better.

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,593 @@
...
+            'volatility': last_rating.volatility if last_rating is not None else None,
+            'organizations': list(profile.organizations.values_list('id', flat=True)),
+            'contests': contest_history,
+        }
+
+
+class APISubmissionList(APIListView):
+    model = Submission
+    basic_filters = (
+        ('user', 'user__user__username'),
```
quantum5, line 477: I feel like it's more efficient to run a query to lookup the profile than joining stuff into submissions here, but that's just a feeling.

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,593 @@
...
+        queryset = Submission.objects.all()
+        use_straight_join(queryset)
+        join_sql_subquery(
+            queryset,
+            subquery=Problem.get_visible_problems(self.request.user).query,
+            params=[],
+            join_fields=[('problem_id', 'id')],
+            alias='visible_problems',
+        )
+        queryset = (
```
quantum5, line 495: Just return directly?

judge/views/api/api_v2.py
Ninjaclasher: It seems it is necessary for flake to pass.

Ninjaclasher:
All requested changes have been made (if they made the queries faster).

quantum5:
For `LoginRequiredMixin`, I think you should not use it. Instead, have a different version that returns 403 if you are not logged in.

Ninjaclasher:
A new `APILoginRequiredMixin` has been added.

judge/views/api/api_v2.py
```diff
@@ -1,122 +1,613 @@
...
+        caught_exceptions = {
+            ValueError: (400, 'invalid filter value type'),
+            ValidationError: (400, 'invalid filter value type'),
+            PermissionDenied: (403, 'permission denied'),
+            APILoginRequiredException: (403, 'login required'),
+            Http404: (404, 'page/object not found'),
+        }
+        exception_type = type(exception)
+        if exception_type in caught_exceptions:
+            exception_data = caught_exceptions[exception_type]
```
quantum5, line 63: Why not tuple unpack this?

Ninjaclasher: This has been updated in the latest commit.

PR #1293, opened 2020-04-02
Autoformat all code with Black

Xyene:
```
black . --line-length 120 --skip-string-normalization --exclude 'judge/migrations/*'
```


TheAvidDev:
Once this is ready, a `pyproject.toml` should be added so running `black` uses the appropriate settings.

Xyene:
Super stale, closing.

PR #1302, opened 2020-04-05
Import latest Can I use... data

dmoj-build:
This PR has been auto-generated to pull in the latest copy of Can I use... `data.json` from [upstream](https://github.com/Fyrd/caniuse/blob/master/data.json).


quantum5 (approved):
Let's update this file since it's been a year...

PR #1305, opened 2020-04-05
Add support for Selenium pdf generation

TheAvidDev:


judge/pdf_problems.py
```diff
@@ -9,6 +10,11 @@
 
 from django.conf import settings
 from django.utils.translation import gettext
+from selenium import webdriver
```
Xyene, line 13: This will just crash on startup if the user doesn't have selenium.

judge/pdf_problems.py
```diff
@@ -17,8 +23,10 @@
 PUPPETEER_MODULE = settings.PUPPETEER_MODULE
 HAS_PUPPETEER = os.access(NODE_PATH, os.X_OK) and os.path.isdir(PUPPETEER_MODULE)
 
+HAS_SELENIUM = settings.SELENIUM
```
Xyene, line 26: Weird name for `settings.SELENIUM`; doesn't indicate it's a bool.

Xyene: Probably want to use the presence of the actual paths themselves.

judge/pdf_problems.py
```diff
@@ -259,11 +267,67 @@ def _make(self, debug):
...
+            f.write(base64.b64decode(self.response['data']))
+
+
 if HAS_PUPPETEER:
     DefaultPdfMaker = PuppeteerPDFRender
 elif HAS_SLIMERJS:
     DefaultPdfMaker = SlimerJSPdfMaker
 elif HAS_PHANTOMJS:
     DefaultPdfMaker = PhantomJSPdfMaker
+elif HAS_SELENIUM:
```
Xyene, line 330: Probably want to give this higher priority than Slimer.

judge/pdf_problems.py
TheAvidDev: Selenium will automatically look in path, so they work perfectly with with `None` as their values, hence the use of a boolean at all. Is `USE_SELENIUM` better?

judge/pdf_problems.py
TheAvidDev: Fixed

judge/pdf_problems.py
```diff
@@ -259,8 +269,48 @@ def _make(self, debug):
...
+        options.add_argument("--disable-gpu")
+        options.binary_location = settings.SELENIUM_CUSTOM_CHROME_PATH
+
+        browser = webdriver.Chrome(settings.SELENIUM_CHROMEDRIVER_PATH, options=options)
+        browser.get("file://" + os.path.abspath(os.path.join(self.dir, 'input.html')))
+
+        try:
+            WebDriverWait(browser, 15).until(EC.presence_of_element_located((By.CLASS_NAME, 'math-loaded')))
+        except TimeoutException:
+            logger.error('PDF math rendering timed out')
```
Xyene, line 296: This should be logged to `self.log` as well, as should all other exceptions.

TheAvidDev: Added and formatted

judge/pdf_problems.py
```diff
@@ -259,8 +269,55 @@ def _make(self, debug):
...
+
+    def _make(self, debug):
+        options = webdriver.ChromeOptions()
+        options.add_argument("--headless")
+        options.add_argument("--disable-gpu")
+        options.binary_location = settings.SELENIUM_CUSTOM_CHROME_PATH
+
+        browser = webdriver.Chrome(settings.SELENIUM_CHROMEDRIVER_PATH, options=options)
+        browser.get("file://" + os.path.abspath(os.path.join(self.dir, 'input.html')))
+        self.log = browser.get_log('driver')
```
quantum5, line 292: Is there a need to store the list version in `self.log`? I think the list should be a local variable if possible to avoid `self.log` being a type union. While types don't matter now, I think it's only time before we introduce mypy.

TheAvidDev: Fixed. `self.log` now only contains the string version, however I'm not sure if that is the case for the other pdf generator/renderer logs.

TheAvidDev:
Added SeleniumPDFRender to `render_pdf` manage.py command.

TheAvidDev:
Important note: this will *not* work if being run under the `root` account because chrome (or chromium) refuses to launch with a sandbox:
```
[12548:12548:0409/040242.574915:ERROR:zygote_host_impl_linux.cc(89)] Running as root without --no-sandbox is not supported. See https://crbug.com/638180.
```
This also throws the following error in Selenium:
```
selenium.common.exceptions.WebDriverException: Message: unknown error: Chrome failed to start: exited abnormally.
  (unknown error: DevToolsActivePort file doesn't exist)
```
The "solution" is to add the following line in the `judge/pdf_problems.py` `SeleniumPDFRender` class:
```
browser.add_argument("--no-sandbox")
```
Two options to fix this are to either add a setting in `settings.py` or to add a check in python to see under what account it is running. Which one is favored?

PR #1308, opened 2020-04-06
Move problem filtering into class methods

Ninjaclasher:
Similar PR as #1263.

Note that unlike that #1263, this PR has some changes to problem visibilities:
 - [judge/models/profile.py](https://github.com/DMOJ/online-judge/compare/master...Ninjaclasher:problem-filter?expand=1#diff-2a2fcc9e01812a880588bc5c5e8f36b6R147) - Before, all `is_public=True` problems were included, even if they were organization private. This has been changed to `is_public=True, is_organization_private=False`, as it feels like a bug.
 - [judge/views/blog.py](https://github.com/DMOJ/online-judge/compare/master...Ninjaclasher:problem-filter?expand=1#diff-fcb42be28f0a0164573dbcf8b8b48f77R55) - Similar modification as above.
 - [judge/views/problem.py](https://github.com/DMOJ/online-judge/compare/master...Ninjaclasher:problem-filter?expand=1#diff-76c07cd9466d20d4489fb6cc3fa8a0adR363) - Before, only problems that were public or where the user was explicitly set as accessible were visible on the problems list. Now, _all_ visible problems are shown. **Cannot tell if this is a bug or a feature.**
 - [judge/views/select2.py](https://github.com/DMOJ/online-judge/compare/master...Ninjaclasher:problem-filter?expand=1#diff-e5e7ff9f4e4a093600d0999663dd30c0R57) - Before, any public problem or problem where the user was author or curator were visible here, even if they were organization private. This seems like a bug, and has been changed to any visible problem.

### DIFF
--- a/judge/feed.py
+++ b/judge/feed.py
@@ -15,7 +15,7 @@ class ProblemFeed(Feed):
     description = 'The latest problems added on the %s website' % settings.SITE_LONG_NAME
 
     def items(self):
-        return Problem.objects.filter(is_public=True, is_organization_private=False).order_by('-date', '-id')[:25]
+        return Problem.get_public_problems().order_by('-date', '-id')[:25]
 
     def item_title(self, problem):
         return problem.name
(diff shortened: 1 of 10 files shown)


Ninjaclasher:
Note that depending on whether this or #1310 is merged first, `Problem.get_visible_problems` will need to be modified to account for `judge.edit_organization_problem`.

PR #1317, opened 2020-04-10
Convert bridge to use threads

quantum5:
`event_socket_server` is basically broken because it expects MySQL to be instantaneous, but it's not. As a result, despite the supposed async processing, it's actually blocked on MySQL basically all the time.

We are going to switch to a threaded-approach, which we will then convert to greenlet with gevent.

Outstanding issues:

* [x] not responding to a ping does not DC judge
* [x] judge disappearing during grading leads to IE, but admins are not notified


judge/bridge/base_handler.py
```diff
@@ -0,0 +1,195 @@
...
+        return self.request.gettimeout()
+
+    @timeout.setter
+    def timeout(self, timeout):
+        self.request.settimeout(timeout or None)
+
+    def read_sized_packet(self, size, initial=None):
+        if size > MAX_ALLOWED_PACKET_SIZE:
+            logger.log(logging.WARNING if self._got_packet else logging.INFO,
+                       'Disconnecting due to large size (%d): %s', size, self.client_address)
```
Xyene, line 74:
```suggestion
                       'Disconnecting client due to too-large message size (%d bytes): %s', size, self.client_address)
```

judge/bridge/base_handler.py
```diff
@@ -0,0 +1,195 @@
...
+            else:
+                self.read_sized_packet(tag)
+
+            while True:
+                self.read_sized_packet(self.read_size())
+        except Disconnect:
+            return
+        except zlib.error:
+            if self._got_packet:
+                logger.warning('Found zlib error in connection', exc_info=True)
```
Xyene, line 173:
```suggestion
                logger.warning('Encountered zlib error during packet handling, disconnecting client', exc_info=True)
```

judge/bridge/daemon.py
```diff
@@ -0,0 +1,46 @@
...
+logger = logging.getLogger('judge.bridge')
+
+
+def reset_judges():
+    Judge.objects.update(online=False, ping=None, load=None)
+
+
+def judge_daemon():
+    reset_judges()
+    Submission.objects.filter(status__in=Submission.IN_PROGRESS_GRADING_STATUS).update(status='IE', result='IE')
```
Xyene, line 23:
```suggestion
    Submission.objects.filter(status__in=Submission.IN_PROGRESS_GRADING_STATUS).update(status='IE', result='IE', error=None)
```

judge/bridge/base_handler.py
quantum5: Actually this doesn't say which client it is...

PR #1332, opened 2020-04-21
Make selenium imports determine whether it is used

TheAvidDev:
Right now, selenium *must* import successfully if the `USE_SELENIUM` variable is set, raising an `ImportError` otherwise. However, this also poses an issue if different parts of the site are running from different clones of the repository. It requires all parts, such as the bridge, to have selenium installed, unless separate `local_settings.py` files are used.

This PR sets the `HAS_SELENIUM` variable based off of whether the importing of `selenium` has succeeded or failed, if `USE_SELENIUM` is set to `True`.

### DIFF
--- a/judge/pdf_problems.py
+++ b/judge/pdf_problems.py
@@ -11,6 +11,8 @@ import uuid
 from django.conf import settings
 from django.utils.translation import gettext
 
+logger = logging.getLogger('judge.problem.pdf')
+
 HAS_SELENIUM = False
 if settings.USE_SELENIUM:
     try:
@@ -20,8 +22,8 @@ if settings.USE_SELENIUM:
         from selenium.webdriver.support import expected_conditions as EC
         from selenium.webdriver.support.ui import WebDriverWait
         HAS_SELENIUM = True
-    except ImportError:
-        pass
+    except ImportError as e:
+        logger.warning(e, exc_info=True)
 
 HAS_PHANTOMJS = os.access(settings.PHANTOMJS, os.X_OK)
 HAS_SLIMERJS = os.access(settings.SLIMERJS, os.X_OK)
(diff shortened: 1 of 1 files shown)


judge/pdf_problems.py
```diff
@@ -13,12 +13,15 @@
...
-    HAS_SELENIUM = True
+    try:
+        from selenium import webdriver
+        from selenium.common.exceptions import TimeoutException
+        from selenium.webdriver.common.by import By
+        from selenium.webdriver.support import expected_conditions as EC
+        from selenium.webdriver.support.ui import WebDriverWait
+        HAS_SELENIUM = True
+    except ImportError:
+        pass
```
Xyene, line 24:
```suggestion
        HAS_SELENIUM = False
```

probably also want to print the stacktrace

judge/pdf_problems.py
```diff
@@ -13,12 +13,15 @@
 
 HAS_SELENIUM = False
```
Xyene, line 14:
```suggestion
```

TheAvidDev: Decided against because one may not have `USE_SELENIUM` set to `True`.

judge/pdf_problems.py
TheAvidDev: Import failure is now logged. It's also done as a warning (as opposed to error) because this can happen under certain normal circumstances.

judge/pdf_problems.py
```diff
@@ -11,14 +11,19 @@
...
-    HAS_SELENIUM = True
+    try:
+        from selenium import webdriver
+        from selenium.common.exceptions import TimeoutException
+        from selenium.webdriver.common.by import By
+        from selenium.webdriver.support import expected_conditions as EC
+        from selenium.webdriver.support.ui import WebDriverWait
+        HAS_SELENIUM = True
+    except ImportError:
+        logger.warning('Failed to import Selenium')
```
Xyene, line 26: You need `exc_info=True` in order for the details of the exception (the useful part here) to be logged out.

PR #1334, opened 2020-04-23
Add scratch code support for 2FA; fixes #784

### LINKED ISSUE #784: Scratch code support for Two Factor Authentication
Currently, we expect people who are locked out of 2FA to contact admins, which will then manually disable 2FA. This works when we actually know the people involved, as identity is an important part of authentication.

For a more general solution, we should implement scratch codes. These codes should be much longer and harder to guess than standard 2FA codes.

This is what we roughly expect from the system:

* code is shown once upon generation, and never shown again
* we offer 5 codes at a time
* using a code invalidates it
* regenerating codes invalidates all old codes
* regeneration button shown on edit profile, when 2FA is enabled
* codes are also generated and displayed after enabling 2FA
* easy code copy button, which adds surrounding text as well to describe what the codes do
* perhaps a print button to generate a PDF (security concerns? at least we can block on phantomjs now that we have gevent)

Carson-Tang:
image
image
image
image


Carson-Tang:
Fixed font 
image

Carson-Tang:
Screenshots of latest commit:
image
image
image
image

judge/forms.py
```diff
@@ -1,6 +1,7 @@
 import json
 from operator import attrgetter
 
+
```
Xyene, line 4:
```suggestion
```

judge/forms.py
```diff
@@ -20,6 +21,22 @@
...
+
+def totp_or_scratch_code_validator(totp_or_scratch_code):
+    err = None
+    for validator in totp_or_scratch_code_validators:
+        try:
+            validator(totp_or_scratch_code)
+            return totp_or_scratch_code
+        except ValidationError as error:
+            err = error
```
Xyene, line 37: Won't this basically always return the `'Scratch codes must be 16 base32 characters.'` error? Any input I can think of that fails the 2fa validator will also fail the scratch code validator, unless it's an actually valid scratch code.

judge/models/profile.py
```diff
@@ -127,6 +129,10 @@ class Profile(models.Model):
                                                             _('API token must be None or hexadecimal'))])
     notes = models.TextField(verbose_name=_('internal notes'), null=True, blank=True,
                              help_text=_('Notes for administrators regarding this user.'))
+    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('Scratch Codes'),
```
Xyene, line 132: Let's move this up to beside the `totp_key` field, since they're logically a group.

judge/models/profile.py
```diff
@@ -174,6 +180,14 @@ def generate_api_token(self):
 
     generate_api_token.alters_data = True
 
+    def generate_scratch_codes(self):
+        codes = [pyotp.random_base32(length=16) for i in range(5)]
+        self.scratch_codes = json.dumps(codes)
+        self.save(update_fields=['scratch_codes'])
+        return '\n'.join(codes)
```
Xyene, line 187: I'd prefer returning the list of codes, and then looping over them in the template rather than joining them here. (Something about separating UI from backend logic; the frontend could conceivably someday want to show these as bullet points.)

templates/registration/totp_enable.html
```diff
@@ -61,6 +65,52 @@
...
+
+                curClipboard.on('success', function (e) {
+                    e.clearSelection();
+                    showTooltip(e.trigger, 'Copied!');
+                });
+
+                curClipboard.on('error', function (e) {
+                    showTooltip(e.trigger, fallbackMessage(e.action));
+                });
+
```
Xyene, line 110:
```suggestion
```

templates/registration/totp_enable.html
```diff
@@ -61,6 +65,52 @@
...
+                    }).text('Copy')));
+
+                $(copyButton.get(0)).mouseleave(function () {
+                    $(this).attr('class', 'btn-clipboard');
+                    $(this).removeAttr('aria-label');
+                });
+
+                var curClipboard = new Clipboard(copyButton.get(0));
+
+                curClipboard.on('success', function (e) {
```
Xyene, line 102:
```suggestion
                curClipboard.on('success', function(e) {
```

Only no-arg functions have a space before the `(`).

templates/user/edit-profile.html
```diff
@@ -177,6 +201,78 @@
...
+
+                    $.ajax({
+                        type: 'POST',
+                        url: $(this).attr('href'),
+                        dataType: 'json',
+                        success: function(data) {
+                            $('#scratch-codes').text(data.data.codes);
+                            $('#scratch-codes-copy-button').attr('data-clipboard-text', data.data.codes);
+                            $('#generate-scratch-codes-button').text("{{ _('Regenerate') }}");
+                            $('#scratch-codes-text').text("{{ _('Below is a list of one-time use scratch codes.') }} "
```
Xyene, line 264: Let's use a single `_('...')` block here too, so that we don't have to translate all these strings twice.

judge/models/profile.py
Xyene, line 132:
```suggestion
    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('scratch codes'),
```

Django handles capitalization in an i18n fashion.

templates/user/edit-profile.html
```diff
@@ -85,7 +98,17 @@
             $('#disable-2fa-button').click(function () {
                 alert("The administrators for this site require all the staff to have Two-factor Authentication enabled, so it may not be disabled at this time.");
             });
-
+        });
+    </script>
+    <script type="text/javascript">
+        $(function () {
+            $('#generate-scratch-codes-button').click(function() {
```
Xyene, line 105:
```suggestion
            $('#generate-scratch-codes-button').click(function () {
```

templates/user/edit-profile.html
```diff
@@ -262,6 +357,24 @@
...
+                                {{ _('Scratch Codes:') }}
+                                {% if profile.scratch_codes %}
+                                    <a id="generate-scratch-codes-button" href="{{ url('generate_scratch_codes') }}"
+                                       class="button inline-button">{{ _('Regenerate') }}</a>
+                                    <div class="scratch-codes-text" id="scratch-codes-text">{{ _('Hidden') }}</div>
+                                    <pre><code id="scratch-codes"></code></pre>
+                                {% else %}
+                                    <a id="generate-scratch-codes-button" href="{{ url('generate_scratch_codes') }}"
+                                       class="button inline-button">{{ _('Generate') }}</a>
+                                    <div class="scratch-codes-text" id="scratch-codes-text"></div>
```
Xyene, line 372: I don't see the reason for having both the class and id to be `scratch-codes-text` here... either would work, but only one is needed.

templates/user/edit-profile.html
```diff
@@ -85,7 +98,17 @@
             $('#disable-2fa-button').click(function () {
                 alert("The administrators for this site require all the staff to have Two-factor Authentication enabled, so it may not be disabled at this time.");
             });
-
+        });
+    </script>
+    <script type="text/javascript">
+        $(function () {
+            $('#generate-scratch-codes-button').click(function() {
+                $('#scratch-codes-text').css({"font-family": "Segoe UI, Lucida Grande, Arial, sans-serif"});
```
Xyene, line 106: This is a bit of a hack... can't we put this in another div instead?

templates/user/edit-profile.html
```diff
@@ -177,6 +200,78 @@
...
+                        $(copyButton.get(0)).mouseleave(function () {
+                            $(this).attr('class', 'btn-clipboard');
+                            $(this).removeAttr('aria-label');
+                        });
+
+                        var curClipboard = new Clipboard(copyButton.get(0));
+
+                        curClipboard.on('success', function(e) {
+                            e.clearSelection();
+                            showTooltip(e.trigger, 'Copied!');
```
Xyene, line 247: i18n

templates/user/edit-profile.html
```diff
@@ -177,6 +200,78 @@
             });
         });
     </script>
+    <script src="{{ static('libs/clipboard/clipboard.js') }}"></script>
+    <script src="{{ static('libs/clipboard/tooltip.js') }}"></script>
+    <script type="text/javascript">
+        $(function () {
+            $('#generate-scratch-codes-button').click(function (event) {
+                event.preventDefault();
+                if (confirm("{{ _('Are you sure you want to generate or regenerate a new set of scratch codes?') }}\n"
```
Xyene, line 209: Batch these into one message like the other multi-line messages so translators have more context.

templates/registration/totp_enable.html
```diff
@@ -61,6 +65,51 @@
...
+                    fix_div(info_float, 55);
+                    $(window).resize(function () {
+                        info_float.width(container.width());
+                    });
+                    info_float.width(container.width());
+                }
+            }
+
+            var copyButton;
+            $('pre code').each(function () {
```
Xyene, line 87: Having this code duplicated twice is a bit yucky, can we merge this with the clipboard code we have everywhere else?

templates/user/edit-profile.html
```diff
@@ -177,6 +200,78 @@
...
+                          + "{{ _('This will invalidate any previous scratch codes you have.') }} "
+                          + "{{ _('You will not be able to view your scratch codes after you leave this page!') }}")) {
+                    $('#scratch-codes').text("{{ _('Generating...') }}");
+
+                    var info_float = $('.info-float');
```
Xyene, line 214: I don't think we have an info float on this page, do we?

judge/models/profile.py
```diff
@@ -174,6 +180,14 @@ def generate_api_token(self):
 
     generate_api_token.alters_data = True
 
+    def generate_scratch_codes(self):
+        codes = [pyotp.random_base32(length=16) for i in range(5)]
```
kiritofeng, line 184: Should we make the number of scratch codes an adjustable constant?

templates/user/edit-profile.html
Carson-Tang: One is used for the text that appears after the `generate/regenerate` button is pressed and the other is to match the font used for `Hidden`.

templates/user/edit-profile.html
Carson-Tang: Another div...which replaces the current `scratch-codes-text` on click?

templates/user/edit-profile.html
```diff
@@ -177,6 +199,65 @@
...
     </script>
+    <script src="{{ static('libs/clipboard/clipboard.js') }}"></script>
+    <script src="{{ static('libs/clipboard/tooltip.js') }}"></script>
+    <script type="text/javascript">
+        $(function () {
+            $('#generate-scratch-codes-button').click(function (event) {
+                event.preventDefault();
+                if (confirm("{{ _('Are you sure you want to generate or regenerate a new set of scratch codes? ') }}" +
+                            "{{ _('This will invalidate any previous scratch codes you have. ') }}" +
+                            "{{ _('You will not be able to view your scratch codes after you leave this page!') }}")) {
```
Carson-Tang, line 210: Changed back to using string concatenation instead of escaping newlines since there were large spaces between each sentence.

judge/forms.py
```diff
@@ -164,20 +169,29 @@ def widget_attrs(self, widget):
...
         self.profile = kwargs.pop('profile')
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
+        if totp_or_scratch_code_len not in [6, 16]:
```
Ninjaclasher, line 182:
```suggestion
        if totp_or_scratch_code_len not in (6, 16):
```

judge/forms.py
```diff
@@ -164,20 +169,29 @@ def widget_attrs(self, widget):
...
+            raise ValidationError(_('Invalid code length.'))
+        else:
+            validator = totp_or_scratch_code_validators[not totp_or_scratch_code_len == 6]
+            try:
+                validator(totp_or_scratch_code)
+            except ValidationError as err:
```
Ninjaclasher, line 188: Doesn't `validator` raise a `ValidationError` when validation fails by default? Is there any point in catching the validation error and then reraising?

templates/registration/totp_enable.html
```diff
@@ -42,10 +42,15 @@
...
+
+        .scratch-codes {
+            white-space: pre-line;
+        }
     </style>
 {% endblock %}
 
-{% block js_media %}
+{% block content_js_media %}
+    {{ super() }}
```
Ninjaclasher, line 53: Isn't `common-content.html`'s `content_js_media` empty? (and also I don't think any other templates that extend from `common-content.html` have a similar `super()` call)

judge/views/two_factor.py
```diff
@@ -54,6 +54,10 @@ def get(self, request, *args, **kwargs):
         if not profile.totp_key:
             profile.totp_key = pyotp.random_base32(length=32)
             profile.save()
+        if not profile.scratch_codes:
+            profile.scratch_codes = json.dumps([pyotp.random_base32(length=16)
+                                                for i in range(settings.DMOJ_SCRATCH_CODES_COUNT)])
+            profile.save()
```
Ninjaclasher, line 60: Not sure if we should be saving the profile twice (once for the `totp_key` and again for the `scratch_codes`). Maybe have a list of `update_fields` and call `profile.save(update_fields=update_fields)` (since having an empty list automatically skips the save).

There might be a better way to do this that I can't think of.

judge/forms.py
```diff
@@ -164,20 +169,29 @@ def widget_attrs(self, widget):
...
     def clean(self):
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
+        if totp_or_scratch_code_len not in [6, 16]:
+            raise ValidationError(_('Invalid code length.'))
+        else:
+            validator = totp_or_scratch_code_validators[not totp_or_scratch_code_len == 6]
```
Ninjaclasher, line 185: `not totp_or_scratch_code_len == 6` feels like a strange way of finding the correct validator. Maybe make the validator list a dictionary of `{len: validator}`.

judge/forms.py
```diff
@@ -164,20 +169,26 @@ def widget_attrs(self, widget):
...
         self.profile = kwargs.pop('profile')
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
+        if totp_or_scratch_code_len not in (6, 16):
```
TheAvidDev, line 182:
```suggestion
        if totp_or_scratch_code_len not in totp_or_scratch_code_validators:
```

judge/forms.py
```diff
@@ -164,20 +169,26 @@ def widget_attrs(self, widget):
...
+            raise ValidationError(_('Invalid code length.'))
+        else:
+            validator = totp_or_scratch_code_validators[totp_or_scratch_code_len]
+            validator(totp_or_scratch_code)
+        if totp_or_scratch_code_len == 6 and not pyotp.TOTP(self.profile.totp_key).verify(totp_or_scratch_code,
```
TheAvidDev, line 187: Now that you're storing the length and validator in `totp_or_scratch_code_validators` it feels worse to hardcode the lengths here again. Perhaps the validation check can also be part of `totp_or_scratch_code_validators`?

templates/user/edit-profile.html
```diff
@@ -177,6 +199,65 @@
...
+                        url: $(this).attr('href'),
+                        dataType: 'json',
+                        success: function(data) {
+                            $('#scratch-codes').text(data.data.codes.join('\n'));
+                            $('#scratch-codes-copy-button').attr('data-clipboard-text', data.data.codes.join('\n'));
+                            $('#generate-scratch-codes-button').text("{{ _('Regenerate') }}");
+                            $('.hidden-word').empty();
+                            $('#scratch-codes-text').text("{{ _('Below is a list of one-time use scratch codes. \
+                                                                 These codes can only be used once and are for emergency use. \
+                                                                 You can use these codes to login to your account or disable Two-factor Authentication. \
```
Xyene, line 251:
```suggestion
                                                                 You can use these codes to login to your account or disable two-factor authentication. \
```

templates/registration/totp_enable.html
```diff
@@ -74,17 +79,26 @@
...
                 </div>
             {% endif %}
             <div>
                 <label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app:') }}</label>
-                <span class="fullwidth">{{ form.totp_token }}</span>
+                <span class="fullwidth">{{ form.totp_or_scratch_code }}</span>
             </div>
             <button style="margin-top: 0.5em" class="button" type="submit">{{ _('Enable Two Factor Authentication') }}</button>
+            <div class="scratch-codes">
+                <label class="inline-header grayed">{{ _('Below is a list of one-time use scratch codes.
```
kiritofeng, line 94: Does this multi-line string render correctly?

Xyene: Apparently yes.

judge/views/two_factor.py
```diff
@@ -51,9 +51,15 @@ class TOTPEnableView(TOTPView):
 
     def get(self, request, *args, **kwargs):
         profile = self.profile
+        update_fields = []
         if not profile.totp_key:
             profile.totp_key = pyotp.random_base32(length=32)
-            profile.save()
+            update_fields.append("totp_key")
```
Ninjaclasher, line 57:
```suggestion
            update_fields.append('totp_key')
```

judge/views/two_factor.py
```diff
@@ -51,9 +51,15 @@ class TOTPEnableView(TOTPView):
...
+        if not profile.scratch_codes:
+            profile.scratch_codes = json.dumps([pyotp.random_base32(length=16)
+                                                for i in range(settings.DMOJ_SCRATCH_CODES_COUNT)])
+            update_fields.append("scratch_codes")
```
Ninjaclasher, line 61:
```suggestion
            update_fields.append('scratch_codes')
```

templates/user/edit-profile.html
```diff
@@ -269,6 +350,26 @@
                                 </a>
                             </td></tr>
                         {% endif %}
+                        {% if profile.is_totp_enabled %}
+                        <tr>
+                            <td colspan="2"><span class="inline-header grayed">
+                                {{ _('Scratch Codes:') }}
+                                {% if profile.scratch_codes %}
```
Ninjaclasher, line 357: From what I can tell, the only thing that changes between these two if statement blocks is `Regenerate` vs `Generate`. Why not `if` on that rather than the entire block?

templates/user/edit-profile.html
```diff
@@ -85,7 +104,10 @@
             $('#disable-2fa-button').click(function () {
                 alert("The administrators for this site require all the staff to have Two-factor Authentication enabled, so it may not be disabled at this time.");
             });
-
+        });
+    </script>
+    <script type="text/javascript">
```
Ninjaclasher, line 109: Why do we need a separate javascript block?

judge/views/two_factor.py
```diff
@@ -51,9 +51,15 @@ class TOTPEnableView(TOTPView):
 
     def get(self, request, *args, **kwargs):
         profile = self.profile
+        update_fields = []
         if not profile.totp_key:
             profile.totp_key = pyotp.random_base32(length=32)
-            profile.save()
+            update_fields.append("totp_key")
+        if not profile.scratch_codes:
+            profile.scratch_codes = json.dumps([pyotp.random_base32(length=16)
```
Ninjaclasher, line 59: I don't particularly like this duplicating of code for scratch code generation. I think either `generate_scratch_codes` should be modified to not call `save()`, or inline `generate_scratch_codes` in the one place that it is called (since right now, there's a dedicated method to generating scratch codes for that one place, but then it's inlined here).

judge/models/profile.py
```diff
@@ -121,6 +123,10 @@ class Profile(models.Model):
                                       help_text=_('32 character base32-encoded key for TOTP'),
                                       validators=[RegexValidator('^$|^[A-Z2-7]{32}$',
                                                                  _('TOTP key must be empty or base32'))])
+    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('scratch codes'),
+                                           help_text=_('JSON of 16 character base32-encoded codes for scratch codes'),
+                                           validators=[RegexValidator('^[A-Z2-7]$',
```
Ninjaclasher, line 128: Is this regex correct? Doesn't the JSON also contain `[`, `]`, and `,`, for example? I think right now it works because calling `save` directly bypasses the validation, but you won't be able to save on the admin site.

On a related note, should this field be added to the admin site with similar permissions required as `totp_key`?

judge/forms.py
```diff
@@ -220,10 +237,16 @@ def clean(self):
 
             credential.counter = sign_count
             credential.save(update_fields=['counter'])
-        elif self.profile.is_totp_enabled and self.cleaned_data.get('totp_token'):
-            if pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'], valid_window=self.TOLERANCE):
+        elif self.profile.is_totp_enabled and totp_or_scratch_code:
+            if pyotp.TOTP(self.profile.totp_key).verify(totp_or_scratch_code, valid_window=self.TOLERANCE):
```
TheAvidDev, line 241: Aren't these now in the `totp_or_scratch_code_validators` dict?

templates/user/edit-profile.html
```diff
@@ -70,6 +70,25 @@
...
             margin-left: 5px;
         }
+
+        .hidden-word {
+            padding-top: 5px;
+            font-family: Consolas, Liberation Mono, Monaco, Courier New, monospace;
+            font-size: 12px;
+        }
+
+        .scratch-codes {
```
Xyene, line 80: I don't think this CSS applies to anything, since you have `id="scratch-codes"` and this is `.scratch-codes` rather than `#scratch-codes`? Same for `.scratch-codes-text`?

Xyene: (`id` is the correct thing to use here)

templates/user/edit-profile.html
```diff
@@ -177,6 +196,65 @@
             });
         });
     </script>
+    <script src="{{ static('libs/clipboard/clipboard.js') }}"></script>
+    <script src="{{ static('libs/clipboard/tooltip.js') }}"></script>
+    <script type="text/javascript">
+        $(function () {
+            $('#generate-scratch-codes-button').click(function (event) {
```
Xyene, line 203:
```suggestion
            $('#generate-scratch-codes-button').click(function(event) {
```

Nit: space comes after `function` iff there are no arguments to the function, i.e. the param list is `()`.

judge/models/profile.py
```diff
@@ -121,6 +123,10 @@ class Profile(models.Model):
                                       help_text=_('32 character base32-encoded key for TOTP'),
                                       validators=[RegexValidator('^$|^[A-Z2-7]{32}$',
                                                                  _('TOTP key must be empty or base32'))])
+    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('scratch codes'),
+                                           help_text=_('JSON of 16 character base32-encoded codes for scratch codes'),
+                                           validators=[RegexValidator('^["\\[\\],A-Z0-9 ]$',
```
Ninjaclasher, line 128:
```suggestion
                                           validators=[RegexValidator(r'^["\[\],A-Z0-9 ]$',
```
(Also use a similar raw string in the migration)

judge/views/two_factor.py
```diff
@@ -67,6 +69,7 @@ def post(self, request, *args, **kwargs):
     def get_context_data(self, **kwargs):
         context = super(TOTPEnableView, self).get_context_data(**kwargs)
         context['totp_key'] = self.profile.totp_key
+        context['scratch_codes'] = '\n'.join(json.loads(self.profile.scratch_codes))
```
Ninjaclasher, line 72: I think for consistency with the Ajax view, the `'\n'.join` should be moved into the template.

See: [Jinja join filter](https://jinja.palletsprojects.com/en/2.11.x/templates/#join).

templates/user/edit-profile.html
```diff
@@ -269,6 +347,23 @@
                                 </a>
                             </td></tr>
                         {% endif %}
+                        {% if profile.is_totp_enabled %}
```
Ninjaclasher, line 350: Assuming the screenshots from https://github.com/DMOJ/online-judge/pull/1334#issuecomment-626259738 are still correct, scratch codes should probably be moved under `Two-factor authentication` but before the API tokens/security keys since they are closely intertwined.

judge/models/profile.py
```diff
@@ -121,6 +123,10 @@ class Profile(models.Model):
                                       help_text=_('32 character base32-encoded key for TOTP'),
                                       validators=[RegexValidator('^$|^[A-Z2-7]{32}$',
                                                                  _('TOTP key must be empty or base32'))])
+    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('scratch codes'),
+                                           help_text=_('JSON of 16 character base32-encoded codes for scratch codes'),
+                                           validators=[RegexValidator(r'^["\[\],A-Z0-9 ]$',
```
Ninjaclasher, line 128: Doesn't this regex only match exactly one character in `"\[\],A-Z0-9 `?

judge/models/profile.py
```diff
@@ -121,6 +123,10 @@ class Profile(models.Model):
...
+                                                                      _('Scratch codes must be empty or base32'))])
```
Ninjaclasher, line 129: Should probably say something about JSON (e.g `Scratch codes must be empty or a JSON array of 16-character base32 codes` or something along those lines).

templates/registration/two_factor_auth.html
```diff
@@ -81,8 +81,8 @@
             {% endif %}
 
             {% if request.profile.is_totp_enabled %}
-                <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app:') }}</label></div>
-                <div id="totp-token-container"><span class="fullwidth">{{ form.totp_token }}</span></div>
+	            <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app <br> or one of your 16-digit scratch codes:') }}</label></div>
```
Ninjaclasher, line 84:
```suggestion
	       <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app <br> or one of your 16-character scratch codes:') }}</label></div>
```

You will need to double check on the indenting, as indenting a suggestion is hard.

templates/user/edit-profile.html
```diff
@@ -226,12 +303,14 @@
                         {% endif %}
                     </table>
                 </div>
-                <div class="pane">
-                    <div style="padding-bottom:0.3em; margin-top:0.3em" class="block-header">
-                        {{ _('Affiliated organizations') }}:
+                {% if form.organizations %}
```
Ninjaclasher, line 306: Should probably fix this failed rebase.

judge/migrations/0108_scratch_codes.py
```diff
@@ -0,0 +1,22 @@
...
+import django.db.models.deletion
+from django.db import migrations
+
+import judge.models.profile
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0107_submission_lock'),
```
Ninjaclasher, line 13: Rebase migration

judge/admin/profile.py
```diff
@@ -66,6 +66,7 @@ def get_fields(self, request, obj=None):
         if request.user.has_perm('judge.totp'):
             fields = list(self.fields)
             fields.insert(fields.index('is_totp_enabled') + 1, 'totp_key')
+            fields.append('scratch_codes')
```
Ninjaclasher, line 69: Right now, this shows the `scratch_codes` field at the bottom of the profile admin page. I think scratch codes should be right below `totp_key`, so maybe something like
```
            fields.insert(fields.index('totp_key') + 1, 'scratch_codes')
```

judge/forms.py
```diff
@@ -222,10 +241,16 @@ def clean(self):
...
+                return
+            elif totp_or_scratch_code in json.loads(self.profile.scratch_codes):
```
TheAvidDev, line 247: Eventually this should probably be added into `totp_or_scratch_code_validators`, converting its contests to a class, but that's for another PR.

dmoj/urls.py
```diff
@@ -80,6 +80,8 @@
     url(r'^2fa/webauthn/assert/$', two_factor.WebAuthnAttestView.as_view(), name='webauthn_assert'),
     url(r'^2fa/webauthn/delete/(?P<pk>\d+)$', two_factor.WebAuthnDeleteView.as_view(), name='webauthn_delete'),
```
Xyene, line 82:
```suggestion
```

templates/registration/two_factor_auth.html
```diff
@@ -81,8 +81,8 @@
             {% endif %}
 
             {% if request.profile.is_totp_enabled %}
-                <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app:') }}</label></div>
-                <div id="totp-token-container"><span class="fullwidth">{{ form.totp_token }}</span></div>
+                <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app <br> or one of your 16-character scratch codes:') }}</label></div>
```
Xyene, line 84:
```suggestion
                <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app or one of your 16-character scratch codes:') }}</label></div>
```

styling should not be included in translated messages, this should "just work" based on the CSS styling on the surrounding div/label.

templates/user/edit-profile.html
```diff
@@ -288,11 +363,41 @@
...
+                            <span class="inline-header grayed">
+                                {{ _('Scratch Codes:') }}
+                                {% if profile.scratch_codes %}
+                                    <a id="generate-scratch-codes-button" href="{{ url('generate_scratch_codes') }}"
+                                        class="button inline-button">{{ _('Regenerate') }}</a>
+                                {% else %}
+                                    <a id="generate-scratch-codes-button" href="{{ url('generate_scratch_codes') }}"
+                                        class="button inline-button">{{ _('Generate') }}</a>
+                                {% endif %}
+                                <div class="hidden-word">{{ _('Hidden') }}</div>
```
Xyene, line 375:
```suggestion
                                <div id="hidden-word">{{ _('Hidden') }}</div>
```

there is only a single one of these on the page

templates/user/edit-profile.html
```diff
@@ -70,6 +70,24 @@
         .ml-5 {
             margin-left: 5px;
         }
+
+        .hidden-word {
```
Xyene, line 74:
```suggestion
        #hidden-word {
```

judge/forms.py
```diff
@@ -20,6 +20,21 @@
 from judge.widgets import HeavyPreviewPageDownWidget, MathJaxPagedownWidget, PagedownWidget, Select2MultipleWidget, \
     Select2Widget
 
+totp_or_scratch_code_validators = {
```
Xyene, line 23:
```suggestion
two_factor_validators_by_length = {
```

just to make clear what the key in this dict is

judge/forms.py
```diff
@@ -166,20 +181,23 @@ def widget_attrs(self, widget):
...
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
-            raise ValidationError(_('Invalid two-factor authentication token.'))
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
+        if totp_or_scratch_code_len not in totp_or_scratch_code_validators:
+            raise ValidationError(_('Invalid code length.'))
+        else:
+            validator = totp_or_scratch_code_validators[totp_or_scratch_code_len]
```
Xyene, line 197:
```suggestion
        try:
            validator = totp_or_scratch_code_validators[len(totp_or_scratch_code)]
        except KeyError:
            raise ValidationError(_('Invalid code length.'))
```

more Pythonic to catch the exception

judge/forms.py
TheAvidDev: This is also just over complicating things.

Carson-Tang:
Suggested changes resolved.

judge/forms.py
```diff
@@ -166,20 +181,26 @@ def widget_attrs(self, widget):
...
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
-            raise ValidationError(_('Invalid two-factor authentication token.'))
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
+        if totp_or_scratch_code_len not in two_factor_validators_by_length:
```
Ninjaclasher, line 194: Since we are catching the `KeyError` later on, I don't think this `if` is necessary.

judge/models/profile.py
```diff
@@ -124,6 +126,12 @@ class Profile(models.Model):
                                       help_text=_('32 character base32-encoded key for TOTP'),
                                       validators=[RegexValidator('^$|^[A-Z2-7]{32}$',
                                                                  _('TOTP key must be empty or base32'))])
+    scratch_codes = EncryptedNullCharField(max_length=255, null=True, blank=True, verbose_name=_('scratch codes'),
+                                           help_text=_('JSON array of 16 character base32-encoded codes \
+                                                        for scratch codes'),
+                                           validators=[RegexValidator(r'^["\[\],A-Z0-9 ]*$',
```
Ninjaclasher, line 132:
```suggestion
                                           validators=[RegexValidator(r'^(\[\])?$|^\[("[A-Z2-9]{16}", *)*"[A-Z2-9]{16}"\]$',
```
This regex was bothering me, so I wrote a new one that should perfectly match a JSON array.

If we don't care about a perfect JSON match, this regex would work, but it allows trailing commas (which would `json.loads` to error) :
```
^$|^\[("[A-Z2-9]{16}"(, *)?)*\]$
```

templates/common-content.html
```diff
@@ -40,12 +40,12 @@
 
                     var curClipboard = new Clipboard(copyButton.get(0));
 
-                    curClipboard.on('success', function (e) {
+                    curClipboard.on('success', function(e) {
```
quantum5, line 43: Pointless unrelated change.

templates/common-content.html
```diff
@@ -40,12 +40,12 @@
...
                         e.clearSelection();
                         showTooltip(e.trigger, 'Copied!');
                     });
 
-                    curClipboard.on('error', function (e) {
+                    curClipboard.on('error', function(e) {
```
quantum5, line 48: Pointless unrelated change.

judge/forms.py
```diff
@@ -166,20 +181,23 @@ def widget_attrs(self, widget):
...
         self.profile = kwargs.pop('profile')
         super().__init__(*args, **kwargs)
 
     def clean(self):
-        if (not self.cleaned_data.get('totp_token') or
-                not pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'],
-                                                             valid_window=self.TOLERANCE)):
-            raise ValidationError(_('Invalid two-factor authentication token.'))
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        totp_or_scratch_code_len = len(totp_or_scratch_code)
```
Xyene, line 193: this variable is used in a single place and is just as long as it's RHS, so just inline it

judge/forms.py
```diff
@@ -222,10 +240,16 @@ def clean(self):
...
-        elif self.profile.is_totp_enabled and self.cleaned_data.get('totp_token'):
-            if pyotp.TOTP(self.profile.totp_key).verify(self.cleaned_data['totp_token'], valid_window=self.TOLERANCE):
+        elif self.profile.is_totp_enabled and totp_or_scratch_code:
+            if pyotp.TOTP(self.profile.totp_key).verify(totp_or_scratch_code, valid_window=self.TOLERANCE):
+                return
+            elif totp_or_scratch_code in json.loads(self.profile.scratch_codes):
+                scratch_codes = json.loads(self.profile.scratch_codes)
+                scratch_codes.remove(totp_or_scratch_code)
+                self.profile.scratch_codes = json.dumps(scratch_codes)
+                self.profile.save()
```
quantum5, line 250: Maybe `update_fields=['scratch_codes']`

judge/forms.py
```diff
@@ -166,20 +181,22 @@ def widget_attrs(self, widget):
...
-                                                             valid_window=self.TOLERANCE)):
-            raise ValidationError(_('Invalid two-factor authentication token.'))
+        totp_or_scratch_code = self.cleaned_data.get('totp_or_scratch_code')
+        try:
+            validator = two_factor_validators_by_length[len(totp_or_scratch_code)]
+        except KeyError:
+            raise ValidationError(_('Invalid code length.'))
+        validator['regex_validator'](totp_or_scratch_code)
+        if validator['verify'](totp_or_scratch_code, self.profile):
+            raise validator['err']
```
quantum5, line 199: Now that I think about it, using the same exception object every time is sketchy.

PR #1338, opened 2020-04-25
Rewrite problem submit as a class-based view

Ninjaclasher:


judge/views/problem.py
```diff
@@ -521,128 +520,150 @@ def get(self, request, *args, **kwargs):
...
-                    model = form.save()
+    @cached_property
+    def contest_problem(self):
+        if self.request.profile.current_contest is None:
+            return None
+        return get_contest_problem(self.object, self.request.profile)
+
+    @cached_property
+    def remaining_submission_count(self):
+        max_subs = getattr(self.contest_problem, 'max_submissions', None)
```
Xyene, line 535: Apart from this stuff, this PR looks good, thanks.

Ninjaclasher: This has been changed in the latest commit.

PR #1340, opened 2020-05-01
Clear MiscConfig cache properly when saving/deleting the object; #1339

### LINKED ISSUE #1339: Misc. config cache not cleared when the key is changed.
Two possible fixes:
 1. The signal [here](https://github.com/DMOJ/online-judge/blob/master/judge/signals.py#L123) is changed to a `pre_save` so the old instance can be retrieved. The cache should really be cleared based on the old key rather than the new key.
 2. Change the `key` to only be read-only in the admin page. However, this doesn't prevent the key from being changed from e.g. the Python shell, and the cache still won't be cleared properly in this case.

Ninjaclasher:
Fixes #1339.

### DIFF
--- a/judge/signals.py
+++ b/judge/signals.py
@@ -5,7 +5,7 @@ from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.cache import cache
 from django.core.cache.utils import make_template_fragment_key
-from django.db.models.signals import post_delete, post_save
+from django.db.models.signals import post_delete, post_save, pre_save
 from django.dispatch import receiver
 
 from .caching import finished_submission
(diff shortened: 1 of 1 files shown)


judge/signals.py
```diff
@@ -120,13 +120,29 @@ def organization_update(sender, instance, **kwargs):
...
+@receiver(pre_save, sender=MiscConfig)
+def misc_config_update(sender, instance, **kwargs):
+    misc_config_cache_delete(instance.key)
+    try:
+        old_key = MiscConfig.objects.get(id=instance.id).key
+    except MiscConfig.DoesNotExist:
+        pass
+    else:
+        if old_key != instance.key:
+            misc_config_cache_delete(old_key)
```
quantum5, line 138: Isn't this a race? Since this is pre-save, a query could happen just before it saves and cache the old value again.

Ninjaclasher: What would be a solution to fix this?

One solution might be to grab the old key in `pre_save` and clear the cache in `post_save`.

Ninjaclasher: The race should be fixed now by grabbing the old key in `pre_save` and clearing the cache in `post_save`.

PR #1344, opened 2020-05-02
Add zig to languages

kipply:
I'm only like 80% sure I incremented pk correctly? 

Depends on https://github.com/DMOJ/judge-server/pull/658

### DIFF
--- a/judge/fixtures/language_all.json
+++ b/judge/fixtures/language_all.json
@@ -942,5 +942,21 @@
         "description": "",
         "extension": ""
     }
+},
+{
+    "model": "judge.language",
+    "pk": 75,
+    "fields": {
+        "key": "ZIG",
+        "name": "Zig",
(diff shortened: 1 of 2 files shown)


judge/fixtures/language_all.json
```diff
@@ -942,5 +942,21 @@
...
+    "pk": 75,
+    "fields": {
+        "key": "ZIG",
+        "name": "Zig",
+        "short_name": "",
+        "common_name": "Zig",
+        "ace": "zig",
+        "pygments": "zig",
+        "template": "",
+        "info": "zig 0.6.0",
```
Xyene, line 957: Don't override the `info` field unless the judge misreports the version itself.

kipply: I'm assuming just version (`0.6.0`)  and not a name is valid~

PR #1352, opened 2020-05-04
Add WebAuthn support for 2FA

### LINKED ISSUE #1201: Add U2F security key for 2FA
We currently have TOTP for 2FA. This is secure, but not as secure as a hardware security key.

For fun, we should support U2F security keys.

quantum5:
Fixes #1201.


judge/forms.py
```diff
@@ -133,16 +136,70 @@ class TOTPForm(Form):
...
+
+            try:
+                sign_count = assertion.verify()
+            except Exception as e:
+                raise ValidationError(str(e))
+
+            credential.counter = sign_count
+            credential.save(update_fields=['counter'])
+
+            return
```
Xyene, line 194: else...?

judge/migrations/0104_webauthn.py
```diff
@@ -0,0 +1,30 @@
...
+
+    dependencies = [
+        ('judge', '0103_contest_participation_tiebreak_field'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='profile',
+            name='is_webauthn_enabled',
+            field=models.BooleanField(default=False, help_text='check to enable WebAuthn-based two factor authentication', verbose_name='WebAuthn 2FA enabled'),
```
Xyene, line 17:
```suggestion
            field=models.BooleanField(default=False, help_text='check to enable WebAuthn-based two-factor authentication', verbose_name='WebAuthn 2FA enabled'),
```

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+
+    def check_skip(self):
+        raise NotImplementedError()
+
+    def next_page(self):
+        return HttpResponseRedirect(reverse('user_edit_profile'))
+
+
+class TOTPEnableView(TOTPView):
+    title = _('Enable Two Factor Authentication')
```
Xyene, line 49:
```suggestion
    title = _('Enable Two-Factor Authentication')
```

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+        qr.make(fit=True)
+
+        image = qr.make_image(fill_color='black', back_color='white')
+        buf = BytesIO()
+        image.save(buf, format='PNG')
+        return 'data:image/png;base64,' + base64.b64encode(buf.getvalue()).decode('ascii')
+
+
+class TOTPDisableView(TOTPView):
+    title = _('Disable Two Factor Authentication')
```
Xyene, line 96:
```suggestion
    title = _('Disable Two-Factor Authentication')
```

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+            return HttpResponseBadRequest(_('Invalid WebAuthn response'))
+
+        if 'name' not in request.POST or len(request.POST['name']) > 100:
+            return HttpResponseBadRequest(_('Invalid name'))
+
+        credential = json.loads(request.POST['credential'])
+
+        response = webauthn.WebAuthnRegistrationResponse(
+            rp_id=settings.WEBAUTHN_RP_ID,
+            origin='https://' + request.get_host(),
```
Xyene, line 153: uhhh, don't we have a setting for this?

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+        if count <= 1:
+            request.profile.is_webauthn_enabled = False
+            request.profile.save(update_fields=['is_webauthn_enabled'])
+
+        return HttpResponse()
+
+
+class TwoFactorLoginView(SuccessURLAllowedHostsMixin, TOTPView):
+    form_class = TwoFactorLoginForm
+    title = _('Perform Two Factor Authentication')
```
Xyene, line 212:
```suggestion
    title = _('Perform Two-Factor Authentication')
```

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+    template_name = 'registration/two_factor_auth.html'
+
+    def get_form_kwargs(self):
+        result = super().get_form_kwargs()
+        result['webauthn_challenge'] = self.request.session.get('webauthn_assert')
+        result['webauthn_origin'] = 'https://' + self.request.get_host()
```
Xyene, line 218: Same here

templates/registration/two_factor_auth.html
```diff
@@ -0,0 +1,98 @@
...
+                <div><label class="inline-header grayed">{{ _('Enter the 6-digit code generated by your app:') }}</label></div>
+                <div id="totp-token-container"><span class="fullwidth">{{ form.totp_token }}</span></div>
+                <hr>
+            {% endif %}
+            {% if request.profile.is_webauthn_enabled %}
+                {{ form.webauthn_response }}
+                <button style="display: inline" id="use-webauthn">{{ _('Use security key') }}</button>
+            {% endif %}
+            {% if request.profile.is_totp_enabled %}
+                <button style="float:right;" type="submit">{{ _('Login!') }}</button>
```
Xyene, line 93:
```suggestion
                <button style="float:right;" type="submit">{{ _('Log in!') }}</button>
```

judge/forms.py
quantum5: Why do you need an else? It just falls through to the TOTP part.

judge/migrations/0104_webauthn.py
quantum5: Cannot accept suggestion because it's not also made to the model itself. Furthermore, TOTP needs to be fixed too.

judge/views/two_factor.py
quantum5: `Two-Factor` has incorrect capitalization. Also this is out of scope because it was actually from `judge/views/totp.py`. Somehow it was detected as a new file instead of rename.

judge/views/two_factor.py
quantum5: `Two-Factor` has incorrect capitalization. Also this is out of scope because it was actually from `judge/views/totp.py`. Somehow it was detected as a new file instead of rename.

judge/views/two_factor.py
quantum5: What setting? This is supposed to be the scheme + hostname for the request. The same value is included in the `clientDataJSON` blob and signed. We check them for equality to make sure the response is not intended for a different domain.

judge/views/two_factor.py
quantum5: `Two-Factor` has incorrect capitalization. Also this is out of scope because it was actually from `judge/views/totp.py`. Somehow it was detected as a new file instead of rename.

judge/views/two_factor.py
quantum5: What setting? This is supposed to be the scheme + hostname for the request. The same value is included in the `clientDataJSON` blob and signed. We check them for equality to make sure the response is not intended for a different domain.

templates/registration/two_factor_auth.html
quantum5: This is also old code from `templates/registration/totp_auth.html`.

judge/views/two_factor.py
Xyene: You're assuming everyone running has HTTPS, but as far as I could find there's nothing restricting webauthn usage to HTTPS-only.

templates/registration/two_factor_auth.html
Xyene: You're still touching it so it's already going into git blame, so this is a good time for incremental improvements.

judge/views/two_factor.py
Xyene: (`DMOJ_SCHEMA`, IIRC)

judge/migrations/0104_webauthn.py
```diff
@@ -0,0 +1,30 @@
+# Generated by Django 2.2.12 on 2020-05-03 23:58
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0103_contest_participation_tiebreak_field'),
```
Xyene, line 10: New migration went in, so this needs rebasing.

judge/views/two_factor.py
```diff
@@ -0,0 +1,236 @@
...
+    def get_context_data(self, **kwargs):
+        context = super(TOTPEnableView, self).get_context_data(**kwargs)
+        context['totp_key'] = self.profile.totp_key
+        context['qr_code'] = self.render_qr_code(self.request.user.username, self.profile.totp_key)
+        return context
+
+    def form_valid(self, form):
+        self.profile.is_totp_enabled = True
+        self.profile.save()
+        # Make sure users don't get prompted to enter code right after enabling:
```
Xyene, line 76:
```suggestion
        # Make sure users don't get prompted to enter code right after enabling
```

PR #1359, opened 2020-05-05
Update Django documentation URLs to v2.2 in settings.py

TheAvidDev:
It's been a while since v1.11 and we no longer support it, so let's link to the version we do support (v2.2).


Xyene:
Thanks!

PR #1363, opened 2020-05-06
Move comment form to the location of the comment when replying; #1333

### LINKED ISSUE #1333: When replying to a comment, add some visual indicator so users know they are replying
Currently, the user has no way of knowing whether they are replying to a comment or making a new comment without checking the HTML.

One possible implementation would be changing `New comment` to `Replying to comment <id>`.

Ninjaclasher:
Closes #1333.

image


TheAvidDev:
What does this look like if you have quite a few nested comments?

Ninjaclasher:
image

PR #1364, opened 2020-05-06
Make Edit Profile page more mobile friendly

Ninjaclasher:
image

image


Ninjaclasher:
`Preferred language` has been changed to `Language` so it fits on one line.

PR #1366, opened 2020-05-06
 Use class instead of nth-child for hiding problem list columns; #1055

### LINKED ISSUE #1055: DMOJ problem list page on mobile UI issues
Reproduction link: https://dmoj.ca/problems/?show_types=1&type=3&page=8

Courtesy of @WallE256 - https://imgur.com/a/bezoX5q

Ninjaclasher:
Fixes #1055.

Side question: why are columns being hidden in the first place? Why not make the table scroll horizontally instead?

### DIFF
--- a/resources/problem.scss
+++ b/resources/problem.scss
@@ -232,19 +232,19 @@ ul.problem-list {
 }
 
 @media (max-width: 550px) {
-    #problem-table tr :nth-child(6) {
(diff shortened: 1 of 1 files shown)


Xyene:
Thanks for all the UI fixes! Re: columns, very likely because making stuff mobile-friendly was a quick hack job, as you've probably noticed :dash:

PR #1369, opened 2020-05-06
Fix UI issues related to #1368

Ninjaclasher:
1. Tables do not take the full width of the screen in special cases.
2. Runtime versions table header is broken if the table is set to horizontally scrollable.


Ninjaclasher:
Not ready for merge - still some UI issues that need to be fixed.

Ninjaclasher:
All UI issues should be fixed now.

PR #1376, opened 2020-05-08
Add highlighted ranges to user rating chart

Ninjaclasher:
Screenshots:
image
image

Special thanks to [this StackOverflow post](https://stackoverflow.com/questions/36685745/acceptable-range-highlighting-of-background-in-chart-js-2-0) for parts of the code and @Carson-Tang for some of the colours.

### DIFF
--- a/templates/user/user-about.html
+++ b/templates/user/user-about.html
@@ -202,7 +202,7 @@
                                 }).show();
                             },
                         },
-                        yHighlight : [
+                        yHighlight: [
                             {
                                 begin: 0,
                                 end: 1000,
--- a/templates/user/user-about.html
+++ b/templates/user/user-about.html
@@ -85,15 +85,56 @@
                     return rating_history[index];
                 };
 
+                var originalDraw = Chart.controllers.scatter.prototype.draw;
+                Chart.helpers.extend(Chart.controllers.scatter.prototype, {
+                    draw : function() {
+                        var chart = this.chart;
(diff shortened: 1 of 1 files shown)


templates/user/user-about.html
```diff
@@ -85,15 +85,56 @@ <h4>Rating History</h4>
...
+                        var chart = this.chart;
+                        var yHighlight = chart.config.options.yHighlight;
+
+                        if (yHighlight !== undefined) {
+                            var ctx = chart.chart.ctx;
+                            var xaxis = chart.scales['x-axis-1'];
+                            var yaxis = chart.scales['y-axis-1'];
+
+                            ctx.save();
+                            for (var range of yHighlight) {
```
quantum5, line 100: Can we use the ES5 `forEach` instead?

templates/user/user-about.html
```diff
@@ -85,15 +85,56 @@ <h4>Rating History</h4>
                     return rating_history[index];
                 };
 
+                var originalDraw = Chart.controllers.scatter.prototype.draw;
+                Chart.helpers.extend(Chart.controllers.scatter.prototype, {
+                    draw : function() {
+                        var chart = this.chart;
+                        var yHighlight = chart.config.options.yHighlight;
+
+                        if (yHighlight !== undefined) {
```
quantum5, line 94:
```suggestion
                        if (yHighlight != null) {
```

Ninjaclasher:
Added suggested changes.

PR #1378, opened 2020-05-09
Refactor contest scoreboard access checks into two methods

Ninjaclasher:
Currently, there is one method to determine whether a user can access the contest scoreboard, `can_see_scoreboard`. However, it doesn't distinguish between whether the user can see _their_ performance or the _entire_ rankings, and it is left to the caller to distinguish.

This PR refactors this logic into two methods, `can_see_scoreboard` - whether the user can click on "Rankings", and `can_see_full_scoreboard` - whether the user can see the entire rankings. 

Most of the code here has been in production on MCPT for a while now (a few months), so there should not be any accidental scoreboard leaks.

### DIFF
--- a/judge/models/contest.py
+++ b/judge/models/contest.py
@@ -165,17 +165,30 @@ class Contest(models.Model):
         return False
 
     def can_see_scoreboard(self, user):
+        if self.can_see_full_scoreboard(user):
+            return True
+        if not self.can_join:
+            return False
+        if not self.show_scoreboard and not self.is_in_contest(user):
+            return False
+        return True
+
+    def can_see_full_scoreboard(self, user):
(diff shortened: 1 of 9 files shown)


judge/models/contest.py
```diff
@@ -165,17 +165,30 @@ def is_in_contest(self, user):
         return False
 
     def can_see_scoreboard(self, user):
```
quantum5, line 167: Maybe `can_see_own_scoreboard`? Having `can_see_scoreboard` and `can_see_full_scoreboard` feels a bit weird.

Ninjaclasher: This has been changed in the latest commit.

PR #1384, opened 2020-05-11
Implement user reputation

### LINKED ISSUE #839: Vote-based user reputation
Users appear to care significantly about rating. Perhaps adding a point value for "reputation" based on comment votes would help encourage more thoughtful discussion, and less spam.

Ninjaclasher:
Closes #839.

### DIFF
--- a/dmoj/settings.py
+++ b/dmoj/settings.py
@@ -47,6 +47,19 @@ DMOJ_PP_STEP = 0.95
 DMOJ_PP_ENTRIES = 100
 DMOJ_PP_BONUS_FUNCTION = lambda n: 300 * (1 - 0.997 ** n)  # noqa: E731
 
+DMOJ_REPUTATION_STEP = 0.98
+DMOJ_REPUTATION_ENTRIES = 20
+
+
+def DMOJ_REPUTATION_FUNCTION(n):
+    import math
+    if n < 0:
+        return -DMOJ_REPUTATION_FUNCTION(-n)
+    elif n <= 1:
+        return 0
+    return (math.tanh(n / 3 - 3) / 1.3 + 1) * 2 * math.log(n)
+
+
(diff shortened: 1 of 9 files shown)


Ninjaclasher:
Reputation is calculated as follows:
 1. Take the latest `DMOJ_REPUTATION_ENTRIES` (= 20) comment scores by the user.
 2. The latest comment is weighted 1, the second weighted 0.98 ... nth comment is weighted 0.98 ^ n (similar to PP calculation)
 3. Each comment's weight is multiplied by `f(comment's score)`, where `f` is `DMOJ_REPUTATION_FUNCTION` (denote as the comment's "value")
 4. The sum of the comment values is the user's reputation.

By default, `DMOJ_REPUTATION_FUNCTION` is the product of a sigmoid-like function (`tanh`) and the natural logarithm of the comment's score (slightly rescaled).

This is done to combat users who vote for their own comments (<= 4 votes increases reputation very little, >= 4 and <= 16 votes increases reputation a lot, >= 16 votes increases reputation at a logarithmic rate).

Ninjaclasher:
Desmos graph of the default function:
image

kiritofeng:
> 
> 
> Code looks good. However, I don't know if the correct behaviour is boosting the user's reputation if we hide disruptive posts.

What if we apply a penalty for hidden posts?

Ninjaclasher:
I think maybe it's better to just include hidden comment scores into the reputation calculation as well. 

The issue with applying a penalty to hidden posts is if a user posts a bad comment and someone replies to that comment with a good comment (e.g a moderator). Now if the first comment gets hidden, both users would be punished since hiding a comment hides all descendant comments as well.

Ninjaclasher:
I've updated this PR to include hidden comment scores in the reputation calculation.

PR #1391, opened 2020-05-13
Implement submission locks

### LINKED ISSUE #595: Submission Lock
A `BooleanField` on each submission, that would prevent it from being rejudged. Should probably be applied to contest submissions post-contest (automatically?)

Would probably require two new perms as well:
- `RejudgeLockedSubmission` name is self-explanatory
- `UnlockSubmission` allows user to unlock a sub

WallE256:
Closes #595


judge/models/contest.py
```diff
@@ -126,6 +126,8 @@ class Contest(models.Model):
                                             help_text='A custom Lua function to generate problem labels. Requires a '
                                                       'single function with an integer parameter, the zero-indexed '
                                                       'contest problem index, and returns a string, the label.')
+    is_locked = models.BooleanField(verbose_name=_('contest lock'), default=False,
+                                    help_text='If locked, the submissions from this contest will be locked')
```
Carson-Tang, line 130: Period

judge/admin/contest.py
```diff
@@ -211,13 +216,48 @@ def make_hidden(self, request, queryset):
...
             url(r'^(\d+)/rate/$', self.rate_view, name='judge_contest_rate'),
             url(r'^(\d+)/judge/(\d+)/$', self.rejudge_view, name='judge_contest_rejudge'),
+            url(r'^(\d+)/lock/$', self.lock_view, name='judge_contest_lock'),
         ] + super(ContestAdmin, self).get_urls()
 
+    def lock_view(self, request, id):
+        if not request.user.has_perm('judge.contest_lock'):
+            raise PermissionDenied()
+        contest = get_object_or_404(Contest, id=id)
+        self.set_is_locked(contest, not contest.is_locked)
```
quantum5, line 258: Not sure how I feel about the lack of idempotence with this toggle, and the fact it's a GET.

judge/models/contest.py
```diff
@@ -126,6 +126,8 @@ class Contest(models.Model):
                                             help_text='A custom Lua function to generate problem labels. Requires a '
                                                       'single function with an integer parameter, the zero-indexed '
                                                       'contest problem index, and returns a string, the label.')
+    is_locked = models.BooleanField(verbose_name=_('contest lock'), default=False,
+                                    help_text='If locked, the submissions from this contest will be locked.')
```
TheAvidDev, line 130: Either this should explain what:
> submissions from this contest will be locked

means, or `help_text` should be included in the `is_locked` field of a Submission.

judge/admin/contest.py
```diff
@@ -211,13 +216,48 @@ def make_hidden(self, request, queryset):
...
+        self.message_user(request, ungettext('%d contest successfully unlocked.',
+                                             '%d contests successfully unlocked.',
+                                             count) % count)
+    set_unlocked.short_description = _('Unlock contest submissions')
+
+    def set_is_locked(self, contest, is_locked):
+        with transaction.atomic():
+            contest.is_locked = is_locked
+            contest.save()
+            for row in ContestSubmission.objects.filter(participation__contest=contest):
```
quantum5, line 241: Why do we need this loop? `ContestSubmission.objects.filter(participation__contest=contest).update(submission__is_locked=True)` works just as well.

judge/migrations/0106_submission_lock.py
```diff
@@ -0,0 +1,48 @@
...
+
+def updatecontestsubmissions(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    for row in Contest.objects.all():
+        if row.end_time < timezone.now():
+            row.is_locked = True
+            row.save()
+
+    ContestSubmissions = apps.get_model('judge', 'ContestSubmission')
+    for row in ContestSubmissions.objects.all():
```
quantum5, line 15: This can be done in one query.

judge/admin/contest.py
```diff
@@ -211,13 +216,45 @@ def make_hidden(self, request, queryset):
...
+        self.message_user(request, ungettext('%d contest successfully unlocked.',
+                                             '%d contests successfully unlocked.',
+                                             count) % count)
+    set_unlocked.short_description = _('Unlock contest submissions')
+
+    def set_is_locked(self, contest, is_locked):
+        with transaction.atomic():
+            contest.is_locked = is_locked
+            contest.save()
+            ContestSubmission.objects.filter(participation__contest=contest).update(submission__is_locked=is_locked)
```
quantum5, line 241: Does this work? Didn't you use something else in the migration?

templates/admin/judge/contest/change_form.html
```diff
@@ -20,4 +20,11 @@
             <span class="text">{% trans "Rate" %}</span>
         </a>
     {% endif %}
+    {% if original and perms.judge.contest_lock %}
+        <a style="display: none" title="{% trans "Toggle Lock" %}" href="{% url 'admin:judge_contest_lock' original.pk %}"
+           class="button rerate-link">
```
Ninjaclasher, line 25: `rerate-link`?

judge/migrations/0106_submission_lock.py
```diff
@@ -0,0 +1,44 @@
...
+
+def updatecontestsubmissions(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    for row in Contest.objects.all():
+        if row.end_time < timezone.now():
+            row.is_locked = True
+            row.save()
+
+    Submission = apps.get_model('judge', 'Submission')
+    Submission.objects.filter(contest__participation__contest__is_locked=True).update(is_locked=True)
```
Ninjaclasher, line 15:
```suggestion
    Submission.objects.filter(contest_object__is_locked=True).update(is_locked=True)
```

judge/migrations/0106_submission_lock.py
```diff
@@ -0,0 +1,44 @@
+# Generated by Django 2.2.12 on 2020-05-13 20:58
+
+from django.db import migrations, models
+from django.utils import timezone
+
+
+def updatecontestsubmissions(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    for row in Contest.objects.all():
```
Ninjaclasher, line 9: Similar thing as with the contest admin action, is there a reason a `Contest.objects.filter(...).update(...)` can't be used here?

judge/admin/contest.py
Ninjaclasher, line 241: Would it be more efficient to use a `Submission.objects.filter(contest_object=contest).update(is_locked=True)` here instead?

judge/migrations/0106_submission_lock.py
WallE256: Only base field are available in migrations, so the `ended` property has to be manually computed.

Ninjaclasher: `ended` doesn't seem to be used anywhere here. Only `end_time` and `is_locked` are used, which are both base fields (unless I'm misunderstanding something).

judge/admin/contest.py
quantum5: This should at least be a post request, and might make more sense have separate lock and unlock operations.

judge/models/contest.py
```diff
@@ -342,6 +344,7 @@ class Meta:
             ('create_private_contest', _('Create private contests')),
             ('change_contest_visibility', _('Change contest visibility')),
             ('contest_problem_label', _('Edit contest problem label script')),
+            ('lock_contest', 'Change lock status of contest'),
```
Ninjaclasher, line 347:
```suggestion
            ('lock_contest', _('Change lock status of contest')),
```

judge/models/contest.py
WallE256: Maybe something like `Prevent submissions from this contest from being rejudged`?

TheAvidDev: Sounds good.

judge/models/submission.py
```diff
@@ -114,6 +115,8 @@ def long_status(self):
         return Submission.USER_DISPLAY_CODES.get(self.short_status, '')
 
     def judge(self, *args, **kwargs):
+        if self.is_locked:
+            return
         judge_submission(self, *args, **kwargs)
```
Xyene, line 120:
```suggestion
        if not self.is_locked:
            judge_submission(self, *args, **kwargs)
```

templates/admin/judge/contest/change_form.html
```diff
@@ -20,4 +21,18 @@
             <span class="text">{% trans "Rate" %}</span>
         </a>
     {% endif %}
+    {% if original and not original.is_locked and perms.judge.contest_lock %}
+        <a style="display: none" title="{% trans "Lock Contest" %}" href="{% url 'admin:judge_contest_lock' original.pk %}"
+           class="button lock-link">
+            <i class="fa fa-lg fa-lock"></i>
+            <span class="text">{% trans "Lock Contest" %}</span>
```
Xyene, line 28:
```suggestion
            <span class="text">{% trans "Lock submissions" %}</span>
```

templates/admin/judge/contest/change_form.html
```diff
@@ -20,4 +21,18 @@
...
+        </a>
+    {% endif %}
+    {% if original and original.is_locked and perms.judge.contest_lock %}
+        <a style="display: none" title="{% trans "Unlock Contest" %}" href="{% url 'admin:judge_contest_unlock' original.pk %}"
+           class="button lock-link">
+            <i class="fa fa-lg fa-unlock"></i>
+            <span class="text">{% trans "Unlock Contest" %}</span>
```
Xyene, line 35:
```suggestion
            <span class="text">{% trans "Unlock submissions" %}</span>
```

judge/admin/contest.py
```diff
@@ -211,13 +216,53 @@ def make_hidden(self, request, queryset):
...
+        self.message_user(request, ungettext('%d contest successfully unlocked.',
+                                             '%d contests successfully unlocked.',
+                                             count) % count)
+    set_unlocked.short_description = _('Unlock contest submissions')
+
+    def set_is_locked(self, contest, is_locked):
+        with transaction.atomic():
+            contest.is_locked = is_locked
+            contest.save()
+            Submission.objects.filter(contest_object=contest).update(is_locked=is_locked)
```
kiritofeng, line 241: Should we lock all submissions, or only non-virtual submissions?

PR #1393, opened 2020-05-20
Organization improvements

Carson-Tang:


templates/organization/users.html
```diff
@@ -23,7 +31,9 @@
     <script type="text/javascript">
         $(function () {
             $('form.kick-form').find('a.button').click(function () {
-                $(this).parent().submit();
+                if (confirm('Are you sure you want to kick?')){
```
quantum5, line 34:
```suggestion
                if (confirm('Are you sure you want to kick?')) {
```

templates/organization/users.html
```diff
@@ -23,7 +31,9 @@
     <script type="text/javascript">
         $(function () {
             $('form.kick-form').find('a.button').click(function () {
-                $(this).parent().submit();
+                if (confirm('Are you sure you want to kick?')) {
```
Xyene, line 34: Missing i18n, also probably want to say something like `to kick this user?`

Issue #656, opened 2017-12-30
edit_profile page can crash


monouno:
I found that the browser (Chrome\Firefox\Safari) will crash while accessing the edit profile page when the QuerySet of "organizations" is empty.
https://github.com/DMOJ/site/blob/7b5c11371d8224901c507fa0cc1ef8b2475de1dd/judge/forms.py#L59

quantum5:
This is a problem in the JavaScript code in a third party library. The fix is to add an organization.

@Xyene an example organization should be created by our default fixtures to avoid this situation.

Xyene:
An example organization is created by our fixtures; I suppose an alternative for users who don't want to have an organization (or have no need of one) is to simply not display the organization field if there are no organizations. Then, the JS would not be loaded to cause problems.

PR #1399, opened 2020-05-25
Remove widget if no open organizations available

kevinjycui:
In regards to [Issue 656](https://github.com/DMOJ/online-judge/issues/656)

Removes the "Affiliated organizations" widget when no open organizations are available to the user. Prevents the browser from crashing due to a problem in the JavaScript code in a third party library.

[Comment regarding fix](https://github.com/DMOJ/online-judge/issues/656#issuecomment-354601730)

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -64,16 +64,12 @@ class ProfileForm(ModelForm):
     def __init__(self, *args, **kwargs):
         user = kwargs.pop('user', None)
         super(ProfileForm, self).__init__(*args, **kwargs)
-        if not Organization.objects.exists():
-            del self.fields['organizations']
-        elif not user.has_perm('judge.edit_all_organization'):
-            filtered_organizations = Organization.objects.filter(
+        if not user.has_perm('judge.edit_all_organization'):
+            self.fields['organizations'].queryset = Organization.objects.filter(
                 Q(is_open=True) | Q(id__in=user.profile.organizations.all()),
             )
-            if not filtered_organizations.exists():
-                del self.fields['organizations']
-            else:
-                self.fields['organizations'].queryset = filtered_organizations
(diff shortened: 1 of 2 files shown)


judge/forms.py
```diff
@@ -64,10 +64,16 @@ def clean(self):
...
         super(ProfileForm, self).__init__(*args, **kwargs)
-        if not user.has_perm('judge.edit_all_organization'):
-            self.fields['organizations'].queryset = Organization.objects.filter(
+        if not Organization.objects.exists():
+            del self.fields['organizations']
+        elif not user.has_perm('judge.edit_all_organization'):
+            filtered_organizations = Organization.objects.filter(
                 Q(is_open=True) | Q(id__in=user.profile.organizations.all()),
             )
+            if not filtered_organizations.exists():
```
Ninjaclasher, line 73: This hits the database a lot of unnecessary times. Perhaps we can `del self.fields['organizations']` by directly checking `self.fields['organizations'].queryset` (Assuming we _should_ be `del` in the first place...)

kevinjycui: Revised: changed `exists()` check to be directly on `self.fields['organizations'].queryset` and replaced `del self.fields['organizations']` with `self.fields['organizations'].pop('organizations')`

judge/forms.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, *args, **kwargs):
             self.fields['organizations'].queryset = Organization.objects.filter(
                 Q(is_open=True) | Q(id__in=user.profile.organizations.all()),
             )
+        if not self.fields['organizations'].queryset.exists():
```
Ninjaclasher, line 71:
```suggestion
        if not self.fields['organizations'].queryset:
```
We will need the organizations later anyways, so might as well fetch them now.

PR #1401, opened 2020-05-28
Add IOI format with new rules from 2016

adambenali:
This adds an IOI format with the scoring rules from 2016, where, basically, the submissions are not treated independently, but the score is rather an aggregation of all of them.

The rules can be found [here](http://ioi2017.org/contest/rules#feedback).

Not sure if the format name is OK. Also, the query is not very beautiful, it can probably be optimized.


judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+@register_contest_format('ioi16')
+class IOI16ContestFormat(IOIContestFormat):
+    name = gettext_lazy('IOI16')
+    config_defaults = {'cumtime': False}
+    '''
+        cumtime: Specify True if time penalties are to be computed. Defaults to False.
+    '''
+
+    def __init__(self, contest, config):
+        super(IOI16ContestFormat, self).__init__(contest, config)
```
Ninjaclasher, line 18: Since you are only calling the superclass, is this `__init__` necessary?

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                                      GROUP BY cp.id, tc.batch, sub.id
+                                  ) r
+                             GROUP BY prob, batch
+                         ) p
+                                       ON p.prob = q.prob AND p.batch = q.batch
+                         WHERE p.max_batch_points = q.batch_points
+                           AND p.prob = q.prob
+                         GROUP BY q.prob, q.batch
+                     ) best
+                ORDER BY best.date;
```
Ninjaclasher, line 72: Do we need to order by `best.date`? From what I can see, we can find `max(dt)` in the Python code instead (and it might make the intention more clear to others).

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                         SELECT q.prob,
+                                q.date,
+                                q.batch_points
+                         FROM (
+                                  SELECT cp.id          as `prob`,
+                                         sub.id         as `subid`,
+                                         sub.date       as `date`,
+                                         tc.points      as `points`,
+                                         tc.batch       as `batch`,
+                                         Min(tc.points) as `batch_points`
```
Ninjaclasher, line 38:
```suggestion
                                         MIN(tc.points) as `batch_points`
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                                           INNER JOIN judge_submissiontestcase tc
+                                                      ON sub.id = tc.submission_id
+                                  GROUP BY cp.id, tc.batch, sub.id
+                              ) q
+                                  INNER JOIN (
+                             SELECT prob, batch, MAX(r.batch_points) max_batch_points
+                             FROM (
+                                      SELECT cp.id          as `prob`,
+                                             tc.batch       as `batch`,
+                                             Min(tc.points) as `batch_points`
```
Ninjaclasher, line 54:
```suggestion
                                             MIN(tc.points) as `batch_points`
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                                       judge_contestsubmission cs
+                                       ON (cs.problem_id = cp.id AND cs.participation_id = %s)
+                                           LEFT OUTER JOIN
+                                       judge_submission sub ON (sub.id = cs.submission_id)
+                                           INNER JOIN judge_submissiontestcase tc
+                                                      ON sub.id = tc.submission_id
+                                  GROUP BY cp.id, tc.batch, sub.id
+                              ) q
+                                  INNER JOIN (
+                             SELECT prob, batch, MAX(r.batch_points) max_batch_points
```
Ninjaclasher, line 50:
```suggestion
                             SELECT prob, batch, MAX(r.batch_points) as max_batch_points
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                                           ON (cs.problem_id = cp.id AND cs.participation_id = %s)
+                                               INNER JOIN
+                                           judge_submission sub ON (sub.id = cs.submission_id)
+                                               INNER JOIN judge_submissiontestcase tc
+                                                          ON sub.id = tc.submission_id
+                                      GROUP BY cp.id, tc.batch, sub.id
+                                  ) r
+                             GROUP BY prob, batch
+                         ) p
+                                       ON p.prob = q.prob AND p.batch = q.batch
```
Ninjaclasher, line 67: I'm not sure if there is a better way to indent this, since it took me quite a while to find where this `ON` is from.

judge/contest_format/ioi16.py
adambenali, line 72: Sure.
```suggestion
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                time = from_database_time(time)
+                if self.config['cumtime']:
+                    dt = (time - participation.start).total_seconds()
+                else:
+                    dt = 0
+
+                if format_data.get(str(problem_id)) is None:
+                    format_data[str(problem_id)] = {'points': 0, 'time': 0}
+                format_data[str(problem_id)]['points'] += subtask_points
+                format_data[str(problem_id)]['time'] = dt
```
adambenali, line 85:
```suggestion
                format_data[str(problem_id)]['time'] = max(dt, format_data[str(problem_id)]['time'])
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+        cumtime = 0
+        score = 0
+        format_data = {}
+
+        with connection.cursor() as cursor:
+            cursor.execute('''
+                SELECT *
+                FROM (
+                         SELECT q.prob,
+                                q.date,
```
adambenali, line 30: The default behavior of `GROUP BY` here is actually undefined. So we need to make sure we take the minimum.
```suggestion
                                MIN(q.date) as `date`,
```

judge/contest_format/ioi16.py
adambenali: Sure.
```suggestion
```

judge/contest_format/ioi16.py
adambenali:
```suggestion
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,98 @@
...
+                                         Min(tc.points) as `batch_points`
+                                  FROM judge_contestproblem cp
+                                           INNER JOIN
+                                       judge_contestsubmission cs
+                                       ON (cs.problem_id = cp.id AND cs.participation_id = %s)
+                                           LEFT OUTER JOIN
+                                       judge_submission sub ON (sub.id = cs.submission_id)
+                                           INNER JOIN judge_submissiontestcase tc
+                                                      ON sub.id = tc.submission_id
+                                  GROUP BY cp.id, tc.batch, sub.id
```
quantum5, line 45: I think you should filter out only finished submissions, i.e. `status = 'D'`. I think this will prevent abortions and other failures from inflating batch points.

adambenali: If, for example, a submission gets aborted right after batch 1 was scored and actually got full score on that batch, should this submission be ignored?

quantum5: I don't think there is a general method to tell if a batch is actually fully graded (because the site doesn't know how many cases are in a batch, or how many batches there are), so I think we should just ignore the submission entirely.

Xyene: To add onto this: that would make it consistent with the other contest formats, too. Right now, if you time your abort request right, it can happen when the judge sends the last test case packet and when it sends its "I'm done grading" packet, so you can end up with an aborted solution that has all of the test cases present (but of course, the site can't know this).

In general, I don't think making aborted solution scoring particularly robust is worth the effort.

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,95 @@
...
+from django.utils.translation import gettext_lazy
+
+from judge.contest_format.ioi import IOIContestFormat
+from judge.contest_format.registry import register_contest_format
+from judge.timezone import from_database_time
+
+
+@register_contest_format('ioi16')
+class IOI16ContestFormat(IOIContestFormat):
+    name = gettext_lazy('IOI16')
```
Xyene, line 11: I'd prefer renaming the old IOI format to "IOI (pre-2017)", `LegacyIOIContestFormat`, etc. I don't have a huge preference as whether to keep the format key as `ioi` and have this new one be different, or whether a migration should be written to update the format name.

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,95 @@
...
+                         ) p
+                                       ON p.prob = q.prob AND p.batch = q.batch
+                         WHERE p.max_batch_points = q.batch_points
+                           AND p.prob = q.prob
+                         GROUP BY q.prob, q.batch
+                     ) best
+            ''', (participation.id, participation.id))
+
+            for problem_id, time, subtask_points in cursor.fetchall():
+                time = from_database_time(time)
```
Xyene, line 73:
```suggestion
                problem_id = str(problem_id)
                time = from_database_time(time)
```

Xyene: (so that repeated casts are not necessary later)

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,95 @@
...
+                    dt = (time - participation.start).total_seconds()
+                else:
+                    dt = 0
+
+                if format_data.get(str(problem_id)) is None:
+                    format_data[str(problem_id)] = {'points': 0, 'time': 0}
+                format_data[str(problem_id)]['points'] += subtask_points
+                format_data[str(problem_id)]['time'] = max(dt, format_data[str(problem_id)]['time'])
+
+            for problem_id in format_data:
```
Xyene, line 84:
```suggestion
            for _, problem_data in format_data.items():
```

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,95 @@
...
+                penalty = format_data[problem_id]['time']
```
Xyene, line 85:
```suggestion
                penalty = problem_data['time']
```

(same for the following line)

judge/contest_format/ioi16.py
```diff
@@ -0,0 +1,95 @@
...
+from judge.contest_format.registry import register_contest_format
+from judge.timezone import from_database_time
+
+
+@register_contest_format('ioi16')
+class IOI16ContestFormat(IOIContestFormat):
+    name = gettext_lazy('IOI16')
+    config_defaults = {'cumtime': False}
+    '''
+        cumtime: Specify True if time penalties are to be computed. Defaults to False.
```
Xyene, line 14: I know this is copied over from other contest formats, but I've never really liked the name. `use_time_penalty` or `time_penalty` seems more explicit.

adambenali: Given that this class inherits from IOIContestFormat, this would imply some other fixes (like overriding the other methods as they use `cumtime` as well). Is that what we want here ?
I think that to be consistent, either all should be changed to `use_time_penalty`, or none.

Xyene: Ah, you're right, I missed the inheritance. Let's not do that then. I wouldn't be opposed to changing this name everywhere in a later PR, though :+1:

PR #1402, opened 2020-05-28
Improve submission lock UI

WallE256:


templates/submission/row.html
```diff
@@ -49,9 +49,13 @@
...
         {% if perms.judge.rejudge_submission and can_edit %} ·
-            <a href="#" onclick="rejudge_submission({{ submission.id }}, event);return false">
-                <i class="fa fa-refresh fa-fw"></i><span class="label">{{ _('rejudge') }}</span>
-            </a>
+            {% if not submission.is_locked %}
+                <a href="#" onclick="rejudge_submission({{ submission.id }}, event);return false">
+                    <i class="fa fa-refresh fa-fw"></i><span class="label">{{ _('rejudge') }}</span>
+                </a>
+            {% else %}
+                <i class="fa fa-refresh fa-fw"></i><span class="grey-label">{{ _('locked') }}</span>
```
Xyene, line 57: I think we should also add some alt text here, like "This submission has been locked, and cannot be rejudged."

templates/submission/row.html
```diff
@@ -49,9 +49,13 @@
...
         {% if perms.judge.rejudge_submission and can_edit %} ·
-            <a href="#" onclick="rejudge_submission({{ submission.id }}, event);return false">
-                <i class="fa fa-refresh fa-fw"></i><span class="label">{{ _('rejudge') }}</span>
-            </a>
+            {% if not submission.is_locked %}
+                <a href="#" onclick="rejudge_submission({{ submission.id }}, event);return false">
+                    <i class="fa fa-refresh fa-fw"></i><span class="label">{{ _('rejudge') }}</span>
+                </a>
+            {% else %}
+                <i class="fa fa-refresh fa-fw"></i><span class="grey-label" title="This submission has been locked, and cannot be rejudged.">{{ _('locked') }}</span>
```
Ninjaclasher, line 57: i18n

PR #1406, opened 2020-05-31
Remove contest editorial column if empty

kevinjycui:
Removes editorial column on a finished contest if there are no public editorials.
To replicate original bug:
- Find or create a contest with only problems that have no public editorials
- Open contest page and look at "Problems" table

Note the empty column at the end of the table, where editorials would be if public editorials existed
Fix:
image

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -217,6 +217,9 @@ class ContestDetail(ContestMixin, TitleMixin, CommentedDetailView):
             .annotate(has_public_editorial=Sum(Case(When(solution__is_public=True, then=1),
                                                     default=0, output_field=IntegerField()))) \
             .add_i18n_name(self.request.LANGUAGE_CODE)
+        context['contest_has_public_editorials'] = any(
+            problem.is_public and problem.has_public_editorial for problem in context['contest_problems']
+        )
         return context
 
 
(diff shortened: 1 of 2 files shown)


judge/views/contests.py
```diff
@@ -217,6 +217,11 @@ def get_context_data(self, **kwargs):
             .annotate(has_public_editorial=Sum(Case(When(solution__is_public=True, then=1),
                                                     default=0, output_field=IntegerField()))) \
             .add_i18n_name(self.request.LANGUAGE_CODE)
+        context['contest_has_public_editorials'] = False
+        for contest_problem in context['contest_problems']:
```
Xyene, line 221: This sounds like a job for `any`, e.g. `context['contest_has_public_editorials'] = any(problem.has_public_editorial for problem in context['contest_problems']`.

Though, doesn't this check have to also include whether the problem is public? L144 in contest.html checks that too.

templates/contest/contest.html
```diff
@@ -84,7 +84,9 @@ <h2 style="margin-bottom: 0.2em"><i class="fa fa-fw fa-question-circle"></i>{{ _
                     <th>{{ _('Points') }}</th>
                     <th>{{ _('AC Rate') }}</th>
                     <th>{{ _('Users') }}</th>
-                    <th></th>
+                    {% if contest_has_public_editorials %}
+                        <th></th>
```
kiritofeng, line 88: Heading should say `Editorials`

PR #1415, opened 2020-06-02
Add utils unit testing

TheAvidDev:
This PR adds unit testing for the following utilities (`judge/utils/`):
 - [ ] `cachedict`
 - [ ] `camo`
 - [ ] `caniuse`
 - [ ] `celery`
 - [ ] `file_cache`
 - [x] `infinite_paginator`
 - [ ] `mathoid`
 - [ ] `opengraph`
 - [ ] `problem_data`
 - [ ] `problems`
 - [x] `ranker`
 - [ ] `raw_sql`
 - [ ] `recaptcha`
 - [x] `stats`
 - [x] `strings`
 - [ ] `subscription`
 - [ ] `texoid`
 - [ ] `tickets`
 - [ ] `timedelta`
 - [ ] `two_factor`
   - [x] `webauthn_encode`
   - [x] `webauthn_decode`
   - [ ] `WebAuthnJSONEncoder`
 - [x] `unicode`
 - [ ] `views`

`pwned` is not in this list because it is a third party library, with tests available [here](https://github.com/ubernostrum/pwned-passwords-django/tree/master/tests).


judge/utils/tests/test_two_factor.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class TwoFactorTestCase(SimpleTestCase):
+    def test_webauthn_decode(self):
+        self.assertEqual(webauthn_decode(''), b'')
+        self.assertEqual(webauthn_decode('AA'), bytes(range(1)))
+        self.assertEqual(webauthn_decode('AAE'), bytes(range(2)))
+        self.assertEqual(webauthn_decode('AAEC'), bytes(range(3)))
+        self.assertEqual(webauthn_decode(
+            'AAECAwQFBgcICQoLDA0ODxAREhMUFRYXGBkaGxwdHh8gISIjJCUmJygpKissLS4vMDEyMzQ1Njc4OTo7PD0-Pw'), bytes(range(64)))
```
quantum5, line 13: Should probably exercise `_` handling...

TheAvidDev: Done

judge/utils/tests/test_cachedict.py
```diff
@@ -0,0 +1,20 @@
...
+from judge.utils.cachedict import CacheDict
+
+
+class CacheDictTestCase(SimpleTestCase):
+    def setUp(self):
+        self.test_dict = {'a': True}
+        self.cache_dict = CacheDict(lambda id: self.test_dict[id])
+
+    def test_original_item_in_cachedict(self):
+        self.assertFalse('a' in self.cache_dict)
```
Ninjaclasher, line 12: does `self.assertNotIn` not work here (and the other places below)?

TheAvidDev: Fixed

TheAvidDev:
I'd rather not have this get stale.

judge/utils/tests/test_cachedict.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+class CacheDictTestCase(SimpleTestCase):
+    def setUp(self):
+        self.test_dict = {'a': True}
+        self.cache_dict = CacheDict(lambda id: self.test_dict[id])
+
+    def test_original_item_in_cachedict(self):
+        self.assertNotIn('a', self.cache_dict)
+        self.assertTrue(self.cache_dict['a'])
```
quantum5, line 13: To quote the python documentation:

> Note that this is equivalent to `bool(expr) is True` and not to `expr is True` (use `assertIs(expr, True)` for the latter).

judge/utils/tests/test_cachedict.py
```diff
@@ -0,0 +1,20 @@
...
+        self.assertIn('a', self.cache_dict)
+
+    def test_new_item_in_cachedict(self):
+        self.test_dict['b'] = True
+        self.assertNotIn('b', self.cache_dict)
+        self.assertTrue(self.cache_dict['b'])
```
quantum5, line 19: To quote the python documentation:

> Note that this is equivalent to `bool(expr) is True` and not to `expr is True` (use `assertIs(expr, True)` for the latter).

judge/utils/tests/test_cachedict.py
```diff
@@ -0,0 +1,20 @@
...
+    def setUp(self):
+        self.test_dict = {'a': True}
+        self.cache_dict = CacheDict(lambda id: self.test_dict[id])
+
+    def test_original_item_in_cachedict(self):
+        self.assertNotIn('a', self.cache_dict)
+        self.assertTrue(self.cache_dict['a'])
+        self.assertIn('a', self.cache_dict)
+
+    def test_new_item_in_cachedict(self):
```
quantum5, line 16: What's the point of this test case exactly? `self.cache_dict` is refreshed every case.

judge/utils/tests/test_camo.py
```diff
@@ -0,0 +1,92 @@
...
+from hashlib import sha1
+
+from django.test import SimpleTestCase
+from lxml.html import HTMLParser, fromstring, tostring
+
+from judge.utils.camo import CamoClient
+from judge.utils.unicode import utf8bytes
+
+
+# TODO: Test the setting of judge.utils.camo.client depending on settings variables. I've tried my hardest... but the
```
quantum5, line 11: I would be very surprised if https://docs.djangoproject.com/en/2.2/topics/testing/tools/#overriding-settings doesn't work.

judge/utils/tests/test_camo.py
```diff
@@ -0,0 +1,92 @@
...
+            self.http_camo.excluded[0],
+            self.test_url,
+            self.test_url_http,
+            self.test_url_https,
+        )
+        for camo in (self.http_camo, self.https_camo):
+            # Test that every element that's supposed to get updated is updated to the rewritten element
+            for element in update_elements:
+                for url in urls:
+                    html = fromstring(element.format(url), parser=HTMLParser())
```
quantum5, line 82: Why do you need the `parser=HTMLParser()`?

judge/utils/tests/test_camo.py
```diff
@@ -0,0 +1,92 @@
...
+                    camo.update_tree(html)
+                    self.assertEqual(tostring(html, encoding='unicode'), element.format(camo.rewrite_url(url)))
+
+            # Test that other elements are left untouched
+            for element in ignore_elements:
+                for url in urls:
+                    html_orig = fromstring(element.format(url), parser=HTMLParser())
```
quantum5, line 89: Why do you need the `parser=HTMLParser()`?

judge/utils/tests/test_camo.py
```diff
@@ -0,0 +1,92 @@
...
+                    html = html_orig
+                    camo.update_tree(html)
+                    self.assertEqual(html, html_orig)
```
quantum5, line 92: Why aren't you simply converting it back to string and compare with `assertHTMLEqual`?

judge/utils/tests/test_camo.py
```diff
@@ -0,0 +1,92 @@
...
+            self.test_url_http,
+            self.test_url_https,
+        )
+        for camo in (self.http_camo, self.https_camo):
+            # Test that every element that's supposed to get updated is updated to the rewritten element
+            for element in update_elements:
+                for url in urls:
+                    html = fromstring(element.format(url), parser=HTMLParser())
+                    camo.update_tree(html)
+                    self.assertEqual(tostring(html, encoding='unicode'), element.format(camo.rewrite_url(url)))
```
quantum5, line 84: Use `assertHTMLEqual`.

judge/utils/tests/test_file_cache.py
```diff
@@ -0,0 +1,111 @@
...
+            self.cache.get_url('foo', 'b'),
+        )
+
+        # Test that these files are actually gzip files
+        with gzip.open(self.gzip_cache.get_path('foo', 'a.gz'), 'rb') as f:
+            f.read()
+        with gzip.open(self.gzip_cache.get_path('foo', 'b.gz'), 'rb') as f:
+            f.read()
+
+    def tearDown(self):
```
quantum5, line 109: Put `tearDown` near `setUp` improves readability.

judge/utils/tests/test_file_cache.py
```diff
@@ -0,0 +1,111 @@
...
+class FileCacheTestCase(SimpleTestCase):
+    def setUp(self):
+        self.temp_dir = tempfile.mkdtemp()
+        self.cache = HashFileCache(self.temp_dir, 'https://example.com')
+        self.gzip_temp_dir = tempfile.mkdtemp()
+        self.gzip_cache = HashFileCache(self.gzip_temp_dir, 'https://example.com', gzip=True)
+
+        # This acts as a test for create() as well as setup for other tests
+        self.cache.create('foo')
+        self.cache.create('foo')
```
quantum5, line 22: Why do this twice?

judge/utils/tests/test_file_cache.py
```diff
@@ -0,0 +1,111 @@
...
+            ('foo/', '/bar'),
+            ('/foo', 'bar/'),
+            ('/foo', '/bar/'),
+        ]
+        for hash, file in urls:
+            self.assertTrue(self.cache.get_url(hash, file), urljoin(self.cache.url, '%s/%s' % (hash, file)))
+
+    def test_read_file(self):
+        # The directory removal in tear down will clean up this file, so we don't have to cleanup here
+        with tempfile.NamedTemporaryFile(dir=os.path.join(self.temp_dir, 'foo'), delete=False) as f:
```
quantum5, line 72: Do we really need to use `NamedTemporaryFile` when we can just create a file with a fixed name?

judge/utils/tests/test_file_cache.py
```diff
@@ -0,0 +1,111 @@
...
+        self.assertIsNone(self.gzip_cache.cache_data('foo', 'a', b'bar', url=False))
+        self.assertEqual(self.gzip_cache.read_data('foo', 'a'), b'bar')
+        self.assertEqual(
+            self.gzip_cache.cache_data('foo', 'b', b'bar', url=True, gzip=True),
+            self.cache.get_url('foo', 'b'),
+        )
+
+        # Test that these files are actually gzip files
+        with gzip.open(self.gzip_cache.get_path('foo', 'a.gz'), 'rb') as f:
+            f.read()
```
quantum5, line 105: Maybe assert that it's actually the data we expect?

judge/utils/tests/test_opengraph.py
```diff
@@ -0,0 +1,66 @@
...
+from django.test import SimpleTestCase
+
+from judge.utils.opengraph import generate_opengraph
+
+
+HTML = '''
+%s
+
+![Alt text](https://example.com/image.png)
+''' % ('word\\\\(\\\\)\\\\[\\\\] ' * 50)  # Make sure literal backslashes followed by brackets are removed
```
quantum5, line 11: Why not use a raw string?

judge/utils/tests/test_opengraph.py
```diff
@@ -0,0 +1,66 @@
...
+        cache.delete('foo')
+
+    def test_opengraph_cache(self):
+        with self.assertRaises(TypeError):
+            generate_opengraph('foo', None, None)
+        cache.set('foo', True, 1000)
+        # This would through a TypeError if it wasn't directly taken from the cache
+        self.assertTrue(generate_opengraph('foo', None, None))
+
+    def test_generaete_opengraph(self):
```
quantum5, line 43:
```suggestion
    def test_generate_opengraph(self):
```

judge/utils/tests/test_opengraph.py
```diff
@@ -0,0 +1,66 @@
...
+        # Django doesn't setup a test cache (yet*) so we make sure to clear it here.
+        # [*] https://docs.djangoproject.com/en/3.0/topics/testing/overview/#other-test-conditions
+        cache.delete('foo')
+
+    def test_opengraph_cache(self):
+        with self.assertRaises(TypeError):
+            generate_opengraph('foo', None, None)
+        cache.set('foo', True, 1000)
+        # This would through a TypeError if it wasn't directly taken from the cache
+        self.assertTrue(generate_opengraph('foo', None, None))
```
quantum5, line 41: To quote the python documentation:

> Note that this is equivalent to `bool(expr) is True` and not to `expr is True` (use `assertIs(expr, True)` for the latter).

judge/utils/tests/test_camo.py
TheAvidDev: I've tried it and the other options I could find to. Perhaps it has something to do with how python modules are loaded in general, I'm not really sure.

I'll try again though, but overall this probably isn't too important considering it would just check whether a client gets set or not.

judge/utils/tests/test_file_cache.py
TheAvidDev: To make sure that creating a file that already exists doesn't throw the error. It tests the `False` outcome of [this](https://github.com/DMOJ/online-judge/blob/master/judge/utils/file_cache.py#L17).

judge/utils/tests/test_file_cache.py
TheAvidDev: I suppose we could, but it would have to be unique across all tests since there's an option to run tests in parallel.

judge/utils/tests/test_file_cache.py
quantum5: Shouldn't this be a separate test case?

judge/utils/tests/test_file_cache.py
quantum5: `setUp` and `tearDown` are run once per test case, so that's not a problem.

judge/utils/tests/test_ranker.py
```diff
@@ -0,0 +1,43 @@
...
+
+        self.assertEqual(list(tie_ranker([])), [])
+        self.assertEqual(list(tie_ranker([a])), [(1, a)])
+        self.assertEqual(list(tie_ranker([a, b])), [(1, a), (2, b)])
+        self.assertEqual(list(tie_ranker([c, d])), [(1.5, c), (1.5, d)])
+        self.assertEqual(list(tie_ranker([b, c, d])), [(1, b), (2.5, c), (2.5, d)])
+        self.assertEqual(list(tie_ranker([c, d, e])), [(2, c), (2, d), (2, e)])
+        self.assertEqual(list(tie_ranker([d, e, f])), [(1.5, d), (1.5, e), (3, f)])
+        self.assertEqual(list(tie_ranker([a, b, c, d, e, f])), [(1, a), (2, b), (4, c), (4, d), (4, e), (6, f)])
+        self.assertEqual(list(ranker([d, c, b, a], key=attrgetter('neg'))), [(1, d), (1, c), (3, b), (4, a)])
```
TheAvidDev, line 43: Just noticed this is wrong. Will fix as well.

judge/utils/tests/test_file_cache.py
TheAvidDev: Agreed.

judge/utils/tests/test_file_cache.py
TheAvidDev: Wouldn't there be a possible race conditions?
---
On another note, in basically all my test cases, I don't need the `setUp` and `tearDown` to run for every test. (In fact I didn't realize it happened until quite late.) This would also very very marginally increase performance.

quantum5: It's not a problem because `self.temp_dir` is different for every test case. I wouldn't recommend (accidentally or not) depending on the test execution order in any particular group of test cases, so it's better to setup and teardown every case.

TheAvidDev: Oh right, it's not an issue for the very reason that `setUp` and `tearDown` happen every test.
> I wouldn't recommend (accidentally or not) depending on the test execution order in any particular group of test cases, so it's better to setup and teardown every case.

Agreed.

quantum5:
What's going on with this one?

PR #1416, opened 2020-06-02
Add unittests for models

Ninjaclasher:


Ninjaclasher:
Todo:
 - [ ] Add tests for `Comment`
 - [ ] Add fixture (?) for users, profiles, permissions, organizations
 - [ ] Fix wrong `Problem.is_accessible_by` method
 - [ ] Reduce the amount of copy pasting for `Contest`, `Problem`, and `BlogPost` method testing

Ninjaclasher:
This PR covers unittests for most of the models.

Models excluded:
 - Comments (this will be done in a future PR)
 - Problem Data (this should probably just be done alongside unittests for the views)
 - Judge/Runtime (this requires a running bridge/judges for thorough testing)

judge/models/tests/util.py
```diff
@@ -0,0 +1,332 @@
+from django.contrib.auth.models import AnonymousUser, Permission, User
+from django.utils import timezone
+
+from judge.models import (
```
TheAvidDev, line 4: Same thing here.

judge/models/tests/test_submission.py
```diff
@@ -0,0 +1,131 @@
+from django.test import TestCase
+
+from judge.models import ContestSubmission, Language, Submission
+from judge.models.tests.util import (
```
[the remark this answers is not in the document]
Ninjaclasher: Updated.

judge/models/tests/util.py
```diff
@@ -0,0 +1,330 @@
...
+
+    def _test_object_user(self, obj, user, methods):
+        for method, func in methods.items():
+            with self.subTest(method=method):
+                func(
+                    getattr(obj, method)(user),
+                    msg='Method "%s" failed for user "%s", object "%s".' % (method, user.username, obj),
+                )
+
+    def _test_object(self, obj, data):
```
quantum5, line 327: This name is too generic. This is really for testing methods that takes a user, no?

judge/models/tests/util.py
```diff
@@ -0,0 +1,330 @@
...
+
+    @classmethod
+    def setUpTestData(self):
+        self.users = {
+            'superuser': create_user(
+                username='superuser',
+                is_superuser=True,
+                is_staff=True,
+            ),
+            'staff_blogpost_edit_own': create_user(
```
quantum5, line 240: Is this user relevant outside of blog testing, for example? Maybe it should be more limited in scope. Similar issues apply to the other users.

judge/models/tests/util.py
Ninjaclasher: Updated the name to `_test_object_methods_with_users`

judge/models/tests/util.py
Ninjaclasher: I've moved most of the users to their respective test case. The users with problem permissions are used in both `ProblemTestCase` and `SubmissionTestCase`. I'm not sure if these users should be duplicated in both test cases or kept in `CommonDataMixin`.

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+            contest.full_clean()
+        contest.format_config = {}
+        with self.assertRaisesRegex(ValidationError, 'Contest problem label script'):
+            contest.full_clean()
+        contest.problem_label_script = '''
+            function(n)
+                return n
+            end
+        '''
+        # Test for caching
```
TheAvidDev, line 350:
```suggestion
        # Test bad label script catching
```

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+    def test_basic_tag(self):
+        self.assertEqual(str(self.basic_tag), self.basic_tag.name)
+        self.assertEqual(self.basic_tag.text_color, '#000')
+
+    def test_dark_tag(self):
+        self.assertEqual(self.dark_tag.text_color, '#fff')
+
+
+class MinValueOrNoneValidatorTestCase(SimpleTestCase):
+    def test_success(self):
```
TheAvidDev, line 434: This should probably be split up into individual tests with more verbose names to fall in line with similar test like those for DiggPaginator and InfinitePaginator.

judge/models/tests/test_profile.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertIsInstance(self.profile.api_token, str)
+
+        user_id, raw_token = struct.unpack('>I32s', base64.urlsafe_b64decode(token))
+
+        self.assertEqual(self.users['normal'].id, user_id)
+        self.assertEqual(len(raw_token), 32)
+
+        self.assertTrue(
+            hmac.compare_digest(
+                hmac.new(force_bytes(settings.SECRET_KEY), msg=force_bytes(raw_token), digestmod='sha256').hexdigest(),
```
TheAvidDev, line 77: I know this also tests that the secret key can actually be made into bytes, but putting this inside of `test_generate_api_token` will throw off debugging. A separate test should probably be made to ensure the secret key can actually be converted into bytes and this test should just use a hard coded string.

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+    def test_contest_clean(self):
+        _now = timezone.now()
+        contest = create_contest(
+            key='contest',
+            start_time=_now,
+            end_time=_now - timezone.timedelta(days=1),
+            problem_label_script='invalid',
+            format_config={'invalid': 'invalid'},
+        )
+        with self.assertRaisesRegex(ValidationError, 'ended before it starts'):
```
TheAvidDev, line 337: Perhaps add a second test to make sure a contest can't end and start at the same time.

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+            virtual=1,
+        )
+
+        self.assertFalse(participation.live)
+        self.assertFalse(participation.spectate)
+        self.assertEqual(participation.start, participation.real_start)
+        self.assertIsInstance(participation.end_time, timezone.datetime)
+
+
+class ContestTagTestCase(TestCase):
```
TheAvidDev, line 413: Should this test that `text_color` caches its values and uses them on repeated runs?

judge/models/tests/test_problem.py
```diff
@@ -0,0 +1,183 @@
+from django.test import TestCase
+
+from judge.models import Language, LanguageLimit, Problem
+from judge.models.tests.util import CommonDataMixin, create_problem, create_problem_type
+
+
+class ProblemTestCase(CommonDataMixin, TestCase):
```
TheAvidDev, line 7: From codecov analysis, test changing of problem `code` and that when `case_points == case_total`, `result_class_from_code` returns `AC`.

judge/models/tests/test_profile.py
```diff
@@ -0,0 +1,125 @@
...
+        self.assertIn(self.profile.id, self.organizations['open'])
+
+        self.assertNotIn(self.profiles['superuser'], self.organizations['open'])
+        self.assertNotIn(self.profiles['superuser'].id, self.organizations['open'])
+
+        with self.assertRaisesRegex(TypeError, 'Organization membership test'):
+            'aaaa' in self.organizations['open']
+
+
+class ProfileTestCase(CommonDataMixin, TestCase):
```
TheAvidDev, line 32: From codecov analysis, maybe check `__str__` since you do it in other models. This also applies to `SubmissionSource`.

judge/models/tests/test_contest.py
TheAvidDev: I'd actually like to know this for other tests as well (such as utils).

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
+from django.core.exceptions import ValidationError
+from django.test import SimpleTestCase, TestCase
+from django.utils import timezone
+
+from judge.models import Contest, ContestParticipation, ContestTag
+from judge.models.contest import MinValueOrNoneValidator
+from judge.models.tests.util import CommonDataMixin, create_contest, create_contest_participation, create_user
+
+
+class ContestTestCase(CommonDataMixin, TestCase):
```
TheAvidDev, line 10: From codecov analysis, you should test the `raise AttributeError` of `get_label_for_problem`.

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+            is_visible=True,
+            is_private=True,
+        )
+
+    def setUp(self):
+        self.profiles['normal'].refresh_from_db()
+
+    def test_basic_contest(self):
+        self.assertTrue(self.basic_contest.show_scoreboard)
+        self.assertIsInstance(self.basic_contest.contest_window_length, timezone.timedelta)
```
quantum5, line 94: Why not assert the actual values?

judge/models/tests/test_contest.py
```diff
@@ -0,0 +1,454 @@
...
+    def test_contests_list(self):
+        for name, user in self.users.items():
+            with self.subTest(user=name):
+                # We only care about consistency between Contest.is_accessible_by and Contest.get_visible_contests
+                contest_keys = []
+                for contest in Contest.objects.prefetch_related('organizers', 'private_contestants', 'organizations'):
+                    if contest.is_accessible_by(user):
+                        contest_keys.append(contest.key)
+
+                self.assertListEqual(
```
quantum5, line 323: Have you tried `assertCountEqual`? This applies for every instance of `assertListEqual` in which you sorted.

judge/models/tests/test_contest.py
Ninjaclasher: Done

judge/models/tests/test_contest.py
Ninjaclasher: Done

judge/models/tests/test_contest.py
Ninjaclasher: Done

judge/models/tests/test_contest.py
Ninjaclasher: From some discussion on Slack, we decided that it's not worth to test this, as some Python hacks would be needed to grab the `cache` from the `test_color` property.

judge/models/tests/test_problem.py
Ninjaclasher: Done the testing `result_class_from_code` returning `AC`. However, I think that testing problem code changing ought to be done alongside `ProblemData` testing, since there is no way to really test this properly currently (unless we just test that `save()` works?)

judge/models/tests/test_profile.py
Ninjaclasher: Done

judge/models/tests/test_contest.py
Ninjaclasher: Even relooking that the `lupa` documentation, I can't find a way to cause `AttributeError` to be raised. This is because the argument passed in is an integer, so there are no attributes you can even access.

judge/models/tests/test_contest.py
Ninjaclasher: Done for this (since the `timedelta` is fixed). However, I don't think we should assert the actual values for the ones that are dependant on the current time (e.g `time_remaining)`.

judge/models/tests/test_contest.py
Ninjaclasher: Done

judge/models/tests/test_profile.py
Ninjaclasher: I'm not sure what you mean by this. What hard coded string should be used? The API token is hashed with the `SECRET_KEY`, so logically, this should also use the `SECRET_KEY` for hashing and comparison.

I also don't see a case of the `SECRET_KEY` ever becoming not convertable to bytes.

judge/models/tests/test_problem.py
TheAvidDev: Yep, testing this alongside `ProblemData` makes sense.

judge/models/tests/test_profile.py
TheAvidDev: Right, this is fine. Django [docs](https://docs.djangoproject.com/en/3.0/ref/settings/#std:setting-SECRET_KEY) note that you shouldn't just assume it is, but this could just be telling users not to manually check if it's str or bytes.

judge/models/tests/test_contest.py
TheAvidDev: Hm, I guess we have it there just in case. Sounds good though.

judge/models/tests/test_contest.py
TheAvidDev: Could you not get an accurate calculation by using the Contest's `_now` cached property?

Ninjaclasher: Personally, I don't think it is worth it to check for the exact match in these cases. If you're using the Contest's `_now` property, you basically end up rewriting that method's calculation as a test, which doesn't really test what the method "should" return.

PR #1427, opened 2020-06-04
Update robots.txt with new URL endpoints; #504

### LINKED ISSUE #504: robots.txt rework
We have not touched that file, despite adding new URLs for the past two years. Something tells me it's not relevant anymore...

Ninjaclasher:
Closes #504.


robots.txt
```diff
@@ -1,21 +1,50 @@
 User-agent: *
-Disallow: /admin/
-Disallow: /accounts/
-Disallow: /problem/*/pdf
+Disallow: /500
```
Xyene, line 2: Minor nit: we should probably sort these so that checking if a URL is in this file is not O(n) for a human. `sort < robots.txt` should do that for you; you'd just need to move `User-agent` back up.

PR #1430, opened 2020-06-05
Hide `locked` word on submissions list for small-width screens

Ninjaclasher:


Ninjaclasher:
> Looks fine, but greying out the icon would be a nice bonus.

The current icon is black. Should the black be changed to grey?

E.g.
image

quantum5:
Yes, that would be a nice change.

Ninjaclasher:
Done

PR #1433, opened 2020-06-06
Use a custom contest object manager to annotate _now; #1115

### LINKED ISSUE #1115: Front page 500s when requesting at the exact moment a contest ends
```
  File "/code/site/siteenv/lib/python3.6/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/code/site/siteenv/lib/python3.6/site-packages/django/core/handlers/base.py", line 156, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/code/site/siteenv/lib/python3.6/site-packages/django/core/handlers/base.py", line 154, in _get_response
    response = response.render()
  File "/code/site/siteenv/lib/python3.6/site-packages/django/template/response.py", line 106, in render
    self.content = self.rendered_content
  File "/code/site/siteenv/lib/python3.6/site-packages/django/template/response.py", line 83, in rendered_content
    content = template.render(context, self._request)
  File "/code/site/siteenv/lib/python3.6/site-packages/django_jinja/backend.py", line 106, in render
    return mark_safe(self.template.render(context))
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "./templates/home.html", line 1, in top-level template code
    {% extends "blog/list.html" %}
  File "./templates/blog/list.html", line 1, in top-level template code
    {% extends "base.html" %}
  File "./templates/base.html", line 282, in top-level template code
    <div id="content-body">{% block body %}{% endblock %}</div>
  File "./templates/blog/list.html", line 139, in block "body"
    {{ _('Ends in %(countdown)s.', countdown=contest.time_before_end|as_countdown) }}
  File "/code/site/siteenv/lib/python3.6/site-packages/django_jinja/library.py", line 92, in _wrapper
    data = render_to_string(template, fn(*args, **kwargs))
  File "/code/site/siteenv/lib/python3.6/site-packages/django/template/loader.py", line 62, in render_to_string
    return template.render(context, request)
  File "/code/site/siteenv/lib/python3.6/site-packages/django_jinja/backend.py", line 106, in render
    return mark_safe(self.template.render(context))
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/code/site/siteenv/lib/python3.6/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "./templates/time-remaining-fragment.html", line 1, in top-level template code
    <span data-secs="{{ countdown|seconds }}" class="time-remaining">
  File "/code/site/judge/jinja2/timedelta.py", line 22, in seconds
    return timedelta.total_seconds()
AttributeError: 'NoneType' object has no attribute 'total_seconds'
```
This happens if you request at *just* the right moment such that the contest gets filtered https://github.com/DMOJ/site/blob/master/judge/views/blog.py#L82 but returns `None` in https://github.com/DMOJ/site/blob/master/judge/models/contest.py#L169

Ninjaclasher:
Fixes #1115.

### DIFF
--- a/judge/models/contest.py
+++ b/judge/models/contest.py
@@ -2,6 +2,7 @@ from django.core.exceptions import ValidationError
 from django.core.validators import MinValueValidator, RegexValidator
 from django.db import models, transaction
 from django.db.models import CASCADE, Q
+from django.db.models.functions import Now
 from django.urls import reverse
 from django.utils import timezone
 from django.utils.functional import cached_property
@@ -51,6 +52,11 @@ class ContestTag(models.Model):
     class Meta:
         verbose_name = _('contest tag')
         verbose_name_plural = _('contest tags')
+
+
+class ContestManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().annotate(_now=Now())
 
 
 class Contest(models.Model):
@@ -129,6 +135,8 @@ class Contest(models.Model):
     is_locked = models.BooleanField(verbose_name=_('contest lock'), default=False,
                                     help_text=_('Prevent submissions from this contest from being rejudged.'))
 
+    objects = ContestManager()
+
     @cached_property
     def format_class(self):
         return contest_format.formats[self.format_name]
@@ -198,6 +206,7 @@ class Contest(models.Model):
     def contest_window_length(self):
         return self.end_time - self.start_time
 
+    # This property is still required for when a contest is fetched from a related object.
     @cached_property
     def _now(self):
         # This ensures that all methods talk about the same now.
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -12,7 +12,7 @@ from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMix
 from django.core.cache import cache
 from django.core.exceptions import ImproperlyConfigured, ObjectDoesNotExist
 from django.db import IntegrityError
-from django.db.models import Case, Count, FloatField, IntegerField, Max, Min, Q, Sum, Value, When
+from django.db.models import Case, Count, F, FloatField, IntegerField, Max, Min, Q, Sum, Value, When
 from django.db.models.expressions import CombinedExpression
 from django.http import Http404, HttpResponse, HttpResponseBadRequest, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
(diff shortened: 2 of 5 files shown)


judge/views/contests.py
```diff
@@ -102,7 +98,7 @@ def get_context_data(self, **kwargs):
         context['active_participations'] = active
         context['current_contests'] = present
         context['future_contests'] = future
-        context['now'] = self._now
+        context['now'] = timezone.now()
```
quantum5, line 101: I don't like this. We should pull the now from the queryset and only use `timezone.now()` as a last resort.

judge/models/contest.py
```diff
@@ -53,6 +54,11 @@ class Meta:
         verbose_name_plural = _('contest tags')
 
 
+class ContestManager(models.Manager):
+    def get_queryset(self):
+        return super().get_queryset().annotate(_now=Now())
```
quantum5, line 59: I am unconvinced whether SQL `Now()` is the best approach. Hypothetically, if we were to run separate queries for past contests for pagination purposes, the time would be desynchronized between different queries. If instead, you used `timezone.now()`, we could do `Contest.objects.all()` and then filter separately, and they will all have the same time.

judge/views/contests.py
Ninjaclasher: The only use of this `now` is for the contest calendar (only the month and the year). I'm not sure how to pull from the queryset for the calendar.

quantum5: If you follow used `timezone.now()` you could just pull it out of the queryset. If not, you can set a variable to `timezone.now()` and then overwrite in the loop above.

PR #1435, opened 2020-06-06
Add keyboard shortcut for triggering abort action; #1191

### LINKED ISSUE #1191: Add keyboard shortcut for triggering abort action
It is sometimes annoying to click when results are scrolling fast. This should probably be accompanied with a tooltip (like on the judge status page) informing users of the keyboard shortcut.

Ninjaclasher:
Fixes #1191.

### DIFF
--- a/templates/submission/status.html
+++ b/templates/submission/status.html
@@ -52,6 +52,18 @@
         });
         </script>
     {% endif %}
+    {% if not submission.is_graded %}
+        <script type="text/javascript">
+            $(function () {
+                $(document).keydown(function(e) {
+                    // Ctrl-Enter or Command-Enter
+                    if ((e.metaKey || e.ctrlKey) && e.which == 13) {
+                        $('#abort-button form').submit();
+                    }
+                });
+            });
+        </script>
+    {% endif %}
 {% endblock %}
 
 {% block body %}
@@ -83,7 +95,8 @@
                 <br>
                 <hr class="half-hr">
                 <br>
-                <form action="{{ url('submission_abort', submission.id) }}" method="post">
+                <form action="{{ url('submission_abort', submission.id) }}" method="post"
+                      title="{{ _('Press Ctrl-Enter or Command-Enter to abort.') }}">
                     {% csrf_token %}
                     <input style="float:left" type="submit" value="{{ _('Abort') }}" class="button">
                     <br><br>
--- a/templates/submission/status.html
+++ b/templates/submission/status.html
@@ -2,54 +2,55 @@
 {% block js_media %}
     <script type="text/javascript" src="{{ static('event.js') }}"></script>
     {% if not submission.is_graded and last_msg %}
-        <script type="text/javascript">$(function () {
-            var blocked = false, request = false;
-            var list = $('#test-cases');
+        <script type="text/javascript">
+            $(function () {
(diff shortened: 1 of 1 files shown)


templates/submission/status.html
```diff
@@ -52,6 +53,17 @@
         });
         </script>
     {% endif %}
+    {% if not submission.is_graded %}
+        <script type="text/javascript">
+            $(function () {
+                $(document).keydown(function(ev) {
```
TheAvidDev, line 59:
```suggestion
                $(document).keydown(function(e) {
```
We use `e` for events everywhere else, unless there's another reason?

templates/submission/status.html
```diff
@@ -52,6 +53,17 @@
...
+                    if ((ev.metaKey || ev.ctrlKey) && ev.which == 13) {
```
TheAvidDev, line 60:
```suggestion
                    if ((e.metaKey || e.ctrlKey) && e.which == 13) {
```
Same as above.

templates/submission/status.html
Ninjaclasher: This JS was copied from somewhere else in this repository, but I can't seem to find it anymore. Will change to `e`.

TheAvidDev:
Actually, it seems like the better way to check keys is [`KeyboardEvent.code`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code), although it's [not supported](https://caniuse.com/#search=KeyboardEvent.code) by IE or mobile browsers. Not sure if we should switch to it.

Ninjaclasher:
This is how the DMOJ has always handled keyboard shortcuts.

TheAvidDev:
> This is how the DMOJ has always handled keyboard shortcuts.

Sounds alright then, but it _is_ [officially deprecated](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode) so we should eventually switched.

templates/submission/status.html
```diff
@@ -2,7 +2,8 @@
 {% block js_media %}
     <script type="text/javascript" src="{{ static('event.js') }}"></script>
     {% if not submission.is_graded and last_msg %}
-        <script type="text/javascript">$(function () {
+        <script type="text/javascript">
+        $(function () {
```
Xyene, line 6: If we're fixing the spacing here, might as well fix the indentation too.

templates/submission/status.html
```diff
@@ -52,6 +53,15 @@
         });
         </script>
     {% endif %}
+    <script type="text/javascript">
+        $(function () {
+            $(document).keydown(function(e) {
+                if ((e.metaKey || e.ctrlKey) && e.which == 13) {
```
Xyene, line 59: Should have a comment for those not fluent in keycode numbers :)

Ninjaclasher:
> Otherwise looks great! Perhaps we can get a screenshot of what the form title looks like?

@TheAvidDev 
image

(I can't get the cursor to show up in the screenshot.)

templates/submission/status.html
Ninjaclasher: Done

templates/submission/status.html
Ninjaclasher: Done

PR #1442, opened 2020-06-10
Switch to a tab layout for submission detail views

Ninjaclasher:
image

image


kiritofeng:
Would resubmit as a tab work? (i.e. I think that clicking on resubmit would move to a page without tabs), and I feel that making it a button would be better...

Ninjaclasher:
Closing as this PR as this doesn't seem like something we want to do.

PR #1446, opened 2020-06-12
Use `has_other_pages()` to check if the pagination bar should be shown

Ninjaclasher:
...Rather than the number of pages since the infinite paginator
returns a constant 1e3000 even if there is only 1 page.


quantum5:
I mean, the infinite paginator probably shouldn't return infinity if it knows the number of pages, but this change will make it look more elegant than `> 1`.

PR #1448, opened 2020-06-12
Add WCIPEG account merging

fluix-dev:
Merge Process:
 - WCIPEG user logs in to DMOJ with their account and sees a button on their profile to merge.
 - WCIPEG user fills in the name of a native DMOJ user to merge into.
 - An email is sent to the DMOJ user to accept the merge.
 - WCIPEG user clicks this link, has their data merged, get logged out, and deactivated.
 - DMOJ user now has the WCIPEG user's data and can be logged in to as before.

TODO:
 - [x] Request Form
    - [x] Scary JavaScript confirm dialog
    - [ ] ~~Separate AJAX endpoint to filter out peg users with select2 (or js that does this frontend)?~~
    - [x] Generating token
    - [x] Email sending functionality
    - [x] ReCaptcha
 - [x] Activation form
    - [x] Verify acc_peg login
    - [x] Log out user
    - [x] Merge user
    - [x] `generic_message` success message
 - [x] `PEGUser` model
    - [x] `user` `OneToOneField` to WCIPEG account
    - [x] `merge_user` `OneToOneField` (nullable) to DMOJ account


judge/forms.py
```diff
@@ -230,6 +232,21 @@ def clean(self):
...
 
 
+class WCIPEGMergeRequestForm(Form):
+    handle = CharField(max_length=50, validators=[UnicodeUsernameValidator()])
+
+    def clean_handle(self):
+        username = self.cleaned_data['handle']
+        if not (User.objects.filter(username=username).exists() and User.objects.get(username=username).is_active):
+            raise ValidationError(_("Account doesn't exist."))
+        if User.objects.get(username=username).is_peg():
```
Ninjaclasher, line 242: Isn't `is_peg` a property in `Profile`?

judge/models/profile.py
```diff
@@ -211,6 +211,10 @@ def css_class(self):
     def webauthn_id(self):
         return hmac.new(force_bytes(settings.SECRET_KEY), msg=b'webauthn:%d' % (self.id,), digestmod='sha256').digest()
 
+    @cached_property
+    def is_peg(self):
+        return self.user.email.endswitch('@accounts.wcipeg.com')
```
Ninjaclasher, line 216: `endswitch`

templates/user/edit-profile.html
```diff
@@ -190,6 +190,12 @@
 {% block body %}
     <div class="centered-form">
         <form id="edit-form" action="" method="post" class="form-area">
+            {% if request.profile.is_peg %}
+                <div class="alert alert-info">
```
TheAvidDev, line 194: Should this entire section be a link? It will also make it easier to use i18n.

judge/forms.py
```diff
@@ -230,6 +232,22 @@ def clean(self):
...
 
+class WCIPEGMergeRequestForm(Form):
+    handle = CharField(max_length=50, validators=[UnicodeUsernameValidator()])
+
+    def clean_handle(self):
+        username = self.cleaned_data['handle']
+        if not User.objects.filter(username=username, is_active=True).exists():
+            raise ValidationError(_("Account doesn't exist."))
+        user = User.objects.get(username=username)
+        if Profile.objects.get(user=user).is_peg():
```
Ninjaclasher, line 243: Stiil don't think this is correct. `is_peg` is a property, and you'll likely get an error similar to
```py
TypeError: 'bool' object is not callable
```

TheAvidDev: Right... back to what it was before.

judge/forms.py
```diff
@@ -230,6 +232,22 @@ def clean(self):
...
 
 
+class WCIPEGMergeRequestForm(Form):
+    handle = CharField(max_length=50, validators=[UnicodeUsernameValidator()])
+
+    def clean_handle(self):
+        username = self.cleaned_data['handle']
+        if not User.objects.filter(username=username, is_active=True).exists():
+            raise ValidationError(_("Account doesn't exist."))
+        user = User.objects.get(username=username)
```
quantum5, line 242: Why not just get the profile and catch the exception instead of doing this exists stuff and then fetching the profile?

TheAvidDev: Done.

judge/forms.py
```diff
@@ -254,6 +256,30 @@ def clean(self):
...
+class WCIPEGMergeRequestForm(Form):
+    handle = CharField(max_length=50, validators=[UnicodeUsernameValidator()])
+
+    if ReCaptchaField is not None:
+        captcha = ReCaptchaField(widget=ReCaptchaWidget())
+
+    def clean_handle(self):
+        try:
+            profile = Profile.objects.get(user__username=self.cleaned_data['handle'], user__is_active=True)
+            if profile.is_peg:
```
Ninjaclasher, line 272: We should probably move this `if` into the `else` branch since it's more "Pythonic".

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+    def merge(self, from_user, to_user):
+        from_user.merge_user = to_user
+        from_user.save()
+        with transaction.atomic():
+            Submission.objects.filter(user=from_user).update(user=to_user)
+            Comment.objects.filter(user=from_user).update(user=to_user)
+            CommentVote.objects.filter(voter=from_user).update(voter=to_user)
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user.is_authenticated:
```
Ninjaclasher, line 354: This feels a bit awkward, since you have to think "wait, what happens when the user is not authenticated?", before realizing that `LoginRequiredMixin` occurs _after_. Maybe we should use `UserPassesTestMixin` with a `test_func`. If we do, remember to place `LoginRequiredMixin` *before* `UserPassesTestMixin`.

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+            token = 'wcipeg_%d_%s' % (int(kwargs['pk']), request.user.email)
+            hmac_token = hmac.new(force_bytes(settings.SECRET_KEY), msg=token.encode('utf-8'), digestmod='sha256')
+            if not hmac.compare_digest(hmac_token.hexdigest(), kwargs['token']):
+                raise PermissionDenied()
+        return super(WCIPEGMergeActivate, self).dispatch(request, *args, **kwargs)
+
+    def form_valid(self, form):
+        self.merge(self.request.user, User.objects.get(pk=self.kwargs['pk']))
```
Ninjaclasher, line 362: Aren't we dealing with a `PegUser`/`Profile` and `Profile` object instead of 2 `User` objects?

e.g.
```py
from_user.merge_user = to_user
```

`from_user` is a `PEGUser` object and `to_user` should be a `Profile`.

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
         return response
 
 
+class WCIPEGMergeActivate(LoginRequiredMixin, TitleMixin, FormView):
+    template_name = 'user/wcipeg-merge-activate.html'
+    form_class = WCIPEGMergeActivationForm
+
+    def merge(self, from_user, to_user):
+        from_user.merge_user = to_user
```
Ninjaclasher, line 346: I think we should move this inside the transaction, since we don't want the `merge_user` to be set if the objects fail to be migrated over.

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+    def get_title(self):
+        return _('WCIPEG Merge Activation')
+
+
+class WCIPEGMergeRequest(LoginRequiredMixin, TitleMixin, FormView):
+    template_name = 'user/wcipeg-merge-request.html'
+    form_class = WCIPEGMergeRequestForm
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user.is_authenticated and not request.profile.is_peg:
```
Ninjaclasher, line 377: Same comment as the previous class.

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+            raise PermissionDenied()
+        return super(WCIPEGMergeRequest, self).dispatch(request, *args, **kwargs)
```
Ninjaclasher, line 379:
```suggestion
        return super().dispatch(request, *args, **kwargs)
```

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+            Comment.objects.filter(user=from_user).update(user=to_user)
+            CommentVote.objects.filter(voter=from_user).update(voter=to_user)
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user.is_authenticated:
+            token = 'wcipeg_%d_%s' % (int(kwargs['pk']), request.user.email)
+            hmac_token = hmac.new(force_bytes(settings.SECRET_KEY), msg=token.encode('utf-8'), digestmod='sha256')
+            if not hmac.compare_digest(hmac_token.hexdigest(), kwargs['token']):
+                raise PermissionDenied()
+        return super(WCIPEGMergeActivate, self).dispatch(request, *args, **kwargs)
```
Ninjaclasher, line 359:
```suggestion
        return super().dispatch(request, *args, **kwargs)
```

judge/views/user.py
```diff
@@ -335,6 +338,59 @@ def get(self, request, *args, **kwargs):
...
+        from_user.merge_user = to_user
+        from_user.save()
+        with transaction.atomic():
+            Submission.objects.filter(user=from_user).update(user=to_user)
+            Comment.objects.filter(user=from_user).update(user=to_user)
+            CommentVote.objects.filter(voter=from_user).update(voter=to_user)
+
+    def dispatch(self, request, *args, **kwargs):
+        if request.user.is_authenticated:
+            token = 'wcipeg_%d_%s' % (int(kwargs['pk']), request.user.email)
```
Ninjaclasher, line 355: Is an `int` cast necessary here?

TheAvidDev: I may be wrong, but I don't think `(?P<pk>\d+)` would automatically convert to `int`.

Ninjaclasher: We don't need it as an `int`, though. It's getting formatted as part of a string anyways. I guess we can keep it if it makes it more clear that `pk` should be an integer.

TheAvidDev: Although I can do:
```suggestion
            token = 'wcipeg_%s_%s' % (kwargs['pk'], request.user.email)
```
if that's what you mean.

Ninjaclasher: Yes

judge/views/user.py
TheAvidDev: Oh yep, you're right -- didn't realize the consequences when I made the `PEGUser` model.

judge/models/profile.py
```diff
@@ -237,6 +241,12 @@ class Meta:
         verbose_name_plural = _('user profiles')
 
 
+class PEGUser(models.Model):
+    user = models.OneToOneField(Profile, verbose_name=_('user'), related_name='peg_user', on_delete=models.CASCADE)
+    merge_user = models.OneToOneField(Profile, verbose_name=_('user'), related_name='peg_merge_user', null=True,
+            on_delete=models.CASCADE)
```
Xyene, line 247: This CASCADE is scary.

judge/forms.py
```diff
@@ -254,6 +256,30 @@ def clean(self):
             raise ValidationError(_('Must specify either totp_token or webauthn_response.'))
 
 
+class WCIPEGMergeActivationForm(Form):
```
Xyene, line 259: I would prefer this entire block of code to be at the end of the file. It'll be maintained as a patch on top of regular site code, and that would reduce the risk of conflicts.

judge/models/profile.py
Xyene: Also, same end-of-file comment.

judge/forms.py
```diff
@@ -275,9 +275,13 @@ def clean_handle(self):
             raise ValidationError(_("Account doesn't exist."))
         return self.cleaned_data['handle']
 
-    def send_email(self, pk, token):
+    def send_email(self, user, token):
         url = reverse('wcipeg_merge_activate', args=[pk, token])
-        # TODO: Send email
+        email = EmailMessage(
+            subject='WCIPEG Account Merge Request',
+            message='To authenticate the merge, login as your DMOJ account, then use the link below:\n\n%s' % url,
```
Xyene: Presumably this works even when they're not yet logged in (it asks them to log in first, and then continues the flow after they do so, right?)

TheAvidDev: I'll test but I believe so. I'll also have it automatically log the user out of their WCIPEG account when a merge is requested and clarify the `generic_message` (once more :facepalm:) -- this will happen immediately when the merge is requested.

TheAvidDev: Now that I've taken a second look, I've realized I just confused you and anyone else reviewing. You have to be logged in as the WCIPEG user for the merge request *and* activation because the email is taken from the `request.user.email`. This message is wrong.

This is also the correct method for verifying WCIPEG with password and verifying DMOJ with email.

judge/forms.py
```diff
@@ -6,16 +6,19 @@
...
 from django.conf import settings
 from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.validators import UnicodeUsernameValidator
 from django.core.exceptions import ValidationError
+from django.core.mail import EmailMessage
 from django.core.validators import RegexValidator
 from django.db.models import Q
 from django.forms import BooleanField, CharField, ChoiceField, Form, ModelForm, MultipleChoiceField
-from django.urls import reverse_lazy
+from django.urls import reverse, reverse_lazy
```
quantum5, line 15: These imports should be placed right before the class if we are to apply this as a patch in prod. Otherwise changing imports will immediately break the patch and cause issues in prod.

judge/forms.py
```diff
@@ -272,3 +275,31 @@ def clean_key(self):
...
+            raise ValidationError(_("Account doesn't exist."))
+        else:
+            if profile.is_peg:
+                raise ValidationError(_('Account must be a DMOJ account.'))
+        return self.cleaned_data['handle']
+
+    def send_email(self, url, email):
+        EmailMessage(
+            subject='WCIPEG Account Merge Request',
+            body='To authenticate the merge, login as your DMOJ account, then use the link below:\n\n%s' % url,
```
quantum5, line 299: Should probably tell the user what to do if they didn't initiate this merge.

judge/models/profile.py
```diff
@@ -276,3 +280,8 @@ class OrganizationRequest(models.Model):
     class Meta:
         verbose_name = _('organization join request')
         verbose_name_plural = _('organization join requests')
+
+
+class PEGUser(models.Model):
```
quantum5, line 285: Model changes should probably be placed into a separate django app to avoid problems with migrations. You should be able to create a blank app called `peg_migration` with just `models.py` containing this model and create the migrations. Then change all the imports. The view and other code can remain in the `judge` app.

judge/migrations/0110_peg_merge.py
```diff
@@ -0,0 +1,22 @@
+# Generated by Django 2.2.12 on 2020-08-01 03:18
```
quantum5, line 1: As suggested elsewhere, this should be moved into a separate `peg_migration` app.

templates/user/edit-profile.html
```diff
@@ -267,6 +267,12 @@
 {% block body %}
     <div class="centered-form">
         <form id="edit-form" action="" method="post" class="form-area">
+            {% if request.profile.is_peg %}
```
quantum5, line 270: Is this discoverable enough? Maybe this should be on the user page for the current user, so that they will see when they log in. Perhaps this should even be a notification like an announcement, or displayed in the navigation bar, or something.

TheAvidDev: I've added the same notification to the rest of the profile pages.

peg_merge/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def test_func(self):
+        return self.request.profile.is_peg
+
+    def get_title(self):
+        return _('WCIPEG Merge Request')
+
+    def form_valid(self, form):
+        user = User.objects.get(username=form.cleaned_data['handle'])
+        token = 'wcipeg_%d_%s' % (user.pk, self.request.user.email)
```
Ninjaclasher, line 30: Is this email correct? In both views, we're using `self.request.user.email`, but the user changes between merge request and merge activate.

peg_merge/forms.py
```diff
@@ -0,0 +1,38 @@
...
+        else:
+            if profile.is_peg:
+                raise ValidationError(_('Account must be a DMOJ account.'))
+        return self.cleaned_data['handle']
+
+    def send_email(self, url, email):
+        print(url)
+        EmailMessage(
+            subject=_('WCIPEG Account Merge Request'),
+            body=_('To authenticate the merge, please use the link below:\n\n%s\n\n') % url +
```
Ninjaclasher, line 31: Was https://github.com/DMOJ/online-judge/pull/1448#discussion_r464015423 lost?

peg_merge/forms.py
```diff
@@ -0,0 +1,38 @@
...
+            profile = Profile.objects.get(user__username=self.cleaned_data['handle'], user__is_active=True)
+        except Profile.DoesNotExist:
+            raise ValidationError(_("Account doesn't exist."))
+        else:
+            if profile.is_peg:
+                raise ValidationError(_('Account must be a DMOJ account.'))
+        return self.cleaned_data['handle']
+
+    def send_email(self, url, email):
+        print(url)
```
Ninjaclasher, line 28: ?

peg_merge/views.py
```diff
@@ -0,0 +1,65 @@
...
+            from_user.peg_user.save()
+            Submission.objects.filter(user=from_user).update(user=to_user)
+            Comment.objects.filter(author=from_user).update(author=to_user)
+            CommentVote.objects.filter(voter=from_user).update(voter=to_user)
+            user = from_user.user
+            user.is_active = False
+            user.save()
+
+    def form_valid(self, form):
+        self.merge(Profile.objects.get(user=self.request.user), Profile.objects.get(user__pk=self.kwargs['pk']))
```
Ninjaclasher, line 59:
```suggestion
        self.merge(self.request.profile, Profile.objects.get(user__pk=self.kwargs['pk']))
```
Do we need to refetch the profile?

peg_merge/views.py
```diff
@@ -0,0 +1,65 @@
...
+
+    def merge(self, from_user, to_user):
+        with transaction.atomic():
+            from_user.peg_user.merge_into = to_user
+            from_user.peg_user.save()
+            Submission.objects.filter(user=from_user).update(user=to_user)
+            Comment.objects.filter(author=from_user).update(author=to_user)
+            CommentVote.objects.filter(voter=from_user).update(voter=to_user)
+            user = from_user.user
+            user.is_active = False
```
Ninjaclasher, line 55: I'm not sure what happens after a merge request is sent in. We're deactivating the user, but the user will remain on the leaderboard (for the moment) as long as `Profile.is_unlisted` is not `True`. Should we be setting it to `True` here as well?

peg_merge/forms.py
TheAvidDev: Ah yep, I'll add that in.

peg_merge/views.py
TheAvidDev: It doesn't. The WCIPEG user only gets logged out once the migration is finished.

peg_merge/forms.py
TheAvidDev: Ah, I keep doing this for testing and forget to remove it for commit.

peg_merge/views.py
TheAvidDev: Didn't know about this, yep.

peg_merge/views.py
Ninjaclasher: Wait, so you are confirming the merge from the PEG account?

TheAvidDev: Yes. The whole idea is that we verify PEG with password and verify DMOJ with email.

templates/user/edit-profile.html
```diff
@@ -267,6 +267,12 @@
...
+                <div class="alert alert-info">
+                    This account was migrated from WCIPEG.
+                    Click <a href="{{ url('wcipeg_merge_request') }}">here</a> to merge with a DMOJ account.
```
Xyene, line 273:
```suggestion
                    Click <b><a href="{{ url('wcipeg_merge_request') }}">here</a></b> to merge with a DMOJ account.
```

templates/user/edit-profile.html
```diff
@@ -267,6 +267,12 @@
...
+                </div>
+            {% endif %}
             {% if form.errors %}
```
Xyene, line 276:
```suggestion
            {% if form.non_field_errors() %}
```

templates/user/wcipeg-merge-activate.html
```diff
@@ -0,0 +1,22 @@
+{% extends "base.html" %}
+
+{% block body %}
+    <div class="centered-form">
+        <form id="wcipeg-merge-activate" action="" method="post" class="form-area">
+            {% csrf_token %}
+            <p>{{ _('To complete the merging of your accounts, click on the button below. Beware that this action is
+                     irreversible and will move all of the submissions onto the new account. You will be logged out of
+                     your WCIPEG account, after which it will become unnaccessible. You can then login to your native
+                     DMOJ account and should see all of your submissions migrated.') }}</p>
```
Xyene, line 10:
```suggestion
                     DMOJ account.') }}</p>
```

templates/user/wcipeg-merge-activate.html
```diff
@@ -0,0 +1,22 @@
+{% extends "base.html" %}
+
+{% block body %}
+    <div class="centered-form">
+        <form id="wcipeg-merge-activate" action="" method="post" class="form-area">
+            {% csrf_token %}
+            <p>{{ _('To complete the merging of your accounts, click on the button below. Beware that this action is
+                     irreversible and will move all of the submissions onto the new account. You will be logged out of
+                     your WCIPEG account, after which it will become unnaccessible. You can then login to your native
```
Xyene, line 9:
```suggestion
                     your WCIPEG account, after which it will become unnaccessible. You will then be able to log in to your
```

templates/user/wcipeg-merge-activate.html
```diff
@@ -0,0 +1,22 @@
...
+                     DMOJ account and should see all of your submissions migrated.') }}</p>
+            <b>{{ form.accept.label }}</b>
+            {{ form.accept }}
+            {% if form.errors %}
+                <div id="form-errors">
+                    {{ form.handle.errors }}
+                </div>
+            {% endif %}
+            <hr>
+            <button class="submit-bar" type="submit">{{ _('Merge Accounts') }}</button>
```
Xyene, line 19:
```suggestion
            <button class="submit-bar" type="submit">{{ _('Merge accounts') }}</button>
```

TheAvidDev: Done (all).

templates/user/edit-profile.html
TheAvidDev: Are you sure this is right? This section isn't edited by me and if this is a patch it may conflict with future changes.

Xyene:
This is patched in prod and will remain so, but let's close it out to decrease open PR clutter.

PR #1452, opened 2020-06-13
Make problems list ordering consistent

kevinjycui:
Add default order `code` as secondary order when sorting by `points`, `ac_rate`, or `user_count` in order to keep problems list ordering consistent.

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -302,11 +302,11 @@ class ProblemList(QueryStringSortMixin, TitleMixin, SolvedProblemMixin, ListView
             queryset = queryset.add_i18n_name(self.request.LANGUAGE_CODE)
             sort_key = self.order.lstrip('-')
             if sort_key in self.sql_sort:
-                queryset = queryset.order_by(self.order)
+                queryset = queryset.order_by(self.order, 'id')
             elif sort_key == 'name':
-                queryset = queryset.order_by(self.order.replace('name', 'i18n_name'))
+                queryset = queryset.order_by(self.order.replace('name', 'i18n_name'), 'id')
             elif sort_key == 'group':
-                queryset = queryset.order_by(self.order + '__name')
+                queryset = queryset.order_by(self.order + '__name', 'id')
             elif sort_key == 'solved':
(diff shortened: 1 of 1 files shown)


judge/views/problem.py
```diff
@@ -302,11 +302,11 @@ def get_paginator(self, queryset, per_page, orphans=0,
...
             if sort_key in self.sql_sort:
-                queryset = queryset.order_by(self.order)
+                queryset = queryset.order_by(self.order, 'code')
             elif sort_key == 'name':
-                queryset = queryset.order_by(self.order.replace('name', 'i18n_name'))
+                queryset = queryset.order_by(self.order.replace('name', 'i18n_name'), 'code')
             elif sort_key == 'group':
-                queryset = queryset.order_by(self.order + '__name')
+                queryset = queryset.order_by(self.order + '__name', 'code')
             elif sort_key == 'solved':
```
Ninjaclasher, line 310: Probably also want to add `code` as a secondary key for the other 2 branches of this `if`.

kevinjycui: Sorting by `solved` or `type` uses the Python built-in `list.sort()` rather than Django's `queryset.order_by()`. Since `list.sort()` uses timsort, I'm pretty sure it already remains consistent as it uses the original indices of the elements in the list as a secondary order. Since the default order is already `code`, I believe this is already implemented.

Ninjaclasher: The default order is not `code`. Assuming you are reading `self.default_sort = 'code'`, this only changes the `self.order`, which is used here to actually order the queryset (before this method, the queryset is unordered). See [source](https://github.com/DMOJ/online-judge/blob/master/judge/utils/views.py#L100). 

Although the default order is technically undefined if not specified, the objects are ordered by `id` [in practice](https://github.com/DMOJ/online-judge/commit/2d4d938f307d80a5b1b9b66ad80835e90cd3e889). 

I think we can assume that the queryset is ordered by `id`, but if so, all the other branches of the `if` should have a secondary order of `id` instead of `code` for consistency.

PR #1455, opened 2020-06-15
Add (slightly better) mobile support for comments

Ninjaclasher:
In particular:
 - Less indenting for small-width screens
 - Decreased gravatar width for small-width screens
 - Use flexbox instead of `float:right;` to make the admin controls display nicer.

Everything should look exactly the same on a computer screen.

**Before:**
image

**After:**
image

### DIFF
--- a/resources/comments.scss
+++ b/resources/comments.scss
@@ -58,7 +58,7 @@ a {
 }
 
 .comment-operation {
-    float: right;
+    flex: auto;
 
     .fa {
         color: #444;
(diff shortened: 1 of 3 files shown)


kiritofeng:
I personally am not a fan of the empty space directly below the icon, what if we moved the icon directly beside the username?

Ninjaclasher:
Something like this?
image

Ninjaclasher:
Implemented ^. I'm not sure if there is a better way to do this without having two gravatars (and choosing which one to make visible). 

**New after:**
image

PR #1457, opened 2020-06-15
Update excluded user ratings on contest rate

### LINKED ISSUE #129: Rating updates
Some detection when saving `Contest`s for changes in every rating related field, and rate the contest and all subsequent contests automatically. For the excluded users, run a query to update their rating in `Profile`s.

Note: I will not support the case of deleting participations. Rerate all will be required in that case.

Implementation note: all these contest rating magic should probably be refactored into `ratings.py` instead of just being everywhere in `admin.py`.

Benefits: None really, except that @Xyene will finally be able to say the admin is more logical instead of more efficient.

Ninjaclasher:
Refs #129.

### DIFF
--- a/judge/ratings.py
+++ b/judge/ratings.py
@@ -107,14 +107,12 @@ def rate_contest(contest):
                  judge_rating ON (judge_rating.user_id = judge_contestparticipation.user_id) INNER JOIN
                  judge_contest ON (judge_contest.id = judge_rating.contest_id)
             WHERE judge_contestparticipation.contest_id = %s AND judge_contest.end_time < %s AND
-                  judge_contestparticipation.user_id NOT IN (
-                      SELECT profile_id FROM judge_contest_rate_exclude WHERE contest_id = %s
-                  ) AND judge_contestparticipation.virtual = 0
+                  judge_contestparticipation.virtual = 0
             GROUP BY judge_rating.user_id
             ORDER BY judge_contestparticipation.score DESC, judge_contestparticipation.cumtime ASC,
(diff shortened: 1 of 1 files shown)


judge/ratings.py
```diff
@@ -127,31 +125,41 @@ def rate_contest(contest):
...
     user_ids = [user[1][1] for user in users]
+    old_rating, old_volatility, times_ranked = get_user_data(data, user_ids)
     ranking = list(map(itemgetter(0), users))
-    old_data = [data.get(user, (1200, 535, 0)) for user in user_ids]
-    old_rating = list(map(itemgetter(0), old_data))
-    old_volatility = list(map(itemgetter(1), old_data))
-    times_ranked = list(map(itemgetter(2), old_data))
     rating, volatility = recalculate_ratings(old_rating, old_volatility, ranking, times_ranked)
 
+    excluded_user_ids = contest.users.filter(virtual=0).exclude(id__in=participation_ids) \
```
Ninjaclasher, line 141: Note: this is pretty inefficient (since `participation_ids` is almost all participants while `excluded_user_ids` is basically no one). It probably won't matter that much for the current DMOJ participant numbers (~200). We can use `judge.utils.raw_sql.join_sql_subquery` here to make it more efficient, but is it worth it to sacrifice the readability?

Ninjaclasher:
There are too many hacks at the moment to be able to auto-rerate a contest when a rating-related field changes (as specified in #129). Thus, this PR has been changed to only updated excluded user ratings on contest rate, which is part of #129.

Ninjaclasher:
Superceded by #1697.

PR #1462, opened 2020-06-19
Change Megabyte/Kilobyte to Mebibyte/Kibibyte in the problem creator

ComputerGenius152:
This addresses [Issue #1200](https://github.com/DMOJ/online-judge/issues/1200).

### DIFF
--- a/templates/user/pp-row.html
+++ b/templates/user/pp-row.html
@@ -7,11 +7,13 @@
     </div>
 </div>
 
-<div class="sub-info">
-    <div class="name">
-        <a href="{{ url('problem_detail', breakdown.problem_code) }}">{{ breakdown.problem_name }}</a>
+<div class="sub-main">
+    <div class="sub-info">
+        <div class="name">
+            <a href="{{ url('problem_detail', breakdown.problem_code) }}">{{ breakdown.problem_name }}</a>
+        </div>
+        <div class="time">{{ relative_time(breakdown.sub_date) }}</div>
     </div>
-    <div class="time">{{ relative_time(breakdown.sub_date) }}</div>
 </div>
 
 <div class="sub-pp sub-usage">


TheAvidDev:
If I'm not mistaken, the issue is about making it use megabytes instead of kilobytes, not just changing the wording. This would mean you could write `64` instead of `65536` for 64mb.

Carson-Tang:
There are a few things missing, which I've added to the description in #1200.

ComputerGenius152:
I've fixed the issues, but I don't think it's a good idea to change the box to mib because people may want to set memory limits under 1024 kib.

judge/models/problem.py
```diff
@@ -121,8 +121,8 @@ class Problem(models.Model):
                                    validators=[MinValueValidator(settings.DMOJ_PROBLEM_MIN_TIME_LIMIT),
                                                MaxValueValidator(settings.DMOJ_PROBLEM_MAX_TIME_LIMIT)])
     memory_limit = models.PositiveIntegerField(verbose_name=_('memory limit'),
-                                               help_text=_('The memory limit for this problem, in kilobytes '
-                                                           '(e.g. 64mb = 65536 kilobytes).'),
+                                               help_text=_('The memory limit for this problem, in kibibytes '
+                                                           '(e.g. 64mib = 65536 kibibytes).'),
```
quantum5, line 125: Might as well take this time to use correct type setting to avoid confusion between MiB and Mib and other craziness.
```suggestion
                                                           '(e.g. 64 MiB = 65536 kibibytes).'),
```

quantum5:
Why does the PR name mention "kebibyte" which is not a real word?

ComputerGenius152:
That was a mistake on my part, but it's fixed now.

PR #1464, opened 2020-06-20
Allow org admins to see all org problems

kiritofeng:
Currently a hotfix for ongoing WCIPEG migration.

### DIFF
--- a/judge/models/problem.py
+++ b/judge/models/problem.py
@@ -178,7 +178,9 @@ class Problem(models.Model):
             return False
         if user.has_perm('judge.edit_all_problem') or user.has_perm('judge.edit_public_problem') and self.is_public:
             return True
-        return user.has_perm('judge.edit_own_problem') and user.profile.id in self.editor_ids
+        return user.has_perm('judge.edit_own_problem') and \
+            (user.profile.id in self.editor_ids or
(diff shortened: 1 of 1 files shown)


judge/models/problem.py
```diff
@@ -271,6 +273,7 @@ def get_editable_problems(cls, user):
             return cls.objects.all()
 
         q = Q(authors=user.profile) | Q(curators=user.profile)
+        q |= Q(is_organization_private=True, organizations__in=user.profile.organizations.filter(admins=user.profile))
```
quantum5, line 276: This feels overly convoluted. Are you sure `user.profile.admin_of.all()` doesn't work?

kiritofeng: Ah I forgot that was a thing :blobderpy:

judge/models/problem.py
```diff
@@ -271,6 +273,7 @@ def get_editable_problems(cls, user):
             return cls.objects.all()
 
         q = Q(authors=user.profile) | Q(curators=user.profile)
+        q |= Q(is_organization_private=True, organizations__in=user.profile.admin_of.all())
```
Ninjaclasher, line 276: Since `get_visible_problems` does not use `get_editable_problems`, a similar line to this should be added. Otherwise, `is_accessible_by` and `get_visible_problems` will not be consistent.

PR #1466, opened 2020-06-21
Fix CSS in case where only extended feedback is displayed

kiritofeng:
## New behaviour

With both output and feedback:
image
With only feedback;
image
With only output:
image


kiritofeng:
With both output and feedback:
image
With only feedback;
image
With only output:
image

PR #1469, opened 2020-06-22
Make comment votes icon use pointer cursor

TheAvidDev:
Firefox doesn't switch to the pointer cursor for an `<a>` tag without an `href` attribute. Adding one here fixes that.

In this case, we want it to look like it's clickable so doing so is appropriate. If this is *always* the case  with an `<a>` tag having a pointer, it would be better to just add css along the lines of:
```css
a {
    cursor: pointer;
}
```
---
This affects the following admin comment icon:
image

### DIFF
--- a/templates/comments/list.html
+++ b/templates/comments/list.html
@@ -93,7 +93,7 @@
                                             <a data-featherlight="{{ url('comment_votes_ajax', node.id) }}"
                                                href="javascript:void(0)"
                                                title="{{ _('Votes') }}" class="votes-link"><i class="fa fa-bar-chart fa-fw"></i></a>
-                                            <a href="javascript:" title="{{ _('Hide') }}" data-id="{{ node.id }}"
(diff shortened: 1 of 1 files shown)


templates/comments/list.html
```diff
@@ -91,6 +91,7 @@ <h2><i style="padding-right: 0.3em" class="fa fa-comments"></i>{{ _('Comments')
                                                    class="edit-link"><i class="fa fa-pencil fa-fw"></i></a>
                                             {% endif %}
                                             <a data-featherlight="{{ url('comment_votes_ajax', node.id) }}"
+                                               href="javascript:void(0)"
```
Ninjaclasher, line 94: Two lines below, `javascript:`. We should probably choose one of them instead of having both for consistency.

TheAvidDev: Ah, you're right. There's a variety of options: `javascript:`, `javascript:void`, `javascript:void(0)` and all of them can also have brackets on the end.

I don't know what's better.

Issue #804, opened 2018-11-21
Contest filtering


Xyene:

Xyene:
Sorting was implemented in https://github.com/DMOJ/online-judge/pull/1481, but searching or filtering by contest label is not implemented yet.

PR #1474, opened 2020-07-02
Add sorting to contest page

ComputerGenius152:
This would allow the contest page to be sorted by the name or the number of users.

Relates to issue #804.

### DIFF
--- a/templates/contest/list.html
+++ b/templates/contest/list.html
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.organizers.all() or contest.is_in_contest(request.user) %}
+            {% if request.profile in contest.organizers.all() %}
                 <form action="{{ url('contest_join', contest.key) }}" method="post">
                     {% csrf_token %}
                     <input type="submit" class="unselectable button full participate-button"


templates/contest/list.html
```diff
@@ -255,7 +259,6 @@ <h4>{{ _('Upcoming Contests') }}</h4>
             <br>
         {% endif %}
         <br>
-
```
TheAvidDev, line 258: This shouldn't be removed, it's there to separate the current, upcoming, and past contests.

judge/views/contests.py
```diff
@@ -104,6 +107,9 @@ def get_context_data(self, **kwargs):
         context['future_contests'] = future
         context['now'] = self._now
         context['first_page_href'] = '.'
+        context.update(self.get_sort_context())
+        context.update(self.get_sort_paginate_context())
+
```
TheAvidDev, line 112: This (newline) occasionally happens in the code base, but we're somewhat inconsistent and it shouldn't be added if it wasn't there before.

templates/contest/list.html
```diff
@@ -148,10 +148,14 @@ <h4>{{ _('Active Contests') }}</h4>
...
-                    <th>{{ _('Users') }}</th>
+                    <th class="contest" style="width:90%">
+                        <a href="{{ sort_links.name }}">{{ _('Contest') }}{{ sort_order.name }}</a>
+                    </th>
+                    <th class="users">
+                        <a href="{{ sort_links.user_count }}">{{ _('Users') }}{{ sort_order.user_count }}</a>
+                    </th>
                     {% if not request.in_contest %}
-                        <th style="width:15%"></th>
+                        <th class="join" style="width:15%"></th>
```
TheAvidDev, line 158: Perhaps I'm just not finding it, but I don't think the `join` class is defined anywhere and it's not used anywhere else either where there's a `<th style="width:15%"></th>` such as [here](https://github.com/DMOJ/online-judge/blob/9858c8bf703d9950740c84bdfc3ad6d6da1c4683/templates/contest/list.html#L196) or [here](https://github.com/DMOJ/online-judge/blob/9858c8bf703d9950740c84bdfc3ad6d6da1c4683/templates/contest/list.html#L272).

ComputerGenius152:
Thanks for pointing that out, I've made the needed changes.

templates/contest/list.html
```diff
@@ -255,7 +259,7 @@ <h4>{{ _('Upcoming Contests') }}</h4>
             <br>
         {% endif %}
         <br>
-
+
```
TheAvidDev, line 262:
```suggestion

```
Extra spaces.

ComputerGenius152: It just works :)

ComputerGenius152:
For the various classes I added, I had originally thought to use another method but forgot to remove them. Ongoing, upcoming, and past contests should all be included now.

judge/views/contests.py
```diff
@@ -62,20 +63,23 @@ def get_queryset(self):
...
+    default_sort = '-start_time'
 
     @cached_property
     def _now(self):
         return timezone.now()
 
     def _get_queryset(self):
         return super(ContestList, self).get_queryset() \
-            .order_by('-start_time', 'key').prefetch_related('tags', 'organizations', 'organizers')
+            .order_by(self.order).prefetch_related('tags', 'organizations', 'organizers')
```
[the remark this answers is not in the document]
Ninjaclasher: Currently, active and future contests are not sorted here: see [here](https://github.com/DMOJ/online-judge/blob/master/judge/views/contests.py#L100). Thus, only past and current contests are affected. This can be confusing, since the order the tables are displayed are active, current, future, past (alternating between affected by sorting and not affected).

I think we should only make past contests affected by this sorting, rather than past and current contests.

templates/contest/list.html
```diff
@@ -228,7 +236,9 @@ <h4>{{ _('Upcoming Contests') }}</h4>
             <table class="contest-list table striped">
                 <thead>
                 <tr>
-                    <th>{{ _('Contest') }}</th>
+                    <th style="width:90%">
```
Ninjaclasher, line 239: Is this supposed to be `width:90%`?

ComputerGenius152: I originally did that to maintain consistency with the other contest groups (past, active, and current), but you are correct that it shouldn't be `width:90%` because upcoming contests don't have user counts or join buttons yet.

judge/views/contests.py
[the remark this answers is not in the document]
ComputerGenius152: I have added the secondary order back. Do you think it would be a good idea to instead make future contests also sortable? The order would then be active (not sorted), current (sorted), future (sorted), and past (sorted).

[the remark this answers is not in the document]
Ninjaclasher: I think sorting future contests would be confusing, since by default, it is sorted by `start_time` (as opposed to `-start_time`). This means that when you click on that column to sort, the sort order would not change for the future contests table (until you click again). Another issue is we would now have three states: sorted by `start_time`, sorted by `-start_time`, and default order.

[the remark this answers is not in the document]
ComputerGenius152: There is only `start_time` and the default order now, similar to `user.py` and `problem.py`. Also, current contests are no longer affected by the sorting.

judge/views/contests.py
```diff
@@ -62,20 +63,23 @@ def get_queryset(self):
...
+    default_sort = 'start_time'
 
     @cached_property
     def _now(self):
         return timezone.now()
 
     def _get_queryset(self):
         return super(ContestList, self).get_queryset() \
-            .order_by('-start_time', 'key').prefetch_related('tags', 'organizations', 'organizers')
+            .order_by(self.order, 'key').prefetch_related('tags', 'organizations', 'organizers')
```
Ninjaclasher, line 82: Since this `order_by` only affects past contests (and all other contests are sorted in Python), this `order_by` should probably be moved to `get_queryset` to save an uneeded `ORDER BY` in `get_context_data`.

judge/views/contests.py
```diff
@@ -98,12 +102,15 @@ def get_context_data(self, **kwargs):
                     present.remove(participation.contest)
 
         active.sort(key=attrgetter('end_time'))
+        present.sort(key=attrgetter('end_time'))
```
Ninjaclasher, line 105: Since we are touching this code anyways, let's add a secondary order to these sorts (since `start_time` and `end_time` are commonly duplicated for e.g junior/senior contests).

templates/contest/list.html
```diff
@@ -148,8 +148,12 @@ <h4>{{ _('Active Contests') }}</h4>
             <table class="contest-list table striped">
                 <thead>
                 <tr>
-                    <th style="width:90%">{{ _('Contest') }}</th>
-                    <th>{{ _('Users') }}</th>
+                    <th style="width:90%">
```
Ninjaclasher, line 151: I am not sure if the correct behaviour is to make all the table headers clickable, since only past contests are affected.

judge/views/contests.py
```diff
@@ -62,20 +63,23 @@ def get_queryset(self):
...
-class ContestList(DiggPaginatorMixin, TitleMixin, ContestListMixin, ListView):
+class ContestList(QueryStringSortMixin, DiggPaginatorMixin, TitleMixin, ContestListMixin, ListView):
     model = Contest
     paginate_by = 20
     template_name = 'contest/list.html'
     title = gettext_lazy('Contests')
     context_object_name = 'past_contests'
+    all_sorts = frozenset(('name', 'user_count', 'start_time'))
+    default_desc = frozenset(('name', 'user_count'))
+    default_sort = 'start_time'
```
Ninjaclasher, line 74:
```suggestion
    default_sort = '-start_time'
```

judge/views/contests.py
```diff
@@ -97,13 +100,16 @@ def get_context_data(self, **kwargs):
                     active.append(participation)
                     present.remove(participation.contest)
 
-        active.sort(key=attrgetter('end_time'))
+        active.sort(key=attrgetter('end_time', 'start_time'))
```
Ninjaclasher, line 103: Sorry, I wasn't clear with what I meant. I meant something similar with https://github.com/DMOJ/online-judge/pull/1474#discussion_r449338840 and having `key` as the secondary order.

ComputerGenius152: Like this?

Ninjaclasher: Yup!

ComputerGenius152:
The unresolved conflict was due to [this commit](https://github.com/DMOJ/online-judge/commit/5ba45c46335d50b27cdf5162293d8a67a8bfbfd1), and has since been resolved.

Issue #236, opened 2015-09-16
More user statistics


Xyene:
Apart from "Last seen time", we can also consider a Statistics tab on user pages. There, we could include graphs like:

* Used languages pie chart
* Status rates (? this one is iffy since it may encourage spamming helloworld to get a high AC score)
* History graph, Github-style

FatalEagle:
Don't add "Last seen time", it's too stalker-ish. We need to respect user's privacy.

Xyene:
I think I covered the non-mission-critical aspect...

> None of these are crucial features by any means (aside from perhaps the "Last seen time"), but it would give users more data to play with. More data to play with must be a good thing.

Still, I believe it's a leap to think that just by having a feature people will abuse it. Hackerrank and SPOJ both have graphs of various natures, and they don't appear to be abused by the large majority of users. Something important to note is that *submission spamming is not a bad thing for us* -- we don't incur a penalty of any sort for them. We're paying for judging servers anyways, so there's no clear detriment to them being put to work. Who cares if someone wants to up their AC rate? It's not something as conclusive as a contest rating chart, but does provide info that may be useful to improving their skills (e.g. "Oh, it looks like a lot of my submissions MLE. Might be a good idea to learn some ways to optimize my memory usage"). And on the other hand, if anything, a history chart may actually increase participation (users will want to keep their submission streak, which would lead to more activity on the site).

Depriving legitimate users from such graphs because they may be abused does not seem like the sensible thing to do, especially when they have the potential to boost user involvement.

quantum5:
Graphs seems interesting though I would hope they are not on default tab on the users page. Having a good idea doesn't excuse killing our servers.

PR #1480, opened 2020-07-08
Add submission activity graph to users page; #236

Ninjaclasher:
Refs #236.

As stated in #236, these charts should be moved to a separate tab to reduce load on the server once there are more charts. However, this shouldn't be a problem for now since there are only two charts (with not that many queries).

@Carson-Tang has requested this feature multiple times. :smile: 

**Screenshots:**
image

image

image

### DIFF
--- a/templates/user/user-about.html
+++ b/templates/user/user-about.html
@@ -45,6 +45,69 @@
             <br>
         {% endif %}
 
+        <h4 id="submission-activity-header"></h4>
+        <div id="submission-activity" style="display: none;">
+            <div id="submission-activity-actions">
+                <a href="javascript:void(0)" id="prev-year-action">&laquo;</a>
+                &nbsp;<span id="year"></span>&nbsp;
+                <a href="javascript:void(0)" id="next-year-action">&raquo;</a>
+            </div>
+            <div id="submission-activity-display">
+                <table id="submission-activity-table">
+                    <tr id="submission-0">
+                        <th class="submission-date-col info-text">
+                            {{ _('Mon') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-1">
+                        <th class="submission-date-col info-text">
+                            {{ _('Tues') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-2">
+                        <th class="submission-date-col info-text">
+                            {{ _('Wed') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-3">
+                        <th class="submission-date-col info-text">
+                            {{ _('Thurs') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-4">
+                        <th class="submission-date-col info-text">
+                            {{ _('Fri') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-5">
+                        <th class="submission-date-col info-text">
+                            {{ _('Sat') }}
+                        </th>
+                    </tr>
+                    <tr id="submission-6">
+                        <th class="submission-date-col info-text">
+                            {{ _('Sun') }}
+                        </th>
+                    </tr>
+                </table>
+                <div class="info-bar">
+                    <span id="submission-total-count" class="info-text">
+                    </span>
+                    <table class="info-table">
+                        <tr>
+                            <th class="info-table-text info-text">{{ _('Less') }}</th>
+                            <td class="activity-0"><div></div></td>
+                            <td class="activity-1"><div></div></td>
+                            <td class="activity-2"><div></div></td>
+                            <td class="activity-3"><div></div></td>
+                            <td class="activity-4"><div></div></td>
+                            <th class="info-table-text info-text">{{ _('More') }}</th>
+                        <tr>
+                    </table>
+                </div>
+            </div>
+        </div>
+
         {% if rating %}
             <h4>Rating history</h4>
             <div id="rating-chart">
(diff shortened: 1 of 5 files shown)


templates/user/user-about.html
```diff
@@ -67,10 +130,122 @@ <h4>Rating History</h4>
         {% include "mathjax-load.html" %}
     {% endif %}
 
+    <script type="text/javascript">
+        var submission_activity = {{ submission_data }};
+        var metadata = {{ submission_metadata }};
+        const activity_levels = 5; // 5 levels of activity
+
+        $(document).ready(function () {
```
Xyene, line 138: We usually just `$(...)`

templates/user/user-about.html
```diff
@@ -67,10 +130,122 @@ <h4>Rating History</h4>
...
+                    var level = activity_breakdown.findIndex(x => x >= obj.activity);
+                    // There doesn't seem to be a way to internationalize strings in JavaScript.
+                    var text = "{{ _('%%(cnt)d submissions on %%b %%d, %%Y') }}"
+                        .replace('%(cnt)d', obj.activity)
+                        .replace('%b', metadata.month_names[obj.date.getMonth()])
+                        .replace('%d', obj.date.getDate())
+                        .replace('%Y', obj.date.getFullYear());
+                    $div.find('#submission-' + obj.weekday)
+                        .append(
+                            $('<td></td>').addClass(['activity-label', 'activity-' + level])
```
Xyene, line 223: Pretty sure this works (here and elsewhere)

```suggestion
                            $('td').addClass(['activity-label', 'activity-' + level])
```

templates/user/user-about.html
```diff
@@ -67,10 +130,122 @@ <h4>Rating History</h4>
...
+                    days.push({
+                        date: new Date(day),
+                        weekday: day.getDay(),
+                        day_num: day_num,
+                        activity: submission_activity[isodate] || 0,
+                    });
+                }
+
+                var sum_activity = days.map(obj => obj.activity).reduce((a, b) => a + b, 0);
+                // There doesn't seem to be a way to internationalize strings in JavaScript.
```
Xyene, line 199: /cc @quantum5

TheAvidDev: Doesn't the code right under this comment already internationalize?

Xyene: I'm not sure if it's enough (ref. plurals).

quantum5: Fortunately, submissions are not fireplace pokers. Unfortunately, i18n should be done in JavaScript too, using the `gettext` function, as there are still pluralization issues.

templates/user/user-about.html
Ninjaclasher: It seems this selects the `td` objects already on the page and adds the class, while we want to create a new `td` object.

quantum5: The correct code is `$('<td>')`.

templates/user/user-about.html
Ninjaclasher: Done

templates/user/user-about.html
Ninjaclasher: Done

templates/user/user-about.html
Ninjaclasher: Done

templates/user/user-about.html
```diff
@@ -67,10 +130,122 @@ <h4>Rating History</h4>
...
+
+                for (; current_weekday < days[0].weekday; current_weekday++) {
+                    $div.find('#submission-' + current_weekday)
+                        .append($('<td>').addClass('activity-blank').append('<div>'));
+                }
+
+                days.forEach(obj => {
+                    var level = activity_breakdown.findIndex(x => x >= obj.activity);
+                    var text =
+                        ngettext("%(cnt)d submission on %b %d, %Y", "%(cnt)d submissions on %b %d, %Y", obj.activity)
```
quantum5, line 216: This is terrible i18n for the date. While you can rearrange the symbols, not every language uses full month names. For example, it would be ridiculous to show `2020年八月10日` instead of something more sane like `2020年8月10日`, which is the best you can do with this code.

quantum5: Oh and I forgot, in some languages, the month name needs to be in the genitive case for dates. Your nominative month names won't cut it.

Ninjaclasher: From some discussion on slack, there are a few options (since there doesn't seem to be anything in the Django docs for formatting dates in JavaScript):
 1. Format the activity date thing on the server.
      - I'm not sure how this will work, since wouldn't we need to format _all_ date strings (which could grow very large if the user has been active for multiple years)?
 2. Use the user's browser, and using something like `date.toLocaleDateString()` (or maybe moment.js's `format`?).
    - `date.toLocaleDateString()` could work, but it returns something like `10/08/2020`, which is tolerable, but not ideal for users.
    - moment.js's `format` - We'll need a way to convert from Django's date formatting to moment.js's format if we want to keep the date formatting consistent (using `get_format("DATE_FORMAT")`).

quantum5: Option 3: just use ISO date formats, which are internationally recognized.

Ninjaclasher: I found another option:

Would something like `new Date().toLocaleDateString("{{ LANGUAGE_CODE }}", {month: 'short', year: 'numeric', day: 'numeric'})` work? This will format dates to `2020年8月11日`, `11 Aug 2020`, and ` авг. 2020 г.`.

templates/user/user-about.html
```diff
@@ -45,6 +45,69 @@ <h4>{{ _('About') }}</h4>
             <br>
         {% endif %}
 
+        <h4>{{ _('Submission Activity') }}</h4>
```
Xyene, line 48: %d submissions in the last year?

Might as well switch the other headers on the user page to use sentence case, since most of the site uses sentence case.

templates/user/user-about.html
quantum5: Looks about right. You can use that. The string you quoted for Russian is wrong, but it does not appear to be a bug in the code.

Ninjaclasher: Done

templates/user/user-about.html
Ninjaclasher: Done

Xyene (approved):
Thanks!

Issue #1487, opened 2020-07-25
Extend new user comment mute to tickets and profile editing


kiritofeng:
Users should be required to solve at least one problem before they can create tickets outside of live contests and edit their profile to discourage spam.

Harry7557558:
Agreed. And users should be required to solve some problems before they can vote comments. It may be a good idea to allow people to see `[upvote count]/[downvote count]` when the mouse hover comment vote count area (HTML title) and display a comment tab on users' public profiles that list of all of the user's comments. Users with low average comment quality should be warned by email and be automatically muted for continued low-quality posts. Also, users should not be able to vote themselves' comments. To prevent spam voting comments, there should be an option to turn off a user's right to vote.

TheAvidDev:
> And users should be required to solve some problems before they can vote comments.

This should be a separate issue.

> It may be a good idea to allow people to see [upvote count]/[downvote count] when the mouse hover comment vote count area (HTML title).

Ditto. I don't think this is really necessary, but it probably wouldn't be that hard to add.

> Users with low average comment quality should be warned by email and be automatically muted for continued low-quality posts.

Disagree. Emailing users is simply unnecessary for comments that others dislike. Being muted for having comments that people dislike also seems quite harsh and an easy system to abuse. Muting should be done manually, as it is.

> To prevent spam voting comments, there should be an option to turn off a user's right to vote.

If this is to be done, it should probably just be a part of the `profile.mute` field. Not really sure if this is a good or bad idea though.

> Users should be able to see the time a comment is edited.

Separate issue. This I can get behind.

---

You should probably create a separate issue for all of these potential enhancements and use the bulleted list markdown.

Ninjaclasher:
To add on,

> And users should be required to solve some problems before they can vote comments.

This is [already implemented](https://github.com/DMOJ/online-judge/blob/master/judge/views/comment.py#L35).

> display a comment tab on users' public profiles that list of all of the user's comments.

#73

kiritofeng:
Implemented by #1633 and #1639.

PR #1490, opened 2020-07-30
Add problem solve requirement for user ticket/profile-edit/org-join; #1487

Carson-Tang:


TheAvidDev:
Just a note, before you go too far, this looks like a lot of repeated logic. It's probably better to make a util function.

judge/views/organization.py
```diff
@@ -114,6 +114,10 @@ def handle(self, request, org, profile):
 
 class JoinOrganization(OrganizationMembershipChange):
     def handle(self, request, org, profile):
+        if not profile.submission_set.filter(points=F('problem__points')).exists():
+            return generic_message(request, _('Joining organization'),
```
Xyene, line 118: Hmm, but users can select an organization at signup, before they have any points?

PR #1492, opened 2020-08-02
Add page to view all user submissions in given contest

kiritofeng:
P1 is publicly visible, and P2 is not publicly visible.

As regular user, with public scoreboard:
image

image

(Link for P2 will 404)

Combined submissions:
image

As SU/organizer:
image

Tested:
- SU while contest is running
- Contest organizer while contest was running, and once done, with private scoreboard.
- Contestant while contest was running, and once done, with private scoreboard.
- Random 3rd user, while contest was running, and once done, with and without private scoreboard.
- Not logged in, while contest was running, and once done, with and without private scoreboard.

### DIFF
--- a/judge/views/submission.py
+++ b/judge/views/submission.py
@@ -183,6 +183,16 @@ def abort_submission(request, submission):
     return HttpResponseRedirect(reverse('submission_status', args=(submission.id,)))
 
 
+def filter_submissions_by_visible_problems(queryset, user):
+    join_sql_subquery(
+        queryset,
+        subquery=str(Problem.get_visible_problems(user).distinct().only('id').query),
+        params=[],
+        join_fields=[('problem_id', 'id')],
+        alias='visible_problems',
+    )
+
+
 class SubmissionsListBase(DiggPaginatorMixin, TitleMixin, ListView):
     model = Submission
     paginate_by = 50
@@ -248,13 +258,7 @@ class SubmissionsListBase(DiggPaginatorMixin, TitleMixin, ListView):
     def get_queryset(self):
         queryset = self._get_queryset()
         if not self.in_contest:
-            join_sql_subquery(
-                queryset,
-                subquery=str(Problem.get_visible_problems(self.request.user).distinct().only('id').query),
-                params=[],
-                join_fields=[('problem_id', 'id')],
-                alias='visible_problems',
-            )
+            filter_submissions_by_visible_problems(queryset, self.request.user)
 
         return queryset
 
@@ -319,9 +323,13 @@ class UserMixin(object):
 
 
 class ConditionalUserTabMixin(object):
+    @cached_property
+    def is_own(self):
+        return self.request.user.is_authenticated and self.request.profile == self.profile
+
     def get_context_data(self, **kwargs):
         context = super(ConditionalUserTabMixin, self).get_context_data(**kwargs)
-        if self.request.user.is_authenticated and self.request.profile == self.profile:
+        if self.is_own:
             context['tab'] = 'my_submissions_tab'
         else:
             context['tab'] = 'user_submissions_tab'
@@ -334,12 +342,12 @@ class AllUserSubmissions(ConditionalUserTabMixin, UserMixin, SubmissionsListBase
         return super(AllUserSubmissions, self).get_queryset().filter(user_id=self.profile.id)
 
     def get_title(self):
-        if self.request.user.is_authenticated and self.request.profile == self.profile:
+        if self.is_own:
             return _('All my submissions')
         return _('All submissions by %s') % self.username
 
     def get_content_title(self):
-        if self.request.user.is_authenticated and self.request.profile == self.profile:
+        if self.is_own:
             return format_html('All my submissions')
         return format_html('All submissions by <a href="{1}">{0}</a>', self.username,
                            reverse('user_page', args=[self.username]))
@@ -413,10 +421,6 @@ class ProblemSubmissions(ProblemSubmissionsBase):
 
 class UserProblemSubmissions(ConditionalUserTabMixin, UserMixin, ProblemSubmissions):
     check_contest_in_access_check = False
-
-    @cached_property
-    def is_own(self):
-        return self.request.user.is_authenticated and self.request.profile == self.profile
 
     def access_check(self, request):
         super(UserProblemSubmissions, self).access_check(request)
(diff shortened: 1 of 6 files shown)


judge/views/submission.py
```diff
@@ -531,6 +531,21 @@ def get(self, request, *args, **kwargs):
         return super(ForceContestMixin, self).get(request, *args, **kwargs)
 
 
+class UserAllContestSubmissions(ForceContestMixin, AllUserSubmissions):
+    def get_title(self):
+        return "%s's submissions in %s" % (self.username, self.contest.name)
+
+    def access_check(self, request):
+        super(UserAllContestSubmissions, self).access_check(request)
+        if not self.contest.users.filter(user_id=self.profile.id).exists():
```
TheAvidDev, line 540: Not entirely sure of this but,
```suggestion
        if not self.contest.users.filter(user=self.profile).exists():
```

judge/views/submission.py
```diff
@@ -531,6 +531,21 @@ def get(self, request, *args, **kwargs):
...
+            raise Http404()
+
+    def get_content_title(self):
+        return format_html(_('<a href="{1}">{0}</a>\'s submissions in <a href="{3}">{2}</a>'),
```
TheAvidDev, line 544: I think @quantum5 has some gripes about HTML in translation strings, not sure if that applies here.

judge/views/submission.py
```diff
@@ -531,6 +531,21 @@ def get(self, request, *args, **kwargs):
         return super(ForceContestMixin, self).get(request, *args, **kwargs)
 
 
+class UserAllContestSubmissions(ForceContestMixin, AllUserSubmissions):
+    def get_title(self):
+        return "%s's submissions in %s" % (self.username, self.contest.name)
```
TheAvidDev, line 536: Missing i18n.

judge/views/submission.py
quantum5: There are examples such as https://github.com/DMOJ/online-judge/blob/master/judge/views/ticket.py#L87-L90, but maybe we should do this in a separate PR and get it all out of the way.

judge/views/submission.py
```diff
@@ -531,6 +531,21 @@ def get(self, request, *args, **kwargs):
...
+
+    def access_check(self, request):
```
Ninjaclasher, line 538: I think we should only open up this view for contest organizers, since otherwise this bypasses the checks that were put in place to stop users from seeing contest problem submissions outside of the contest.

judge/views/submission.py
```diff
@@ -531,6 +531,21 @@ def get(self, request, *args, **kwargs):
         return super(ForceContestMixin, self).get(request, *args, **kwargs)
 
 
+class UserAllContestSubmissions(ForceContestMixin, AllUserSubmissions):
```
Ninjaclasher, line 534: Hmm... I think we're missing the actual submission filtering.

judge/views/submission.py
kiritofeng: This uses the same access checks as `UserContestSubmission`

Ninjaclasher: `UserContestSubmissions` has `UserProblemSubmissions` in its inheritance tree, so it also checks for problem accessibility (making sure the user can access the problem before being able to view the related contest submissions). On the other hand, there is only `ForceContestMixin` and `AllUserSubmissions`, which means the only contest-specific is checking that the contest is running from `ForceContestMixin`.

judge/views/submission.py
kiritofeng: Which submission filtering are you referring to? `ForceContestMixin` causes `in_contest` to be True, so when `get_queryset()` is called in `AllUserSubmissions`, it will call `SubmissionListBase`, which with `in_contest` as True will filter for only contest submissions, and then `AllUserSubmissions` will filter by the user.

Ninjaclasher: Ah, you're correct. Filtering is already done properly in `SubmissionListbase`.

judge/views/submission.py
Ninjaclasher: Looks like this was copied from `UserContestSubmissions`. The queries generated are equivalent (and they are equally clear).

judge/views/submission.py
kiritofeng: Hm so in this case, access should be granted if:
1. You can edit the contest in some form
2. Scoreboard is publicly accessible
3. It's your participation
?

Ninjaclasher: Currently, you can only access in-contest submissions (for a certain problem) if:
 1. You can edit the contest.
 2. You can access the problem _outside_ of the contest.
 3. You are in the contest.

I'm not sure whether this is the correct behaviour or not.

judge/views/submission.py
```diff
@@ -183,6 +183,18 @@ def abort_submission(request, submission):
...
+def filter_submissions_by_visible_problems(queryset, user):
+    join_sql_subquery(
+        queryset,
+        subquery=str(Problem.get_visible_problems(user).distinct().only('id').query),
+        params=[],
+        join_fields=[('problem_id', 'id')],
+        alias='visible_problems',
+    )
+
+    return queryset
```
Ninjaclasher, line 195: It doesn't seem we ever use this `return`.

judge/views/submission.py
```diff
@@ -330,16 +336,20 @@ def get_context_data(self, **kwargs):
 
 
 class AllUserSubmissions(ConditionalUserTabMixin, UserMixin, SubmissionsListBase):
+    @cached_property
+    def is_own(self):
```
Ninjaclasher, line 340: Since this code is duplicated in `UserProblemSubmissions`, maybe we should bring this into `ConditionalUserTabMixin` (there is another instance of `self.request.user.is_authenticated and self.request.profile == self.profile` in that class).

judge/views/submission.py
```diff
@@ -531,6 +541,37 @@ def get(self, request, *args, **kwargs):
...
+            'user': self.username,
+            'contest': self.contest.name,
+        }
+
+    def access_check(self, request):
+        super().access_check(request)
+        if not self.contest.users.filter(user_id=self.profile.id).exists():
+            raise Http404()
+        if not self.contest.can_see_full_scoreboard(self.request.user):
+            if not self.is_own:
```
Ninjaclasher, line 558: Though this doesn't matter too much (since it will only matter for when you view your own submissions), maybe we should switch the order of these `if` statements, as computing `self.is_own` is cheap compared to `can_see_full_scoreboard` (not to mention, `is_own` is probably cached at this point).

Issue #79, opened 2015-03-27
Select judge to submit on


Xyene:
Useful to prevent spam submissions to get the fastest judge. Doesn't impact our performance much, either.

quantum5:
Duplicate of #77.

Xyene:
Sigh, misleading issue names. My bad. TooManyIssuesException.

Xyene:
Wait, on further discussion with @FatalEagle this is *not* a duplicate and the title of #77 is what I thought it was, so not misleading. Reopened.

quantum5:
We shouldn't encourage users to religiously select judges as it defeats the purpose of multiple judges.

Xyene:
If it's a 3-level deep nested dropdown, it'd be pretty inconvenient to select at all times. Rather, I see people using regular submit until they have an AC solution, then submitting under the fastest judge for ranking. Which in my opinion is more fair since our judges have unequal capacity.

FatalEagle:
Maybe require something even more annoying like a codecha to submit on a particular judge.

Xyene:
Yea, something annoying like that would prevent abuse.

quantum5:
And of course admin exemption from codechas.

I REALLY REALLY HATE IT.

FatalEagle:
Another idea: either submit codecha, or do a survey (and make $$$ for us) to submit. http://screencloud.net/v/5rmi

quantum5:
Alright as long as we make it hard but not too bad to warrant registering a new account.

Xyene:
This'd be really nice for admins, and debugging.

Xyene:
![image](https://cloud.githubusercontent.com/assets/1403503/25319727/c567dafa-286e-11e7-9ab1-834b59841d26.png)

PR #1496, opened 2020-08-02
Use a tuple of judges when selecting judge to submit on; #79

Ninjaclasher:


judge/bridge/django_handler.py
```diff
@@ -37,11 +37,11 @@ def on_submission(self, data):
         problem = data['problem-id']
         language = data['language']
         source = data['source']
-        judge_id = data['judge-id']
+        judge_ids = data['judge-ids']
```
Xyene, line 40: Thanks for the patch, I'll apply it to prod as-is.

I do have some concerns about naming here though, that I realized I had *after* merging the original patch, and now's a good time to address them.

The main one is that this isn't the judge *id*. It's the judge *name*, i.e. correct logic using this is `judge.name in judge_ids`, and *not* `judge.id in judge_ids` (the latter of which sounds more correct to a reader, but is actually wrong despite not raising any errors).

Ninjaclasher: I've renamed `judge_ids` to `judge_names` for specifying a judge to submit on. However, there are a lot of other places in the bridge which also uses `judge_id` when it should also be `judge_name`. Should those be renamed as well?

Ninjaclasher:
Was used for the PEG migration and is no longer relevant.

Issue #1440, opened 2020-06-09
org admins should have perms to see all problems private to the org


xiaowuc1:
motivated by recent current events

PR #1504, opened 2020-08-11
Add unittests for organization-private problems

Ninjaclasher:
Refs #1440 and #1464.

### DIFF
--- a/judge/models/tests/test_problem.py
+++ b/judge/models/tests/test_problem.py
@@ -2,7 +2,8 @@ from django.test import TestCase
 from django.utils import timezone
 
 from judge.models import Language, LanguageLimit, Problem
-from judge.models.tests.util import CommonDataMixin, create_problem, create_problem_type, create_solution, create_user
+from judge.models.tests.util import CommonDataMixin, create_organization, create_problem, create_problem_type, \
+    create_solution, create_user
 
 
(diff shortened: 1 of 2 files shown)


judge/models/tests/test_problem.py
```diff
@@ -156,6 +172,62 @@ def test_organization_private_problem_methods(self):
...
+                'is_editable_by': self.assertFalse,
+            },
+            'anonymous': {
+                'is_accessible_by': self.assertFalse,
+                'is_editable_by': self.assertFalse,
+            },
+        }
+        self._test_object_methods_with_users(self.organization_admin_private_problem, data)
+
+    def test_organization_admin_problem_methods(self):
```
kiritofeng, line 210: Shouldn't you also test an organization admin on this problem?

Ninjaclasher: The `problem_organization` organization has `normal` and `staff_problem_edit_public` as organization admins (unless you mean a different organization admin (?)).

kiritofeng: Ah oops I didn't read that correctly.

Trying this with the admin of another organization would make for a good unit test.

Ninjaclasher: Done

PR #1508, opened 2020-08-15
Fix language fixture parameters

zecookiez:
Updated the syntax-highlighting parameters to correspond with the language modes in ace and added a template for Java11.


judge/fixtures/language_small.json
```diff
@@ -263,7 +263,7 @@
          "name":"Brain****",
          "short_name":null,
          "common_name":"Brain****",
-         "ace":"bf",
+         "ace":"plaintext",
```
Ninjaclasher, line 266:
```suggestion
         "ace":"plain_text",
```

Looks like the name should actually be `plain_text`.

zecookiez: Yep that looks more correct, thanks for pointing it out 😅

PR #1516, opened 2020-08-23
Add zero default for contest-output-prefix-length; Fixes #1515

### LINKED ISSUE #1515: Make default in-contest output prefix length 0
This seems to be surprising for many, and can easily lead to ruined contests.

Carson-Tang:
Fixes #1515


TheAvidDev:
Fixes #1515
Hopefully GitHub will pick this up?

PR #1532, opened 2020-09-11
Increase user max org count from 3 to 5

Carson-Tang:


quantum5:
What's the rationale for this change? If this is needed on dmoj.ca, why can't it simply be configured in `local_settings.py`?

Carson-Tang:
Forgot that was an option 🤦

PR #1535, opened 2020-09-13
Ignore over/under-indenting in migrations

kiritofeng:
Django's auto-generated indentation doesn't match what Flake wants, and ignoring will help us catch the real errors.


kiritofeng:
Oops my memory is incorrect, no previous migration has failed due to those errors.

PR #1537, opened 2020-09-18
Explicitly set `case_points` and `case_total` to 0

kiritofeng:
In the case that a rejudge leads to a CE (possible if the submission uses several `pragmas`), these fields will not be updated.

This can then lead to the situation where a rescore results in a CE sub with non-zero score.


quantum5 (approved):
This is trivial enough.

PR #1553, opened 2020-09-30
Add file extension field to Language

### LINKED ISSUE #1540: Add file extension field to Language
The field [already exists](https://github.com/DMOJ/online-judge/blob/master/judge/models/runtime.py#L45), but it is not available in the admin page and does not do anything. It should be added to the admin page so we can finally remove this [nasty `file_extension` property](https://github.com/DMOJ/online-judge/blob/master/judge/models/runtime.py#L91).

What needs to be done:
 1. Add `extension` field to admin page.
 2. Add extensions to `language_small` and `language_all` fixtures.
 3. Remove the `file_extension` property and use the field directly in the `Language` model.
 4. Add a migration (?) to update all extensions for languages supported by the judge server (and only update if the language exists on the site).

Ninjaclasher:
Closes #1540.

### DIFF
--- a/judge/admin/runtime.py
+++ b/judge/admin/runtime.py
@@ -27,7 +27,7 @@ class LanguageForm(ModelForm):
 
 
 class LanguageAdmin(VersionAdmin):
-    fields = ('key', 'name', 'short_name', 'common_name', 'ace', 'pygments', 'info', 'description',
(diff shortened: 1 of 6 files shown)


quantum5 (approved):

Xyene (approved):
LGTM, thanks!

PR #1555, opened 2020-09-30
Correctly pluralize assignees on ticket page

zecookiez:
I also tested the change with the i18n translations and it works properly. 

Not sure whether this is the best way to write it. If there's a better way to change this please let me know :(

### DIFF
--- a/templates/ticket/ticket.html
+++ b/templates/ticket/ticket.html
@@ -357,7 +357,13 @@
                     </div>
                 </div>
                 <div class="info-box">
-                    <div class="info-title">{{ _('Assignees') }}</div>
+                    <div class="info-title">
+                        {% trans trimmed count=assignees|length %}
+                            Assignee
+                        {% pluralize %}
+                            Assignees
+                        {% endtrans %}
+                    </div>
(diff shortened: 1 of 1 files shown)


templates/ticket/ticket.html
```diff
@@ -357,7 +357,13 @@
                     </div>
                 </div>
                 <div class="info-box">
-                    <div class="info-title">{{ _('Assignees') }}</div>
+                    <div class="info-title">
+                        {% trans trimmed count=assignees|length, singular=_("Assignee"), pluralized=_("Assignees") %}
```
TheAvidDev, line 361: See #1554 ([line](https://github.com/DMOJ/online-judge/pull/1554/files#diff-742f815f437aaa2ea7933ce5bbed8efbR67)) it looks cleaner.

zecookiez: I tried something similar but the resulting change broke the translations :/

zecookiez:
Yikes I closed this on accident, I'm going to reopen this again

quantum5 (approved):

PR #1560, opened 2020-10-02
Make Ctrl+Click link open raw source instead

### LINKED ISSUE #1545: Make Ctrl+Click "View source" link open raw source

zecookiez:
Fixes #1545

### DIFF
--- a/templates/submission/status.html
+++ b/templates/submission/status.html
@@ -65,11 +65,21 @@
             });
         </script>
     {% endif %}
+    <script type="text/javascript">
+        $(function () {
+            $("#view-source-link").on("click", function(e) {
+                if (e.metaKey || e.ctrlKey) {
+                    window.open("{{ url('submission_source_raw', submission.id) }}", "_blank");
+                    return false;
+                }
+            });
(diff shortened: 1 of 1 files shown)


TheAvidDev (approved):
LGTM, tested on Firefox 81. Not sure if the issue's intention was to still open a new tab, but that's an easy enough change.

kiritofeng:
IMO should not open in new tab for consistency.

TheAvidDev:
Between @quantum5 and I, it seems that just adding this as another link is preferred. It removes any inconsistencies with how browsers normally behave, is more visible to public users, and just so happens to line up with the amount of space taken up by links in the non-raw source page.

zecookiez:
In this case will updating the title of this PR be sufficient, or should this be closed for a separate one to be opened to reflect the proposed changes? I apologize in advance for my lack of experience with this :s

TheAvidDev:
Because people don't want a link, I think it's best to change Ctrl to a different key like Shift, but this really just requires more discussion from the other contributors.

PR #1561, opened 2020-10-02
Do not run user scripts on the Edit Profile page

### LINKED ISSUE #1544: Do not run user scripts on the edit profile page
I thought we already did this, but apparently we don't. Submitted by a user via email.

WallE256:
Fixes #1544


Xyene (changes requested):
Typo in commit message; functionally looks OK.

I think we should call it `ignore_user_script` though, since we do call it `profile.user_script` rather than `profile.userscript`.

Xyene (approved):

PR #1571, opened 2020-10-27
Make admin notes into a codeblock

ehhthing:
Preview:
image


kiritofeng (approved):
Ability to have newlines is much appreciated :)

PR #1573, opened 2020-11-06
Fix minor submission activity graph issues

Ninjaclasher:
1. Shift weekdays so they match the submission activity graph
2. Prevent infinite scrolling when user has no submissions
3. Update submission activity header text when scrolling

Refs #236.


TheAvidDev (approved):
LGTM, thanks!

Xyene (approved):
Thanks!

PR #1576, opened 2020-11-27
remove lambda function from url lazystr

outloudvi:
Django 2.2.17 with Python 3.7.3. It seems that `lazystr` will return garbage for a lambda function. Therefore, we don't need a lambda function here.

```
>>> from django.utils.functional import lazystr
'23'
>>> lazystr("test")
'test'
>>> lazystr(lambda: 3)
'<function <lambda> at 0x7f00eb7a1598>'
```


quantum5:
The lazy part does not work correctly. Superseded by #1640.

Issue #836, opened 2018-12-15
Not to show others' source code even if you solve the problem


jacky030607:
I installed this judge for our school club and I intend to use it as a platform for our members to practice coding. However, if I solve a problem, I will be able to see others' source code. I want to disable this feature because I want our members to try to make their code more efficient than just copying the code from other's source code.
Is there an option to disable this feature? Thanks.

Xyene:
This was requested (and subsequently closed) in #430.

While I fully agree it would be a nice feature to have, it would be difficult for us to maintain it in the long run, when we wouldn't actively make use of it on [dmoj.ca](https://dmoj.ca). There's an associated risk of future updates breaking it (since we wouldn't necessarily test it), and we don't want to expose functionality that people think should work, but which... doesn't.

That said, a number of forks have had the same need as you in the past, and it's fairly simple to implement (and maintain) this in a fork. You'd be looking at some minor changes to the visibility checks in https://github.com/DMOJ/site/blob/2e416d7a7ef2100372cd3a467d080dce45e0705d/judge/views/submission.py#L50

and

https://github.com/DMOJ/site/blob/88c1aa319250444230589ade04aef2e8bd68a6b9/judge/jinja2/submission.py#L5

Actually looking at this it seems enough has changed since #430 was opened in 2016 that this might be feasible to implement. If someone were to propose a non-intrusive PR to read some sort of `local_settings.py` entry and branch accordingly, we might be able to continue support for it in master. But if the required changes end up being too extensive, it's probably best to run it as a fork.

Ninjaclasher:
Would it be better for this to implemented on a per-problem basis (a `BooleanField` in the `Problem` model)? In this case, it could be used on DMOJ for code golf problems where you don't want users to be able to see other people's code in case a solution better than the author's solution is found.

PR #1582, opened 2021-01-11
Added option DMOJ_PRIVATE_SUBMISSION to hide source code

munhyunsu:
I am using DMOJ as a university teaching tutor.
I want to prevent the students I teach from simply copy-pasting more efficient source code while looking at the answers other students have solved.

This Pull Request provides a private submission function as mentioned in #836.

Use the `DMOJ_PRIVATE_SUBMISSION` flag in `settings.py` or `local_setting.py`.
If `DMOJ_PRIVATE_SUBMISSION = True`, only the source code submitted by the user can be checked in `/submissions`.


judge/jinja2/submission.py
```diff
@@ -15,6 +16,8 @@ def submission_layout(submission, profile_id, user, completed_problem_ids, edita
     elif profile_id == submission.user_id:
         can_view = True
     elif submission.problem_id in completed_problem_ids:
-        can_view = submission.problem.is_public or submission.problem_id in tester_problem_ids
+        can_view = submission.problem_id in tester_problem_ids
+        if not private_submission:
```
fluix-dev, line 20: Instead of passing the `settings.DMOJ_PRIVATE_SUBMISSION` into the context and then re-checking, you should check it here directly instead.

fluix-dev (changes requested):
Thanks for the PR! ~~Considering this was rejected prior, I'm not sure it will be accepted this time, but the code around permissions has certainly cleaned up a bit so I think it's alright.~~ See [comment](https://github.com/DMOJ/online-judge/pull/1582#issuecomment-758024991).

What you've done only affects template rendering, not actual access to the submission page. For example, going to the submission URL or through the API, the submission can still be accessed. You should modify the [`can_see_detail`](https://github.com/DMOJ/online-judge/blob/707c015eb31722299aad9160331fffedcc4cf0b8/judge/models/submission.py#L128) function of the Submission model and write some unit tests to verify it works [here](https://github.com/DMOJ/online-judge/blob/707c015eb31722299aad9160331fffedcc4cf0b8/judge/models/tests/test_submission.py).

fluix-dev:
@Xyene says he's okay with merging, assuming this work.

judge/jinja2/submission.py
munhyunsu: Thank you for the reply.
I did it directly.

fluix-dev:
Totally forgot to mention this, can you change the name of this new setting and make it a string to allow for future changes. From Slack, a while back:
```
xyene: actually
xyene: I'd like to see this be a string
xyene: DMOJ_SUBMISSION_SOURCE_VISIBILITY = 'all' | 'all-solved' | 'only-own'
xyene: I don't expect them to implement the others
xyene: but keep it a string because I'm sure someone will want it
xyene: and having two bool fields is dumb
```

Pretty much, set the current one in `settings.py` to `DMOJ_SUBMISSION_SOURCE_VISIBILITY = 'all'` and then check it's value for the new changes. Don't worry about `'all-solved'` for now and let me know if you need any help.

munhyunsu:
I Understand that it would be useful to more users to provide three options.
Revisions were made according to https://github.com/DMOJ/online-judge/pull/1582#issuecomment-770977456
Thank you for not forgetting to reply.

judge/jinja2/submission.py
```diff
@@ -18,7 +18,9 @@ def submission_layout(submission, profile_id, user, completed_problem_ids, edita
         can_view = True
     elif submission.problem_id in completed_problem_ids:
         can_view = submission.problem_id in tester_problem_ids
-        if not settings.DMOJ_PRIVATE_SUBMISSION:
+        if settings.DMOJ_SUBMISSION_SOURCE_VISIBILITY == 'all-solved':
             can_view = can_view or submission.problem.is_public
+        elif settings.DMOJ_SUBMISSION_SOURCE_VISIBILITY == 'only-own':
```
Xyene, line 23: I think we don't need this branch, since it's a no-op. On the other hand, it'd be good to have a new `settings.DMOJ_SOURCE_VISIBILITY == 'all'` branch at the top of this if/else tree `return True`.

Xyene (commented):
Thanks for taking the time to implement the other options :)

judge/jinja2/submission.py
munhyunsu: I agree.
I also worried about the line the most.
The purpose was to inform the existence of the branch in case somebody edits it in the future, but I know that it is two lines of unnecessary code.
Now that I wanted to match the branch order in the jinja2 and models source code, I think how about removing the 2 lines.
If you agree, I will only remove 2 unnecessary lines.

fluix-dev: I think it's better to remove these two lines. Your concern is totally valid but the proper solution is writing some unit tests.

munhyunsu: I understand.
Thank you for paying attention to the small suggestions and reviewing the code.

fluix-dev (approved):
LGTM, thanks for working on this!

Xyene (approved):
Thanks!

PR #1587, opened 2021-01-30
Expose participation start and end times through API; fixes #1586

### LINKED ISSUE #1586: Expose participation start and end times through API
Title

WallE256:


fluix-dev (approved):
LGTM, I would also add `real_time` though. The fact that these are properties shouldn't be a significant processing overhead, I don't think (the related contest that gets queried is already fetched, and the values are mostly if statements).

Also, don't forget to make a DMOJ/docs PR.

judge/views/api/api_v2.py
```diff
@@ -315,6 +315,8 @@ def get_object_data(self, participation):
         return {
             'user': participation.user.username,
             'contest': participation.contest.key,
+            'start_time': participation.start.isoformat(),
```
Ninjaclasher, line 318: The fields these properties access will need to be added to the `.only()`

- `contest.start_time`
- `contest.end_time`
- `contest.time_limit`
- `real_start`

Ninjaclasher (commented):
Perhaps we should also add these fields to the `APIContestDetail` endpoint, as most people will probably only care about start and end times for a specific contest.

judge/views/api/api_v2.py
```diff
@@ -257,6 +257,8 @@ def get_object_data(self, contest):
             'rankings': [
                 {
                     'user': participation.username,
+                    'start_time': participation.start.isoformat(),
```
Ninjaclasher, line 260: I'm not sure how Django handles fetching related objects through an object, so we need to make sure that we're not fetching `contest` for each participation (there could be > 200 participations in a single contest resulting with > 200 queries).

Also, we might not even need to `select_related`, since we can just overwrite the `contest` attribute with the `contest` we already have (since the contest doesn't change for any of these participations).

JoshuaTianYangLiu: Based on some testing, seems like the way django handles many-to-one relationships is by loading their dependency from the db. Meaning it would result in > 200 queries if there was > 200 participations in a single contest.

I think `select_related` would be the correct option as modifying the `contest` attribute seems a little sketch.

quantum5: Just set `.contest` and be done with it.

kiritofeng (approved):
LGTM

PR #1589, opened 2021-02-04
Consolidate pagination bar spacings

Ninjaclasher:
Everything should line up now...

Todo:
 - [x] The bottom pagination bar is inside a `div` while the top pagination bar is outside. Should probably move either both bars outside or both bars inside.

### DIFF
--- a/resources/widgets.scss
+++ b/resources/widgets.scss
@@ -322,6 +322,16 @@ ul.pagination {
     }
 }
 
+.top-pagination-bar {
+    margin-top: 11px;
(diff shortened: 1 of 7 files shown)


Ninjaclasher:
Any thoughts on whether to move both pagination bars outside the `div`, inside the `div`, or keep as-is?

Ninjaclasher:
Decided to move all the pagination bars outside.

kiritofeng (approved):
LGTM

PR #1594, opened 2021-02-10
Switch from pyotp to secrets for generating backup 2FA tokens

fluix-dev:
pyotp recently released a new version[1] which changes the minimum
length of base32 secret generation to 32 characters.

[1]: https://github.com/pyauth/pyotp/releases/tag/v2.6.0

### DIFF
--- a/judge/models/profile.py
+++ b/judge/models/profile.py
@@ -5,7 +5,6 @@ import secrets
 import struct
 from operator import mul
 
-import pyotp
 import webauthn
 from django.conf import settings
 from django.contrib.auth.models import User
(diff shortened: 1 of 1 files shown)


judge/models/profile.py
```diff
@@ -188,7 +188,9 @@ def generate_api_token(self):
     generate_api_token.alters_data = True
 
     def generate_scratch_codes(self):
-        codes = [pyotp.random_base32(length=16) for i in range(settings.DMOJ_SCRATCH_CODES_COUNT)]
+        def generate_scratch_code():
+            return "".join(secrets.choice(string.ascii_uppercase) for _ in range(16))
+        codes = [generate_scratch_code() for i in range(settings.DMOJ_SCRATCH_CODES_COUNT)]
```
Xyene, line 193: Nit: I know this was like this originally, but s/i/_ for consistency with the new code.

judge/models/profile.py
```diff
@@ -188,7 +188,9 @@ def generate_api_token(self):
     generate_api_token.alters_data = True
 
     def generate_scratch_codes(self):
-        codes = [pyotp.random_base32(length=16) for i in range(settings.DMOJ_SCRATCH_CODES_COUNT)]
+        def generate_scratch_code():
+            return "".join(secrets.choice(string.ascii_uppercase) for _ in range(16))
```
quantum5, line 192: The alphabet size for this new scratch code generator is actually smaller, since it used to be base32 and now it's only the 26 ASCII uppercase letters. There is no need to decrease security in this case.

judge/models/profile.py
fluix-dev: Fixed.

judge/models/profile.py
fluix-dev: Fixed.

quantum5 (approved):

Xyene:
Thanks!

PR #1601, opened 2021-02-20
Use the correct permissions for showing contest admin tab

Ninjaclasher:
The permission `judge.change_contest` isn't used for creating contests, so it doesn't make sense to use it to determine whether to show the admin page tab or not.

This change also makes the contest admin tab logic the same as the [problem admin tab logic](https://github.com/DMOJ/online-judge/blob/master/templates/problem/problem-list-tabs.html#L4).


kiritofeng (approved):
LGTM

Xyene (approved):
Thanks!

PR #1603, opened 2021-02-25
Show admin tabs on own profile page

Ninjaclasher:
There has been quite a number of times I've wanted to view my own profile/user on the admin site, and realized that the shortcut tabs didn't show up on my profile page, leading me to have to search up myself on the admin site and find the correct user.

I'm not sure why these tabs were hidden in the first place (maybe to reduce the number of tabs?), but I feel like showing them would be more beneficial in the long run.


Xyene (approved):
Sure, sounds good. Thanks!

PR #1609, opened 2021-02-28
Allow unlisted users to see contest rank; fixes #1607

### LINKED ISSUE #1607: Unlisted users should still be able to see their rank in-contest
Currently, they can't. The filter is too global.

kiritofeng:
Not yet tested...


Ninjaclasher (approved):
LGTM (tested and nothing crashed). The only thing that is a bit weird is the rating rank computation. Now, on the profile page, we include unlisted users. However, if someone sorts the users list by rating, their rating rank on the users list would not match the rating rank on their profile page. I don't think this is too big of a deal, though.

Xyene (approved):
This LGTM, but it would be nice to fix @Ninjaclasher's comment (maybe separately).

Xyene:
This has acquired conflicts with `judge/ratings.py`.

quantum5 (approved):

PR #1622, opened 2021-03-30
Properly centre forms on login and 2fa pages

Ninjaclasher:
Previously, we were using a relatively positioned `div`, and so each time the form was modified, we had to manually "center" the form by adjusting the CSS values. It also caused some issues with horizontal scrollbars showing up when there should be no scrollbar (notably on the 2fa page).

After this commit, the form should auto-centre even if there are changes, and the horizontal scrollbar no longer appears.

Before login page:
image

After login page:
image

Before 2fa page:
image

After 2fa page:
image

Note that the changes are more noticeable if the images are opened in a new tab and compared side-by-side.

### DIFF
--- a/templates/registration/login.html
+++ b/templates/registration/login.html
@@ -2,14 +2,6 @@
 
 {% block media %}
     <style>
-        #login-panel {
-            display: inline-block;
-            position: relative;
-            margin: 5em auto auto -10em;
-            top: 40%;
-            left: 50%;
-        }
-
         h4 {
             padding-top: 1em;
         }
@@ -35,7 +27,7 @@
 {% endblock %}
 
 {% block body %}
-    <div id="login-panel">
+    <div class="auth-flow-form">
         <form action="" method="post" class="form-area">
             {% csrf_token %}
             {% if form.errors %}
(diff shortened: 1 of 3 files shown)


templates/registration/login.html
```diff
@@ -3,11 +3,11 @@
...
-            display: inline-block;
-            position: relative;
-            margin: 5em auto auto -10em;
-            top: 40%;
-            left: 50%;
+            display: flex;
+            flex-flow: column wrap;
+            justify-content: center;
+            align-content: center;
+            padding-top: 10%;
```
Xyene, line 10:
```suggestion
            padding-top: 5%;
```

templates/registration/two_factor_auth.html
```diff
@@ -3,10 +3,11 @@
...
         #login-panel-2fa {
-            position: relative;
-            margin: 5em auto auto -10em;
-            top: 40%;
-            left: 50%;
+            display: flex;
+            flex-flow: column wrap;
+            justify-content: center;
+            align-content: center;
+            padding-top: 10%;
```
Xyene, line 10:
```suggestion
            padding-top: 5%;
```

Xyene (commented):
Thanks for the cleanup! Since the styles are the same, perhaps they can be extracted into one of the general stylesheets? Some kind of `.auth-flow-form` (or something better-named).

Ninjaclasher:
I've moved the styles as `.auth-flow-form` into `misc.scss` (maybe there's a better file for this?).

quantum5 (approved):

PR #1623, opened 2021-03-31
Add validation on problem statement for disallowed characters; fixes …

### LINKED ISSUE #1612: Add disallowed characters statement validator
We should add a `DMOJ_PROBLEM_STATEMENT_DISALLOWED_CHARACTERS = ('“', '”', '‘', '’') ` settings option, and validate (at least) problem statements to not contain any of the characters within it.

We've tried the social solution, but it's proven hard to enforce.

Carson-Tang:
Fixes #1612


dmoj/settings.py
```diff
@@ -64,6 +64,7 @@
 DMOJ_PROBLEM_MAX_MEMORY_LIMIT = 1048576  # kilobytes
 DMOJ_PROBLEM_MIN_PROBLEM_POINTS = 0
 DMOJ_PROBLEM_HOT_PROBLEM_COUNT = 7
+DMOJ_PROBLEM_STATEMENT_DISALLOW_CHARACTERS = ('“', '”', '‘', '’',)
```
kiritofeng, line 67:
```suggestion
DMOJ_PROBLEM_STATEMENT_DISALLOWED_CHARACTERS = ('“', '”', '‘', '’',)
```

judge/models/problem.py
```diff
@@ -23,6 +24,12 @@
            'License', 'Solution', 'TranslatedProblemQuerySet', 'TranslatedProblemForeignKeyQuerySet']
 
 
+def DisallowCharactersValidator(text):
+    for c in text:
+        if c in settings.DMOJ_PROBLEM_STATEMENT_DISALLOW_CHARACTERS:
+            raise ValidationError(_('Invalid character "%(value)s"'), params={'value': c})
```
kiritofeng, line 30: Potential concern: those who are most likely to make this error are the same ones that cannot distinguish this from the standard quote/double quote.

Xyene: I think rather than "Invalid", we should say "Disallowed" here, since that's more semantically correct.

Carson-Tang: > I think rather than "Invalid", we should say "Disallowed" here, since that's more semantically correct.

fixed

judge/models/problem.py
```diff
@@ -23,6 +24,12 @@
            'License', 'Solution', 'TranslatedProblemQuerySet', 'TranslatedProblemForeignKeyQuerySet']
 
 
+def DisallowCharactersValidator(text):
```
Xyene, line 27: Final nit: perhaps `DisallowedCharactersValidator` is a better name, to match `DISALLOWED_CHARACTERS`?

Though, I guess this is really `disallowed_characters_validator` -- the casing misled me into thinking this was a class-based validator, which it isn't.

Carson-Tang: fixed

judge/models/problem.py
```diff
@@ -23,6 +24,12 @@
            'License', 'Solution', 'TranslatedProblemQuerySet', 'TranslatedProblemForeignKeyQuerySet']
 
 
+def disallowed_characters_validator(text):
+    for c in text:
```
quantum5, line 28: `if set(text) & set(settings.DMOJ_PROBLEM_STATEMENT_DISALLOWED_CHARACTERS)` is probably faster, especially when the setting is turned into a set.

```console
$ python3 -m timeit -s "d = ('“', '”', '‘', '’'); t = '<snippet from Wikipedia>'" $'for c in t:\n if c in d:\n  break'
10000 loops, best of 5: 30.9 usec per loop
$ python3 -m timeit -s "d = ('“', '”', '‘', '’'); t = '<snippet from Wikipedia>'" 'set(t) & set(d)'
50000 loops, best of 5: 5.8 usec per loop
$ python3 -m timeit -s "d = {'“', '”', '‘', '’'}; t = '<snippet from Wikipedia>'" 'set(t) & d'
50000 loops, best of 5: 5.27 usec per loop

File: judge/models/problem.py [hunk 5]
```diff
@@ -23,6 +24,12 @@
            'License', 'Solution', 'TranslatedProblemQuerySet', 'TranslatedProblemForeignKeyQuerySet']
 
 
+def disallowed_characters_validator(text):
```
Ninjaclasher, line 27: Since the infrastructure is set up, please add a few tests for this validator in `tests/problem.py` (https://docs.djangoproject.com/en/2.2/topics/testing/tools/#overriding-settings and https://github.com/DMOJ/online-judge/blob/master/judge/models/tests/test_contest.py#L566 may be useful).

dmoj/settings.py
```diff
@@ -64,6 +64,7 @@
 DMOJ_PROBLEM_MAX_MEMORY_LIMIT = 1048576  # kilobytes
 DMOJ_PROBLEM_MIN_PROBLEM_POINTS = 0
 DMOJ_PROBLEM_HOT_PROBLEM_COUNT = 7
+DMOJ_PROBLEM_STATEMENT_DISALLOWED_CHARACTERS = {'“', '”', '‘', '’',}
```
quantum5, line 67:
```suggestion
DMOJ_PROBLEM_STATEMENT_DISALLOWED_CHARACTERS = {'“', '”', '‘', '’'}
```

Carson-Tang:
Tests added, thanks to @Ninjaclasher

quantum5 (approved):

Ninjaclasher (approved):

Xyene (approved):
Thanks for working on this!

PR #1625, opened 2021-04-02
Exclude unlisted user submissions from total submissions

quantum5:
This prevents weird things like 86.1% AC rate for problems that always
return AC verdicts.

### DIFF
--- a/judge/models/problem.py
+++ b/judge/models/problem.py
@@ -337,12 +337,12 @@ class Problem(models.Model):
         return ProblemClarification.objects.filter(problem=self)
 
     def update_stats(self):
-        self.user_count = self.submission_set.filter(points__gte=self.points, result='AC',
-                                                     user__is_unlisted=False).values('user').distinct().count()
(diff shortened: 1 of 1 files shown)


Ninjaclasher (commented):
Unlisted users were originally excluded from problem stats so that the stats matched up with what was shown on the Best Submissions page (since unlisted users are removed from the Best Submissions page). The oversight was in including unlisted user submissions into the total submission count that is used for the AC rate [here](https://github.com/DMOJ/online-judge/pull/1625/files#diff-31cc24428c587c9e058846d7d9fd2eb98d89b15f7b5f33ff0eb915e6446b6944L342).

That said, including unlisted users in the problem stats is probably fine, and the code looks good. However, if problem stats are supposed to match up with the Best Submissions page (which is what it looked like back when unlisted users were implemented), then the related code for excluding unlisted users in that page should also be removed.

kiritofeng (approved):
LGTM

Xyene (approved):

PR #1627, opened 2021-04-04
Added scoreboard_visibility

JoshuaTianYangLiu:
This will expose the full scoreboard once your window is over compared to the `hide_scoreboard` option which will expose the full scoreboard once the contest is over.

Essentially taken from https://github.com/mcpt/wlmoj/pull/22/


judge/models/contest.py
```diff
@@ -77,6 +77,10 @@ class Contest(models.Model):
     view_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('view contest scoreboard'), blank=True,
                                                      related_name='view_contest_scoreboard',
                                                      help_text=_('These users will be able to view the scoreboard.'))
+    partially_hide_scoreboard = models.BooleanField(verbose_name=_('Partially hide scoreboard'),
```
Xyene, line 80: I think instead of having a new boolean this, we should change the existing `hide_scoreboard` field into a `scoreboard_visibility = (Visible | Hidden_for_duration_of_contest | Hidden_for_duration_of_participation | Hidden) ChoiceField`. Then the migration would map `hide_scoreboard = True` to `scoreboard_visibility = Hidden_for_duration_of_contest`.

Xyene:
Apart from that, thanks for the PR -- it largely looks good.

JoshuaTianYangLiu:
What should I do about parts of the code which references `hide_scoreboard`? Change them to accordingly to use `scoreboard_visibility`? What about the api?

Ninjaclasher:
I think the right approach is to switch everything from using `hide_scoreboard` to use `scoreboard_visibility`. API should just be a single line change from `hide_scoreboard: boolean` to `scoreboard_visibility: string`.

JoshuaTianYangLiu:
Ok changes applied, this will also require a slight change on https://github.com/DMOJ/docs but that's for tomorrow me.
 
Please double check if I switched every correctly as i'm not familiar with some parts of the code base and I did my best educative guess.

judge/models/contest.py
```diff
@@ -70,13 +70,15 @@ class Contest(models.Model):
...
-                                          help_text=_('Whether the scoreboard should remain hidden for the duration '
-                                                      'of the contest.'),
-                                          default=False)
     view_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('view contest scoreboard'), blank=True,
                                                      related_name='view_contest_scoreboard',
                                                      help_text=_('These users will be able to view the scoreboard.'))
+    scoreboard_visibility = models.CharField(verbose_name=_('scoreboard visibility'), default='Visible', max_length=32,
+                                             help_text=_("Scoreboard visibility through the duration of the contest"),
+                                             choices=[
+                                                 ('Visible', _('Visible')),
```
quantum5, line 79: There is no need to make the actual string in the database more than one or two characters long. They should be introduced as constants on the class to avoid typos.

judge/migrations/0115_contest_scoreboard_visibility.py
```diff
@@ -0,0 +1,24 @@
...
+        ('judge', '0114_contest_partially_hide_scoreboard'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='contest',
+            name='scoreboard_visibility',
+            field=models.CharField(choices=[('Visible', 'Visible'), ('Hidden_for_duration_of_contest', 'Hidden for duration of contest'), ('Hidden_for_duration_of_participation', 'Hidden for duration of participation')], default='Visible', help_text='Scoreboard visibility through the duration of the contest', max_length=64, verbose_name='scoreboard visibility'),
+        ),
+        migrations.RunPython(hide_scoreboard_eq_true, atomic=True),
```
quantum5, line 23: Migration does not reverse the operation nor delete the defunct `hide_scoreboard` field.

judge/views/problem.py
```diff
@@ -422,7 +422,7 @@ def get_context_data(self, **kwargs):
         else:
             context['hot_problems'] = None
             context['point_start'], context['point_end'], context['point_values'] = 0, 0, {}
-            context['hide_contest_scoreboard'] = self.contest.hide_scoreboard
+            context['hide_contest_scoreboard'] = self.contest.scoreboard_visibility == "Contest"
```
kiritofeng, line 425: This leaks the problem's user count if the visibility is set to `Participation`: https://github.com/DMOJ/online-judge/blob/master/templates/problem/list.html#L292

judge/models/tests/test_contest.py
```diff
@@ -42,7 +42,7 @@ def setUpTestData(self):
             start_time=_now - timezone.timedelta(days=1),
             end_time=_now + timezone.timedelta(days=100),
             is_visible=True,
-            hide_scoreboard=True,
+            scoreboard_visibility="Contest",
```
kiritofeng, line 45: Ideally also add tests for the other visibility options

judge/models/contest.py
```diff
@@ -190,13 +194,23 @@ def can_see_full_scoreboard(self, user):
...
+            if participation and participation.ended:
+                return True
         return False
 
     @cached_property
     def show_scoreboard(self):
         if not self.can_join:
             return False
-        if self.hide_scoreboard and not self.ended:
+        if ((self.scoreboard_visibility == 'Contest' or self.scoreboard_visibility == 'Participation') and
```
kiritofeng, line 212:
```suggestion
        if (self.scoreboard_visibility in ('Contest', 'Participation') and
```

judge/views/api/api_v2.py
```diff
@@ -236,7 +236,7 @@ def get_object_data(self, contest):
             'has_rating': contest.ratings.exists(),
             'rating_floor': contest.rating_floor,
             'rating_ceiling': contest.rating_ceiling,
-            'hidden_scoreboard': contest.hide_scoreboard,
```
kiritofeng, line 239: Do we want to keep this in for backwards compatibility? (`hidden_scoreboard': contest.scoreboard_visibility in ('Contest', 'Participation')`)

/cc @Xyene @quantum5

judge/views/submission.py
```diff
@@ -245,7 +245,7 @@ def _get_queryset(self):
             if not self.request.user.has_perm('judge.see_private_contest'):
                 # Show submissions for any contest you can edit or visible scoreboard
                 contest_queryset = Contest.objects.filter(Q(organizers=self.request.profile) |
```
kiritofeng, line 247: I think you also need a clause to check for `Participation` visibility contests where your participation has ended, cc @Xyene @quantum5

JoshuaTianYangLiu:
I haven't added test cases for participation in scoreboard visibility.
I assume you'll probably want backwards compatability in the api, so I changed it as suggested.
I'm unsure about the implementation in submission.py so I'm still waiting for a response from  @Xyene or @quantum5

judge/models/contest.py
```diff
@@ -54,6 +54,11 @@ class Meta:
 
 
 class Contest(models.Model):
+    SCOREBOARD = (
```
kiritofeng, line 57:
```suggestion
    SCOREBOARD_VISIBILITY = (
```

judge/models/tests/test_contest.py
```diff
@@ -42,7 +42,7 @@ def setUpTestData(self):
             start_time=_now - timezone.timedelta(days=1),
             end_time=_now + timezone.timedelta(days=100),
             is_visible=True,
-            hide_scoreboard=True,
+            scoreboard_visibility="C",
```
kiritofeng, line 45:
```suggestion
            scoreboard_visibility='C',
```

judge/views/problem.py
```diff
@@ -422,7 +422,7 @@ def get_context_data(self, **kwargs):
         else:
             context['hot_problems'] = None
             context['point_start'], context['point_end'], context['point_values'] = 0, 0, {}
-            context['hide_contest_scoreboard'] = self.contest.hide_scoreboard
+            context['hide_contest_scoreboard'] = self.contest.scoreboard_visibility == "V"
```
kiritofeng, line 425:
```suggestion
            context['hide_contest_scoreboard'] = self.contest.scoreboard_visibility == 'V'
```

judge/views/submission.py
```diff
@@ -245,7 +245,7 @@ def _get_queryset(self):
...
-                                                          Q(hide_scoreboard=False) |
+                                                          Q(scoreboard_visibility="V") |
```
kiritofeng, line 248:
```suggestion
                                                          Q(scoreboard_visibility='V') |
```

judge/models/contest.py
```diff
@@ -70,13 +75,12 @@ class Contest(models.Model):
...
                                    default=False)
-    hide_scoreboard = models.BooleanField(verbose_name=_('hide scoreboard'),
-                                          help_text=_('Whether the scoreboard should remain hidden for the duration '
-                                                      'of the contest.'),
-                                          default=False)
     view_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('view contest scoreboard'), blank=True,
                                                      related_name='view_contest_scoreboard',
                                                      help_text=_('These users will be able to view the scoreboard.'))
+    scoreboard_visibility = models.CharField(verbose_name=_('scoreboard visibility'), default='V', max_length=1,
+                                             help_text=_("Scoreboard visibility through the duration of the contest"),
```
kiritofeng, line 82:
```suggestion
                                             help_text=_('Scoreboard visibility through the duration of the contest'),
```

kiritofeng (commented):
Please use single quote for strings unless they contain an apostrophe :slightly_smiling_face:

judge/migrations/0114_contest_scoreboard_visibility.py
```diff
@@ -0,0 +1,28 @@
...
+        ('judge', '0113_contest_decimal_points'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='contest',
+            name='scoreboard_visibility',
+            field=models.CharField(choices=[('V', 'Visible'), ('C', 'Hidden for duration of contest'), ('P', 'Hidden for duration of participation')], default='V', help_text='Scoreboard visibility through the duration of the contest', max_length=64, verbose_name='scoreboard visibility'),
+        ),
+        migrations.RunPython(hide_scoreboard_eq_true, atomic=True),
```
kiritofeng, line 23: You're missing the [reverse operation](https://docs.djangoproject.com/en/2.2/ref/migration-operations/#django.db.migrations.operations.RunPython).

judge/models/contest.py
```diff
@@ -54,6 +54,11 @@ class Meta:
 
 
 class Contest(models.Model):
+    SCOREBOARD = (
+        ('V', _('Visible')),
```
quantum5, line 58: Would suggest that you create constants like `SCOREBOARD_VISIBLE` on the `Contest` to avoid future typos.

judge/migrations/0114_contest_scoreboard_visibility.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def hide_scoreboard_eq_true(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    Contest.objects.filter(hide_scoreboard=True).update(scoreboard_visibility='C')
+
+
+def scoreboard_visibility_eq_contest(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    Contest.objects.filter(scoreboard_visibility='C').update(hide_scoreboard=True)
```
kiritofeng, line 13: I think `'P'` (Participation) should also be marked as `hide_scoreboard=True`

judge/migrations/0114_contest_scoreboard_visibility.py
```diff
@@ -0,0 +1,33 @@
...
+
+    dependencies = [
+        ('judge', '0113_contest_decimal_points'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='contest',
+            name='scoreboard_visibility',
+            field=models.CharField(choices=[('V', 'Visible'), ('C', 'Hidden for duration of contest'), ('P', 'Hidden for duration of participation')], default='V', help_text='Scoreboard visibility through the duration of the contest', max_length=64, verbose_name='scoreboard visibility'),
```
kiritofeng, line 26: `max_length` doesn't match: the model has `1`, migration has `64`.

judge/models/contest.py
```diff
@@ -54,6 +54,14 @@ class Meta:
 
 
 class Contest(models.Model):
+    SCOREBOARD_VISIBLE = 'V'
+    SCOREBOARD_CONTEST = 'C'
```
quantum5, line 58: It's not exactly clear how scoreboards are linked with contests. Should be something like `SCOREBOARD_AFTER_CONTEST`.

judge/models/contest.py
```diff
@@ -54,6 +54,14 @@ class Meta:
...
+    SCOREBOARD_PARTICIPATION = 'P'
```
quantum5, line 59: Similarly, should be something like `SCOREBOARD_AFTER_PARTICIPATION`.

judge/views/submission.py
kiritofeng: Discussed with quantum in Slack, we should not run such a query.

judge/models/tests/test_contest.py
JoshuaTianYangLiu: I'm unsure how to simulate a user outside their contest window. There's an example for a user in a contest though.

judge/models/contest.py
```diff
@@ -70,13 +78,12 @@ class Contest(models.Model):
...
     is_rated = models.BooleanField(verbose_name=_('contest rated'), help_text=_('Whether this contest can be rated.'),
                                    default=False)
-    hide_scoreboard = models.BooleanField(verbose_name=_('hide scoreboard'),
-                                          help_text=_('Whether the scoreboard should remain hidden for the duration '
-                                                      'of the contest.'),
-                                          default=False)
     view_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('view contest scoreboard'), blank=True,
                                                      related_name='view_contest_scoreboard',
                                                      help_text=_('These users will be able to view the scoreboard.'))
+    scoreboard_visibility = models.CharField(verbose_name=_('scoreboard visibility'), default='V', max_length=1,
```
Ninjaclasher, line 84:
```suggestion
    scoreboard_visibility = models.CharField(verbose_name=_('scoreboard visibility'), default=SCOREBOARD_VISIBLE, max_length=1,
```

judge/models/contest.py
```diff
@@ -179,6 +186,8 @@ def can_see_own_scoreboard(self, user):
             return False
         if not self.show_scoreboard and not self.is_in_contest(user):
             return False
+        if self.scoreboard_visibility == self.SCOREBOARD_AFTER_PARTICIPATION and not self.is_in_contest(user):
```
Ninjaclasher, line 189: Do we need this check? Shouldn't it be covered by the check 2 lines above?

judge/views/problem.py
```diff
@@ -422,7 +422,7 @@ def get_context_data(self, **kwargs):
         else:
             context['hot_problems'] = None
             context['point_start'], context['point_end'], context['point_values'] = 0, 0, {}
-            context['hide_contest_scoreboard'] = self.contest.hide_scoreboard
+            context['hide_contest_scoreboard'] = self.contest.scoreboard_visibility == self.contest.SCOREBOARD_VISIBLE
```
Ninjaclasher, line 425: I don't think this condition is correct. Shouldn't it be `!=`? Making specifying an `in` would be easier to read.

judge/models/contest.py
JoshuaTianYangLiu: I just checked, yes that statement seems useless

Ninjaclasher (commented):
The code looks good, but this definitely needs some tests since there's so many checks for scoreboard visibility and it's very easy to get wrong. At the very least, a test for each `SCOREBOARD_*` choice, and a test for `has_completed_contest` (user has never joined contest, user has joined contest but participation hasn't ended, and user has joined contest but participation has ended).

JoshuaTianYangLiu:
I'm pretty sure I only have to add a test case for `SCOREBOARD_AFTER_PARTICIPATION` as the other two are already added as `hidden_scoreboard_contest` and `basic_contest`.

judge/models/tests/test_contest.py
```diff
@@ -42,14 +51,61 @@ def setUpTestData(self):
...
+            key='visible_scoreboard',
+            start_time=_now - timezone.timedelta(days=10),
+            end_time=_now + timezone.timedelta(days=100),
+            time_limit=timezone.timedelta(days=1),
+            is_visible=True,
+            scoreboard_visibility=Contest.SCOREBOARD_VISIBLE,
+        )
+
+        for contest_key in ('contest_scoreboard', 'particip_scoreboard', 'visible_scoreboard'):
+            create_contest_participation(
```
kiritofeng, line 90: This doesn't look right, you want to see what happens to a user that has not joined the contest (i.e. has no participations)

judge/models/tests/test_contest.py
```diff
@@ -42,14 +51,61 @@ def setUpTestData(self):
...
+                contest=contest_key,
+                user='normal_before_window',
+            ).save()
+
+            create_contest_participation(
+                contest=contest_key,
+                user='normal_during_window',
+                real_start=_now - timezone.timedelta(hours=1),
+                virtual=ContestParticipation.LIVE,
```
kiritofeng, line 99: You should also add a test for `particip_scoreboard` and `virtual = 1` (i.e. not live participation) to ensure it handles that case correctly.

JoshuaTianYangLiu: I'm assuming `virtual = -1` for spectate?

judge/models/tests/test_contest.py
```diff
@@ -42,14 +51,70 @@ def setUpTestData(self):
...
+            ).save()
+
+            create_contest_participation(
+                contest=contest_key,
+                user='normal_after_window',
+                real_start=_now - timezone.timedelta(days=3),
+                virtual=ContestParticipation.LIVE,
+            ).save()
```
Ninjaclasher, line 102: I don't think calling `save()` on any of these is necessary.

judge/models/tests/test_contest.py
```diff
@@ -42,14 +51,70 @@ def setUpTestData(self):
...
+            user='normal',
+            real_start=_now - timezone.timedelta(days=3),
+            virtual=ContestParticipation.LIVE,
+        ).save()
+
+        create_contest_participation(
+            contest='particip_scoreboard',
+            user='normal',
+            real_start=_now + timezone.timedelta(days=101),
+            virtual=-1,
```
Ninjaclasher, line 115:
```suggestion
            virtual=ContestParticipation.SPECTATE,
```

judge/models/tests/test_contest.py
```diff
@@ -210,6 +275,197 @@ def test_hidden_scoreboard_contest_methods(self):
         }
         self._test_object_methods_with_users(self.hidden_scoreboard_contest, data)
 
+    def test_contest_hidden_scoreboard_contest_methods(self):
+        data = {
+            'staff_contest_edit_own': {
```
Ninjaclasher, line 280: Hmm... do we need to test these specific staff users for all the types of hidden scoreboards? There doesn't seem to be anything special in the methods that differentiates these users. At the very least, I think we should remove the other methods besides `can_see_own_scoreboard`, `can_see_full_scoreboard`, and `has_completed_contest` from all three tests. They don't really add anything (Maybe it's better to test everything together, but personally I try to make each test function to focus on a specific section of the contest code and then one function `test_basic_contest_methods` to test everything together).

judge/migrations/0114_contest_scoreboard_visibility.py
```diff
@@ -0,0 +1,33 @@
...
+
+def scoreboard_visibility_eq_contest(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    Contest.objects.filter(scoreboard_visibility__in=('C', 'P')).update(hide_scoreboard=True)
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0113_contest_decimal_points'),
```
Ninjaclasher, line 19: Rebase please.

judge/models/tests/test_contest.py
JoshuaTianYangLiu: We don't need the specific staff users for these test, but I thought it would be nice to keep it as a redundancy.

judge/models/tests/test_contest.py
```diff
@@ -42,14 +51,70 @@ def setUpTestData(self):
...
+            )
+
+            create_contest_participation(
+                contest=contest_key,
+                user='normal_after_window',
+                real_start=_now - timezone.timedelta(days=3),
+                virtual=ContestParticipation.LIVE,
+            )
+
+        create_contest_participation(
```
Ninjaclasher, line 104: What's the difference between this participation and `normal_after_window`?

JoshuaTianYangLiu: This was added to simulate a user joining a contest, and spectating after. Line 111 adds another participation for spectating the contest

Ninjaclasher (approved):
LGTM

quantum5 (approved):

PR #1628, opened 2021-04-06
Filter submissions by language before generating rankings

### LINKED ISSUE #1616: /rank language filters act after the query for fetching the best submissions instead of before
Imagine that user 1 has solved problem A in C and Python 3, and their C submission is faster than their submission in Python 3. When we load /rank, we see the user's C submission and not their Python submission. Imagine that we now have the language filter and enable it for Python 3.

Current behaviour:
We don't see the best Python 3 submission from user 1 because it is shadowed by the C submission.

Requested behaviour:
We expect to see the best Python 3 submission from user 1.

quantum5:
This shows a user's best submission in the filtered language(s) even if
it is not the user's personal best. This fixes #1616.

### DIFF
--- a/judge/views/ranked_submission.py
+++ b/judge/views/ranked_submission.py
@@ -2,6 +2,7 @@ from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.translation import gettext as _
 
+from judge.models import Language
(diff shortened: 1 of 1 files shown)


kiritofeng (approved):
Yolo

PR #1630, opened 2021-04-06
Always show top border line on submission table

### LINKED ISSUE #1598: When live updating submissions list, don't create new submission row until after Ajax call
Currently, when live updating, if there is a new submission, we first [append the new row and hide it](https://github.com/DMOJ/online-judge/blob/master/templates/submission/list.html#L158). The row only becomes visible after the following Ajax call. If the Ajax call takes a while, it leads to the following visual bug (occurs on both Chrome and Firefox):
image
Note: the top border of the list is not visible.

We should probably change it so that the new row is only created once the Ajax call finishes (in the `.done` section). This also fixes the following two issues:
 - There is a period of time when there are only 49 submissions on the list (after the bottom submission is removed and before the top submission becomes visible).
 - If the Ajax call fails, the submission list will become shorter and shorter since the bottom submission gets removed but the top submission never becomes visible.

quantum5:
This works with an arbitrary number of empty elements anywhere in the table.

Fixes #1598

### DIFF
--- a/resources/submission.scss
+++ b/resources/submission.scss
@@ -14,6 +14,7 @@
 
 .submission-row {
     display: flex;
+    border-top: #ccc 1px solid;
     border-left: #ccc 1px solid;
     border-right: #ccc 1px solid;
     transition: background-color linear 0.2s;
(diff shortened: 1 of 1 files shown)


kiritofeng (approved):
Yolo

PR #1633, opened 2021-04-08
Require a solved problem to update profile about

quantum5:
This prevents spam bots from posting sketchy links.

Tested on a test site.

### DIFF
--- a/judge/models/profile.py
+++ b/judge/models/profile.py
@@ -10,7 +10,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.validators import RegexValidator
 from django.db import models
-from django.db.models import Max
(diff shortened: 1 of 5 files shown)


judge/models/profile.py
```diff
@@ -150,6 +150,10 @@ def organization(self):
     def username(self):
         return self.user.username
 
+    @cached_property
+    def has_any_ac(self):
```
kiritofeng, line 154: nit: "AC" might not be the best name (as you're technically not checking for an AC, and having 100/100 doesn't necessarily imply AC)

kiritofeng:
Tagging #1487.

kiritofeng (approved):
LGTM

PR #1638, opened 2021-04-12
Limit width of registration form for mobile

leduythuccs:
Set max-width of the registration form to 90% to avoid overflow in small screen phones.

### DIFF
--- a/templates/registration/registration_form.html
+++ b/templates/registration/registration_form.html
@@ -79,9 +79,6 @@
         #edit-form {
             border: unset;
             background: unset;
-            max-width: 100%;
-            width: 450px;
-            margin: 0 -10px 0 -15px;
         }
 
         #content-body {
@@ -91,7 +88,7 @@
         #center-float {
             display: inline-block;
             text-align: initial;
-            max-width: 90%;
+            max-width: 100%;
             width: 450px;
         }
 
--- a/templates/registration/registration_form.html
+++ b/templates/registration/registration_form.html
@@ -79,7 +79,7 @@
         #edit-form {
             border: unset;
             background: unset;
-            max-width: 450px;
+            max-width: 100%;
             width: 450px;
         }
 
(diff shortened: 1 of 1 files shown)


templates/registration/registration_form.html
```diff
@@ -79,8 +79,9 @@
         #edit-form {
             border: unset;
             background: unset;
-            max-width: 450px;
+            max-width: 100%;
             width: 450px;
+            margin: 0 -10px 0 -15px;
```
quantum5, line 84: If you set `#edit-form`'s max-width to 90% and `#center-float`'s `max-width` to 100%, you don't need this margin hack.

leduythuccs: It makes the form not fully centering on large screens (i.e laptop)

quantum5 (approved):

templates/registration/registration_form.html
```diff
@@ -79,8 +79,9 @@
         #edit-form {
             border: unset;
             background: unset;
-            max-width: 450px;
+            max-width: 100%;
```
Ninjaclasher, line 82: Does `max-width: 450px` and `width:100%` (for both elements) without the margin hack not work?

leduythuccs: Yep it doesn't work. The form inputs still go outside the screen.

Ninjaclasher: Hmm it seems the form originally was not centred anyways. I would suggest setting `max-width: 450px` and `width: 100%` on `#center-float` and removing both `max-width` and `width` on `#edit-form`, which seems to centre things and doesn't overflow on a small screen without needing the negative margin. Maybe someone else should check to make sure this is reasonable.

leduythuccs: Removing both `max-width` and `width` on `#edit-form` works. Thanks. I committed my changes.

quantum5 (approved):

Ninjaclasher (approved):

PR #1643, opened 2021-04-16
Add is_organization_private to automatic problem rescore check

Ninjaclasher:
`is_organization_private` problems are ignored when computing
a user's points, so turning a problem organization private
(and vice versa) must recompute that user's points.


kiritofeng (approved):
LGTM

Xyene (approved):
Thanks!

PR #1644, opened 2021-04-17
Use gettext_lazy to translate title

leduythuccs:
Using `gettext` will not translate the title of the login view (I still don't know why). I used `gettext_lazy` and it works. 

Maybe there is some more place that didn't translate correctly but I haven't found any yet.


quantum5 (approved):

quantum5:
As for why it works, see the [Django documentation](https://docs.djangoproject.com/en/3.2/topics/i18n/translation/#lazy-translation).

PR #1647, opened 2021-04-20
Don't set negative margin followed by padding on ticket messages

Ninjaclasher:
This fixes an issue where part of the ticket title cannot be selected
because the first ticket message is overlapping with the title text.

Reported by @int-y1.


kiritofeng (approved):
Thanks!

PR #1652, opened 2021-04-23
Improve balancing between judges with same load

### LINKED ISSUE #1581: Better judge load balancing
We currently always dispatch to the least loaded judge, but on very light loads, all loads are 0 and we always select the oldest judge (by uptime). This does not appear to be fair.

quantum5:
This PR has not been tested.

Fixes #1581.


kiritofeng (approved):
LGTM

PR #1655, opened 2021-04-23
Add minor improvements to user data downloads

Ninjaclasher:
- Filtering by problem code takes a long time. When a request is initiated by the user, they're brought to this screen, with no progress bar, until the filtering finishes. Since the filtering takes quite a while for large amounts of submissions, let's add a progress bar for filtering just so the page isn't blank (and there's something to display to the user, even if it's not much).
image
 - Speed up filtering by problem code by using `fnmatch.translate` and then a native regex filter.
    - [`fnmatch.fnmatch` translates to regex anyways](https://github.com/python/cpython/blob/master/Lib/fnmatch.py#L45).
 - Make sure we grab `problem`, `language`, and `source` when we fetch the submissions, since we need them anyways.

### DIFF
--- a/judge/tasks/user.py
+++ b/judge/tasks/user.py
@@ -20,16 +20,10 @@ def apply_submission_filter(queryset, options):
     if options['submission_results']:
         queryset = queryset.filter(result__in=options['submission_results'])
 
-    problem_code_glob_cache = {}
+    if options['submission_problem_glob'] != '*':
+        queryset = queryset.filter(problem__code__regex=fnmatch.translate(options['submission_problem_glob']))
 
-    def problem_code_glob_match(obj):
-        code = obj.problem.code
-        result = problem_code_glob_cache.get(code)
-        if result is None:
-            result = problem_code_glob_cache[code] = fnmatch.fnmatch(code, options['submission_problem_glob'])
-        return result
-
-    return list(filter(problem_code_glob_match, queryset))
+    return list(queryset)
 
 
 def apply_comment_filter(queryset, options):
@@ -41,8 +35,17 @@ def apply_comment_filter(queryset, options):
 @shared_task(bind=True)
 def prepare_user_data(self, profile_id, options):
     options = json.loads(options)
-    submissions = apply_submission_filter(Submission.objects.filter(user_id=profile_id), options)
-    comments = apply_comment_filter(Comment.objects.filter(author_id=profile_id), options)
+    with Progress(self, 2, stage=_('Applying filters')) as p:
+        # Force an update so that we get a progress bar.
+        p.done = 0
+        submissions = apply_submission_filter(
+            Submission.objects.select_related('problem', 'language', 'source').filter(user_id=profile_id),
+            options,
+        )
+        p.did(1)
+        comments = apply_comment_filter(Comment.objects.filter(author_id=profile_id), options)
+        p.did(1)
+
     with zipfile.ZipFile(os.path.join(settings.DMOJ_USER_DATA_CACHE, '%s.zip' % profile_id), mode='w') as data_file:
         submission_count = len(submissions)
         if submission_count:
--- a/judge/tasks/user.py
+++ b/judge/tasks/user.py
@@ -7,8 +7,9 @@ from celery import shared_task
 from django.conf import settings
 from django.utils.translation import gettext as _
 
-from judge.models import Comment, Submission
+from judge.models import Comment, Problem, Submission
 from judge.utils.celery import Progress
+from judge.utils.raw_sql import use_straight_join
 from judge.utils.unicode import utf8bytes
 
 __all__ = ('prepare_user_data',)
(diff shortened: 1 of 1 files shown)


judge/tasks/user.py
```diff
@@ -20,16 +20,10 @@ def apply_submission_filter(queryset, options):
     if options['submission_results']:
         queryset = queryset.filter(result__in=options['submission_results'])
 
-    problem_code_glob_cache = {}
+    if options['submission_problem_glob'] != '*':
+        queryset = queryset.filter(problem__code__regex=fnmatch.translate(options['submission_problem_glob']))
```
quantum5, line 24: Might be interesting to experiment the subquery version to see if it offers any performance gains (perhaps also forcing a `STRAIGHT JOIN`):

```python
queryset = queryset.filter(problem__in=Problem.objects.filter(code__regex=fnmatch.translate(options['submission_problem_glob'])))
```

Also the MySQL regex flavour is not necessarily the same as that of python. Will this cause any issues?

quantum5 (commented):
Otherwise looks good.

judge/tasks/user.py
Ninjaclasher: > Might be interesting to experiment the subquery version to see if it offers any performance gains (perhaps also forcing a `STRAIGHT JOIN`):
> 
> ```python
> queryset = queryset.filter(problem__in=Problem.objects.filter(code__regex=fnmatch.translate(options['submission_problem_glob'])))
> ```

With some testing on a table size of 100k submissions and 600 problems (unfortunately I don't have a database any larger):

```py
import fnmatch
import timeit

from judge.models import Problem, Submission
from judge.utils.raw_sql import use_straight_join

glob = r'a*'

def f():
    a = Submission.objects.select_related('problem', 'language', 'source').filter(problem__code__regex=fnmatch.translate(glob))
    list(a)

def g():
    a = Submission.objects.select_related('problem', 'language', 'source').filter(problem__in=Problem.objects.filter(code__regex=fnmatch.translate(glob)))
    list(a)

def h():
    a = Submission.objects.select_related('problem', 'language', 'source')
    use_straight_join(a)
    a = a.filter(problem__code__regex=fnmatch.translate(glob))
    list(a)

def i():
    a = Submission.objects.select_related('problem', 'language', 'source')
    use_straight_join(a)
    a = a.filter(problem__in=Problem.objects.filter(code__regex=fnmatch.translate(glob)))
    list(a)


for func in 'fghi':
    print(func, timeit.timeit(f'{func}()', globals=globals(), number=5))
```

```
f 11.460055768024176
g 10.058658519992605
h 10.914277172996663
i 10.024998323991895
```

It seems `g` or `i` is slightly faster (this has been reflected in the latest commit).

> Also the MySQL regex flavour is not necessarily the same as that of python. Will this cause any issues?

From testing with very weird strings, it seems that the subset of regex that is used by `fnmatch.translate` is also valid for MySQL. However, I can't find any online sources to confirm this. In particular, I'm unsure about the implications of the use of [re.escape](https://github.com/python/cpython/blob/master/Lib/fnmatch.py#L137) in `fnmatch.translate`. For now, I think it should be safe to use `fnmatch.translate` directly. If it ever does become an issue though, we could always restrict the character set to only `a-z0-9*?[]!`, since problem codes can't have any other characters anyways.

judge/tasks/user.py
```diff
@@ -17,19 +18,18 @@
...
-
-    def problem_code_glob_match(obj):
-        code = obj.problem.code
-        result = problem_code_glob_cache.get(code)
-        if result is None:
-            result = problem_code_glob_cache[code] = fnmatch.fnmatch(code, options['submission_problem_glob'])
-        return result
+    if options['submission_problem_glob'] != '*':
+        queryset = queryset.filter(
+            problem__in=Problem.objects.filter(code__regex=fnmatch.translate(options['submission_problem_glob'])),
```
quantum5, line 29: `fnmatch.translate` seems to be vulnerable to the generation of pathological expressions:

```pycon
>>> import fnmatch, re
>>> re.match(fnmatch.translate('*'*100+'b'), 'a'*1000)
<does not finish in a reasonable timeframe>
```

However, the query
```sql
SELECT 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' REGEXP '(?s:.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*c)\\Z';
```
does seem to complete quickly. I guess this is okay?

quantum5 (approved):

judge/tasks/user.py
quantum5: On a second thought it's probably best if we did `s/\*+/*/g` on the glob pattern.

PR #1658, opened 2021-04-26
Reduce time complexity for chain of * in problem filter

quantum5:
This has not been integration tested, but the snippets work.

### DIFF
--- a/judge/tasks/user.py
+++ b/judge/tasks/user.py
@@ -1,6 +1,7 @@
 import fnmatch
 import json
 import os
+import re
 import zipfile
 
 from celery import shared_task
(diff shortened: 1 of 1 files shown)


Ninjaclasher (approved):
Hmm it seems I didn't run into this because this exact issue was fixed in Python 3.9 ([commit](https://github.com/python/cpython/commit/b9c46a2c2d7fc68457bff641f78932d66f5e5f59#diff-5078809c85d4fa8475e78198661cd5800368dc1e0f958b93513aac70155dd789)) and I was on Python 3.9 instead of Python 3.8. It might be good to add a comment saying this was fixed in Python 3.9, but we need the regex to be sure nothing bad happens for < 3.9.

kiritofeng (approved):

PR #1660, opened 2021-04-27
Convert submission lock, rejudged date, and contest lock to timestamps

### LINKED ISSUE #1604: Convert submission lock to timestamp
Instead of a boolean field, a submission should have a `locked_on` timestamp, and a timestamp for `rejudged_on`.

This would allow for a rejudge of a problem with locked submissions to set said submission's point value to 0, preventing such oddities as this:
image

Ninjaclasher:
This does all the changes listed in #1604 and fixes #1604. It also converts contest lock to a timestamp, since multiple people have requested pre-locking a contest before it ends (but only taking effect when the contest ends), and it feels reasonable to convert both in the same PR.

~Todo:~
 - [X] ~Determine the best way of setting a submission's point value to 0. In particular:~
      - ~To fix the visual oddity in #1604, we would need to set `case_total` to be the new `case_total`. However, we don't know what the new `case_total` is without rejudging the submission. We also can't touch `case_points`, since that's what's used to compute contest points.~
      - ~What do we do about the `result`/`status`? I think there's three options:~
         - ~Create a new `result` for submissions rejudged after locking. However, this would ovewrite the original `result`, which may be useful to the submission author.~
         - ~Create a new` status`. We could ovewrite `D` (`Done`) with a new status for submissions rejudged after locking.~
         - ~Don't touch `result`/`status` and keep the original. However, this would mean we get submissions that have i.e `0/100 AC`, but opening up the submission shows a non-zero amount of AC cases, which might be confusing to users.~

~Some suggestions for how to handle the issues listed above would be appreciated.~

~It might be better to move the "rejudge of a problem with locked submissions should set submission point's value to 0" to a new issue (and new PR), but for now, I'm marking this PR as a draft until we decide to whether to include it in this PR or move to a different PR.~

Implementing the above will be moved to a separate PR.


Ninjaclasher:
After some further thought, I think it's better to move the "rejudge of a problem with locked submissions should set submission point's value to 0" to a separate PR.

judge/migrations/0116_convert_to_dates.py
```diff
@@ -0,0 +1,61 @@
...
+    Contest.objects.filter(is_locked=True).update(locked_after=F('end_time'))
+
+
+def convert_to_boolean(apps, schema_editor):
+    Contest = apps.get_model('judge', 'Contest')
+    Submission = apps.get_model('judge', 'Submission')
+
+    Submission.objects.filter(rejudged_date__isnull=False).update(was_rejudged=True)
+    Submission.objects.filter(locked_after__isnull=False).update(is_locked=True)
+    Contest.objects.filter(locked_after__isnull=False).update(is_locked=True)
```
kiritofeng, line 23: nit: I would only set this if the date is after `timezone.now()`, although in practice I highly doubt this matters

judge/tasks/submission.py
```diff
@@ -16,7 +16,7 @@ def apply_submission_filter(queryset, id_range, languages, results):
         queryset = queryset.filter(language_id__in=languages)
     if results:
         queryset = queryset.filter(result__in=results)
-    queryset = queryset.filter(is_locked=False).exclude(status__in=Submission.IN_PROGRESS_GRADING_STATUS)
+    queryset = queryset.exclude(status__in=Submission.IN_PROGRESS_GRADING_STATUS)
```
kiritofeng, line 19: Do you handle the equivalent of `filter(is_locked=False)` elsewhere?

judge/migrations/0116_convert_to_dates.py
Ninjaclasher: Updated.

judge/tasks/submission.py
Ninjaclasher: Readded the filter with `locked_after` (accidentally deleted it for testing and forgot to add it back).

kiritofeng (approved):
LGTM

judge/views/submission.py
```diff
@@ -179,8 +179,12 @@ def get(self, request, *args, **kwargs):
...
 def abort_submission(request, submission):
     submission = get_object_or_404(Submission, id=int(submission))
-    if (not request.user.is_authenticated or (submission.was_rejudged or (request.profile != submission.user)) and
-            not request.user.has_perm('judge.abort_any_submission')):
+    if not request.user.is_authenticated:
+        raise PermissionDenied()
+    if (
+        not request.user.has_perm('judge.abort_any_submission') and
+        (submission.rejudged_date is not None or request.profile != submission.user)
+    ):
```
quantum5, line 187: We typically use this style.

```suggestion
    if (request.user.has_perm('judge.abort_any_submission') and
        (submission.rejudged_date is not None or request.profile != submission.user)):
```

judge/admin/contest.py
```diff
@@ -233,19 +233,19 @@ def set_locked(self, request, queryset):
...
 
-    def set_is_locked(self, contest, is_locked):
+    def set_locked_after(self, contest, locked_after):
         with transaction.atomic():
-            contest.is_locked = is_locked
+            contest.locked_after = locked_after
             contest.save()
             Submission.objects.filter(contest_object=contest,
-                                      contest__participation__virtual=0).update(is_locked=is_locked)
+                                      contest__participation__virtual=0).update(locked_after=locked_after)
```
quantum5, line 248: This will not affect submissions submitted after the lock time was set, but before the time of the lock.

Ninjaclasher: Fixed.

judge/views/submission.py
```diff
@@ -181,8 +181,10 @@ def get(self, request, *args, **kwargs):
 @require_POST
 def abort_submission(request, submission):
     submission = get_object_or_404(Submission, id=int(submission))
-    if (not request.user.is_authenticated or (submission.was_rejudged or (request.profile != submission.user)) and
-            not request.user.has_perm('judge.abort_any_submission')):
+    if not request.user.is_authenticated:
```
Ninjaclasher, line 184: Note: this `if` is actually unecessary (`not request.user.has_perm('judge.abort_any_submission')` will be `True, `and `request.profile != submission.user` will also be `True`). Should it be removed or should we keep it in for clarity?

Ninjaclasher:
Decided to add a few small tests for `is_locked`.

judge/views/submission.py
quantum5: It used to be needed from before the `request.profile` days, just drop it.

judge/views/problem.py
```diff
@@ -619,25 +619,32 @@ def form_valid(self, form):
...
-            self.new_submission = form.save()
+            self.new_submission = form.save(commit=False)
 
             contest_problem = self.contest_problem
             if contest_problem is not None:
-                # See ContestSubmission post_save hook in judge/signals.py
-                self.new_submission.contest_object_id = contest_problem.contest_id
+                # Use the contest object from current_contest.contest because we already use it
+                # in profile.update_contest().
+                # See ContestSubmission post_save hook in judge/signals.py.
```
quantum5, line 628: This comment is no longer necessary.

judge/views/problem.py
```diff
@@ -619,25 +619,32 @@ def form_valid(self, form):
...
+                self.new_submission.contest_object = self.request.profile.current_contest.contest
+                if self.request.profile.current_contest.live:
+                    self.new_submission.locked_after = self.new_submission.contest_object.locked_after
+                self.new_submission.save()
                 ContestSubmission(
                     submission=self.new_submission,
-                    problem=contest_problem,
+                    problem=self.contest_problem,
```
quantum5, line 635: Is there any point to adding `self.` here?

Ninjaclasher: Oops, was testing something and forgot to revert this :facepalm:. Fixed.

judge/views/submission.py
Ninjaclasher: Done.

judge/models/contest.py
```diff
@@ -139,8 +139,9 @@ class Contest(models.Model):
                                             help_text='A custom Lua function to generate problem labels. Requires a '
                                                       'single function with an integer parameter, the zero-indexed '
                                                       'contest problem index, and returns a string, the label.')
-    is_locked = models.BooleanField(verbose_name=_('contest lock'), default=False,
-                                    help_text=_('Prevent submissions from this contest from being rejudged.'))
+    locked_after = models.DateTimeField(verbose_name=_('contest lock'), null=True, blank=True,
```
kiritofeng, line 142: I think you should note that this is (currently) not retroactive, otherwise some might think this runs a cron in the backgroun.

kiritofeng: Ignore, I missed something.

quantum5 (approved):
Doesn't look broken to me.

kiritofeng (approved):
LGTM

PR #1662, opened 2021-05-01
Clear flat page cache on save

leduythuccs:
Currently, the flat page is being cached for 1 day and not being cleaned up even the flat page has been edited.


quantum5:
As far as I can tell, the flatpage templates do not use the caching mechanism. Can you link to the place where the cache is used?

leduythuccs:
Holy, I cached the flat pages in my repo months ago and completely forgot about it. I am so sorry about this confusion. I will close the PR.

PR #1664, opened 2021-05-02
Pretty format caniuse.json

quantum5:
This is so that diffs actually make sense.


kiritofeng (approved):
Gonna assume that `resources/caniuse.json` diff is just you pretty printing it

PR #1668, opened 2021-05-06
Add contest curator and testers

### LINKED ISSUE #1556: Add contest tester field
The reasoning for this is the same as for problem testers; problem testers can be used as a model for this feature.

kiritofeng:
Fixes #1556.


kiritofeng:
Obligatroy speling iz harrd

judge/admin/contest.py
```diff
@@ -293,7 +295,7 @@ def get_form(self, request, obj=None, **kwargs):
             form.base_fields['problem_label_script'].widget = AceWidget('lua', request.profile.ace_theme)
 
         perms = ('edit_own_contest', 'edit_all_contest')
-        form.base_fields['organizers'].queryset = Profile.objects.filter(
+        form.base_fields['curators'].queryset = Profile.objects.filter(
```
quantum5, line 298: Why is this not needed for authors?

judge/models/contest.py
```diff
@@ -316,14 +346,16 @@ def is_accessible_by(self, user):
...
     def is_editable_by(self, user):
         # If the user can edit all contests
         if user.has_perm('judge.edit_all_contest'):
             return True
 
-        # If the user is a contest organizer
-        if user.has_perm('judge.edit_own_contest') and \
-                self.organizers.filter(id=user.profile.id).exists():
+        # If the user is a contest organizer or curator
+        if user.has_perm('judge.edit_own_contest') and user.profile.id in self.editor_ids:
```
quantum5, line 358: What's the justification for using `in self.editor_ids` here and `is_editor` in `can_see_full_scoreboard` and `access_check`?

templates/contest/list.html
```diff
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.organizers.all() %}
+            {% if request.profile in contest.editor_ids or request.profile in contest.tester_ids %}
```
quantum5, line 127: Does this actually work? I would have expected `request.profile.id in ...`.

judge/admin/contest.py
kiritofeng: We want authors to not require staff permissions.

This would allow for a model where an mod creates the contest, but tickets are filed to the author, and would also allow for a situation where we could add helpers (say to look over submissions for cheating/MOSS) without having to give them contest editing perms.

templates/contest/list.html
kiritofeng: Should be `request.profile.id`, my bad.

judge/models/contest.py
kiritofeng: Changed both to use `editor_ids`.

judge/views/contests.py
```diff
@@ -130,10 +130,16 @@ class ContestMixin(object):
...
         if not self.request.user.is_authenticated:
             return False
-        return self.object.organizers.filter(id=self.request.profile.id).exists()
+        return self.object.is_editor(self.request.profile)
+
+    @cached_property
+    def is_tester(self):
+        if not self.request.user.is_authenticated:
+            return False
+        return self.object.testers.filter(id=self.request.profile.id).exists()
```
quantum5, line 142: Is there any reason why we aren't just looking up `self.object.tester_ids`?

judge/views/contests.py
```diff
@@ -130,10 +130,16 @@ class ContestMixin(object):
     slug_url_kwarg = 'contest'
 
     @cached_property
-    def is_organizer(self):
+    def is_editor(self):
         if not self.request.user.is_authenticated:
             return False
-        return self.object.organizers.filter(id=self.request.profile.id).exists()
+        return self.object.is_editor(self.request.profile)
```
quantum5, line 136: Is there any reason why we aren't just looking up `self.object.editor_ids`?

kiritofeng:
Removed `Contest.object.is_editor(...)` entirely, and replaced all instances with `... in editor_ids` instead.

quantum5 (approved):
Nothing seems obviously broken to me.

templates/contest/list.html
```diff
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.organizers.all() %}
+            {% if request.profile.id in contest.editor_ids or request.profile.id in contest.tester_ids %}
```
Ninjaclasher, line 127: I'm not sure if using `editor_ids` and `tester_ids` here is a good idea. Doesn't it cause a query regardless of `prefetch_related`? It might be a better to just use `contest.authors.all()`, `contest.curators.all()`, and `contest.testers.all()` directly.

judge/views/contests.py
```diff
@@ -78,7 +78,7 @@ def _now(self):
         return timezone.now()
 
     def _get_queryset(self):
-        return super(ContestList, self).get_queryset().prefetch_related('tags', 'organizations', 'organizers')
+        return super(ContestList, self).get_queryset().prefetch_related('tags', 'organizations', 'authors', 'curators')
```
Ninjaclasher, line 81: Should we prefetch `testers`? (Both here and the one below).

judge/views/contests.py
```diff
@@ -130,10 +130,16 @@ class ContestMixin(object):
...
         if not self.request.user.is_authenticated:
             return False
-        return self.object.organizers.filter(id=self.request.profile.id).exists()
+        return self.request.profile.id in self.object.editor_ids
+
+    @cached_property
+    def is_tester(self):
+        if not self.request.user.is_authenticated:
+            return False
+        return self.request.profile.id in self.object.tester_ids
```
Ninjaclasher, line 142: Just curious, why do we use `profile.id in contest.tester_ids` here but `contest.testers.filter(id=user.profile.id).exists()` is `contest.is_accessible_by`?

judge/migrations/0116_contest_curator_and_tester.py
```diff
@@ -0,0 +1,33 @@
...
+    operations = [
+        migrations.AlterField(
+            model_name='contest',
+            name='organizers',
+            field=models.ManyToManyField(help_text='These users will be able to edit the contest.', related_name='_contest_authors_+', to='judge.Profile'),
+        ),
+        migrations.RenameField(
+            model_name='contest',
+            old_name='organizers',
+            new_name='authors',
```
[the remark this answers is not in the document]
kiritofeng: Authors is a required field, which is why I elected to keep it this way, I am willing to accept some historical inaccuracy, although other opinions would be good

templates/contest/list.html
kiritofeng: The ids should be marked with `@cached_property`, so maybe not(?), I'll have to test that

judge/views/contests.py
```diff
@@ -334,14 +341,14 @@ def join_contest(self, request, access_code=None):
             LIVE = ContestParticipation.LIVE
             try:
                 participation = ContestParticipation.objects.get(
-                    contest=contest, user=profile, virtual=(SPECTATE if self.is_organizer else LIVE),
+                    contest=contest, user=profile, virtual=(SPECTATE if self.is_editor or self.is_tester else LIVE),
```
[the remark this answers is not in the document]
kiritofeng: Marking as out of scope for this PR, will file a separate ticket for this.

templates/contest/list.html
kiritofeng: Because we `prefetch_related`, `contest.authors.all()` is the better choice.

judge/views/contests.py
kiritofeng: Consistency hard.

judge/models/tests/test_contest.py
```diff
@@ -43,7 +47,7 @@ def setUpTestData(self):
             key='basic',
             start_time=_now - timezone.timedelta(days=1),
             end_time=_now + timezone.timedelta(days=100),
-            organizers=('superuser', 'staff_contest_edit_own'),
+            authors=('superuser', 'staff_contest_edit_own'),
         )
```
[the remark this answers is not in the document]
kiritofeng: Added tests.

judge/admin/contest.py
```diff
@@ -151,7 +153,7 @@ def get_queryset(self, request):
         if request.user.has_perm('judge.edit_all_contest'):
             return queryset
         else:
-            return queryset.filter(organizers__id=request.profile.id)
+            return queryset.filter(Q(authors__id=request.profile.id) | Q(curators__id=request.profile.id))
```
Ninjaclasher, line 156:
```suggestion
            return queryset.filter(Q(authors=request.profile) | Q(curators=request.profile))
```
Seems I missed this on the first pass.

templates/contest/list.html
```diff
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.organizers.all() %}
+            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() %}
```
Ninjaclasher, line 127:
```suggestion
            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() or request.profile in contest.testers.all() %}
```
I think we need testers here as well to be consistent with the contest detail page. (not sure if there's a better way to write this line than a triple `or`)

Ninjaclasher (commented):
Otherwise LGTM

Ninjaclasher (approved):

PR #1672, opened 2021-05-07
Allow contest curators and authors to see submissions made in contest

kiritofeng:


judge/views/submission.py
```diff
@@ -391,6 +392,12 @@ def access_check_contest(self, request):
             raise Http404()
 
     def access_check(self, request):
+        # FIXME: This should be rolled into the `is_accessible_by` check when
+        # implementing #1509
+        if self.contest and request.user.is_authenticated \
+                and request.profile.id in self.contest.editor_ids:
```
quantum5, line 398:
```suggestion
        if self.contest and request.user.is_authenticated and request.profile.id in self.contest.editor_ids:
```

judge/views/submission.py
```diff
@@ -391,6 +392,12 @@ def access_check_contest(self, request):
             raise Http404()
 
     def access_check(self, request):
+        # FIXME: This should be rolled into the `is_accessible_by` check when
+        # implementing #1509
```
quantum5, line 396:
```suggestion
        # FIXME: This should be rolled into the `is_accessible_by` check when implementing #1509
```

quantum5 (approved):

PR #1677, opened 2021-05-07
Remove unneeded `profile.update_contest()` call in `ProblemSubmit`

Ninjaclasher:
There is a middleware that calls `profile.update_contest()` (see https://github.com/DMOJ/online-judge/blob/master/judge/middleware.py#L75), so I don't think there's any need to call it here again. I think this code was a remnant of when there was no middleware and the view was a function based view (see 6847439)


quantum5 (approved):
Man, trying to implement the contest system in one night 7 years ago still results in repercussions to this day...

PR #1681, opened 2021-05-08
Consistently cache global submission stats

quantum5:
Instead of using the problems visible to whatever user populating
the cache, we simply cache stats all submissions on the site.

This has not been tested.


kiritofeng (approved):
LGTM

PR #1686, opened 2021-05-13
Use correct spacing for rating column in contest rankings

quantum5:
This avoids stuff like the following when the word for "rating" in the current language is too short:


kiritofeng (approved):
LGTM

PR #1687, opened 2021-05-13
ratings: correctly handle unrated users in SQL

quantum5:
Otherwise, last_rating will be NULL in the query, and rating floors
and ceilings will not work. Using COALESCE fixes this.


kiritofeng (approved):
Note that the change to the default volatility is due to a typo in the previous PR.

PR #1701, opened 2021-05-29
Filter contest submissions properly in live updates

### LINKED ISSUE #1688: The submission list of a contest shows non-contest submissions because of dynamic updating
Problem: If I enter a submissions list of a user in a contest (e.g: https://dmoj.ca/contest/aac2/submissions/leduykhongngu/)  and that user submits a submission to any problem (even the problem does not belong in the contest), that submission will appear in the submission list.

image

Ninjaclasher:
Fixes #1688.

### DIFF
--- a/judge/views/submission.py
+++ b/judge/views/submission.py
@@ -286,6 +286,8 @@ class SubmissionsListBase(DiggPaginatorMixin, TitleMixin, ListView):
         context = super(SubmissionsListBase, self).get_context_data(**kwargs)
         authenticated = self.request.user.is_authenticated
         context['dynamic_update'] = False
+        if self.in_contest:
+            context['dynamic_contest_id'] = self.contest.id
         context['show_problem'] = self.show_problem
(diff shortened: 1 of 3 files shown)


judge/views/submission.py
```diff
@@ -286,6 +286,8 @@ def get_context_data(self, **kwargs):
         context = super(SubmissionsListBase, self).get_context_data(**kwargs)
         authenticated = self.request.user.is_authenticated
         context['dynamic_update'] = False
+        if self.in_contest:
+            context['dynamic_contest_id'] = self.contest.id
```
quantum5, line 290:
```suggestion
        context['dynamic_contest_id'] = self.in_contest and self.contest.id
```

This is mostly so that the variable is always defined. `django-jinja` turns undefined variables into `None`, but in normal jinja2 this would cause the template to crash.

quantum5 (commented):
Otherwise looks good.

quantum5 (approved):

PR #1702, opened 2021-05-30
Atomically rate multiple contests

quantum5:
This prevents data corruption when a contest that is not the most recent
is rerated, and something happens before rating is complete all contests.

Also remove the return value of `rate_contest` because it's never used.


kiritofeng (approved):
LGTM, he said, at 4AM

Ninjaclasher (approved):

Issue #1693, opened 2021-05-21
Add option to refresh TOTP credentials


quantum5:
This should create a workflow similar to that of enabling TOTP, but is used when the user already has TOTP enabled. Once this workflow is completed successfully, including entering the new OTP, the TOTP secret should be replaced, and users will only be able to login using tokens generated with the new secret.

The intention of this workflow is to allow staff members to reset their TOTP credentials, as they are forbidden from disabling 2FA, and thus cannot disable TOTP without a WebAuthn device.

PR #1708, opened 2021-06-05
Refactor TOTPEnableView in preparation for #1693

quantum5:
This will pave the way for #1693.

Changes:

* the new TOTP key in `request.session`, allowing key changes without destroying the current key
* use a special `TOTPEnableForm` to reject scratch codes when enabling TOTP
* make TOTP code easily selectable and copy without whitespace

This was actually tested on a test site.

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -18,12 +18,16 @@ from judge.models import Contest, Language, Organization, Problem, Profile, Subm
 from judge.utils.subscription import newsletter_id
 from judge.widgets import HeavyPreviewPageDownWidget, Select2MultipleWidget, Select2Widget
 
+TOTP_CODE_LENGTH = 6
+
 two_factor_validators_by_length = {
-    6: {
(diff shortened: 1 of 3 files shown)


judge/forms.py
```diff
@@ -18,8 +18,10 @@
 from judge.utils.subscription import newsletter_id
 from judge.widgets import HeavyPreviewPageDownWidget, Select2MultipleWidget, Select2Widget
 
+TOTP_CODE_LENGTH = 6
+
 two_factor_validators_by_length = {
-    6: {
+    TOTP_CODE_LENGTH: {
         'regex_validator': RegexValidator('^[0-9]{6}$',
```
kiritofeng, line 25: `6` is still hardcoded here

quantum5: Fixed.

kiritofeng (approved):
Seems legit

PR #1713, opened 2021-06-05
Allow WebAuthn devices to be managed in admin

### LINKED ISSUE #1698: Add WebAuthn enable/disable on the user profile in the admin view

quantum5:
This fixes #1698. Note that we cannot add WebAuthn devices for the user, so
that part is not and will never be implemented.

Setting is_webauthn_enabled to False when removing all WebAuthn devices is
moved to a signal so that it also works when deleting credentials in the admin.


kiritofeng (approved):
Seems legit

PR #1715, opened 2021-06-05
Use html.unescape instead of deprecated HTMLParser.unescape

quantum5:


fluix-dev:
Is the `parser` itself used anywhere else?

quantum5:
> Is the `parser` itself used anywhere else?

Apparently not.

fluix-dev (approved):
Then LGTM.

kiritofeng (approved):

PR #1716, opened 2021-06-06
Allow scratch codes with only WebAuthn

### LINKED ISSUE #1690: Allow scratch codes when a user only has WebAuthn
Currently, we don't display the scratch code field nor check them if the user has only WebAuthn. Neither is the user able to generate scratch codes (at least this is consistent).

I think scratch codes should be allowed in case the user has lost all security keys, but we should see what the standard practice is first.

quantum5:
This fixes #1690.


judge/forms.py
```diff
@@ -250,8 +250,9 @@ def clean(self):
 
             credential.counter = sign_count
             credential.save(update_fields=['counter'])
-        elif self.profile.is_totp_enabled and totp_or_scratch_code:
-            if pyotp.TOTP(self.profile.totp_key).verify(totp_or_scratch_code, valid_window=self.TOLERANCE):
+        elif totp_or_scratch_code:
+            if (self.profile.is_totp_enabled and
+                    pyotp.TOTP(self.profile.totp_key).verify(totp_or_scratch_code, valid_window=self.TOLERANCE)):
```
Ninjaclasher, line 255: I think the error message farther down (L263) should probably be updated.

quantum5: Fixed.

Ninjaclasher (approved):

kiritofeng (approved):
I believe in @Ninjaclasher

PR #1033, opened 2019-08-13
Add option to hide problem authors while in contest

Ninjaclasher:
Fixes #1032 .

Issue #1032, opened 2019-08-13
for contests, add the ability to hide problem creators


xiaowuc1:
This would be similar to `hide problem tags` in functionality.

PR #1718, opened 2021-06-06
Add option to hide problem authors while in contest

Ninjaclasher:
Refs #1032. (not sure why the issue was closed)

Addressing the issues that were faced in the previous attempt (#1033):
 - With #1668, we will assign the contest authors as assignees to prevent leaking problem authors.
 - The API no longer shows problems that can only be accessed when in a contest, so that route won't leak authors either.
 
Have not tested the changes yet, so will create a draft PR for now.


Ninjaclasher:
Tested and works.

kiritofeng (approved):
LGTM

quantum5 (approved):

PR #1720, opened 2021-06-06
Prevent the reuse of TOTP authentication codes

### LINKED ISSUE #1692: Prevent the reuse of TOTP tokens
Currently, TOTP codes can be reused as long as they are within the valid window. An attacker that is able to obtain a TOTP token, say through shoulder surfing, is able to also login. This leaves a lot to be desired. For example, after #1691, an attacker that was somehow able to obtain an OTP could reuse said OTP to change 2FA settings.

Instead, we should keep track of the last used counter ID, and reject OTPs associated with counters that are not greater than the last used counter ID.

quantum5:
We store the timecodes in an IntegerField, which is guaranteed to support values up to 2147483647. This is okay because timecodes are Unix time divided by 30, so it supports values up to the year 4011.

This fixes #1692.

Ideally, we'd add a unit test that attempts to verify the same code twice, then mock change the time and see that it does indeed validate with the new code.


kiritofeng (approved):
lgtm

PR #1722, opened 2021-06-07
Ensure `edit_own_problem` is a prerequisite for `edit_all_problem`

Ninjaclasher:
`judge.edit_own_problem` is already a required permission for admin-related tasks, so if a user has only `judge.edit_all_problem` without `judge.edit_own_problem`, they'll get a lot of links to 400/404 pages.

Not sure why you would give `judge.edit_all_problem` without `judge.edit_own_problem` in the first place, but we should probably ensure everything is consistent.

### DIFF
--- a/judge/models/tests/test_problem.py
+++ b/judge/models/tests/test_problem.py
@@ -13,6 +13,14 @@ class ProblemTestCase(CommonDataMixin, TestCase):
     def setUpTestData(self):
         super().setUpTestData()
 
+        self.users.update({
+            'staff_problem_edit_only_all': create_user(
+                username='staff_problem_edit_only_all',
+                is_staff=True,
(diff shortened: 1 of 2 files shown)


judge/models/tests/test_problem.py
```diff
@@ -13,6 +13,14 @@ class ProblemTestCase(CommonDataMixin, TestCase):
     def setUpTestData(self):
         super().setUpTestData()
 
+        self.users.update({
```
Ninjaclasher, line 16: Note to reviewer: I kept this user here instead of in `tests/util.py` where all the other problem permission related users are because I felt this user wouldn't be very useful outside of this specific testcase. Should it be moved with the other problem related users?

Xyene (approved):
LGTM, thanks!

PR #1725, opened 2021-06-12
Store rounded participation score in database

### LINKED ISSUE #1595: Float precision issue in contest ranking
In ContestParticipation model, score field is a float. This may lead to float precision issues in judge.views.contests.contest_ranking_list:

``` python
def contest_ranking_list(contest, problems):
    return base_contest_ranking_list(contest, problems, contest.users.filter(virtual=0, user__is_unlisted=False)
                                     .prefetch_related('user__organizations')
                                     .order_by('is_disqualified', '-score', 'cumtime', 'tiebreaker'))
```

In my own fork, I get 2 users having the same displayed score in the contest ranking 2.86 and they pass the same number of tests. But in the database, the one with lower time has score in database 2.8569999... and the other has 2.857. This leads to a bug in the contest ranking page.  The solution is to round this number to around 6 or 8 decimal digits in `contest_ranking_list` or maybe after judging.

Ninjaclasher:
Fixes #1595.

### DIFF
--- a/judge/contest_format/atcoder.py
+++ b/judge/contest_format/atcoder.py
@@ -90,7 +90,7 @@ class AtCoderContestFormat(DefaultContestFormat):
                 points += score
 
         participation.cumtime = cumtime + penalty
-        participation.score = points
+        participation.score = round(points, self.contest.points_precision)
         participation.tiebreaker = 0
         participation.format_data = format_data
         participation.save()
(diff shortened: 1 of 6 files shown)


kiritofeng (approved):
LGTM; only remaining question is if it's worth it to run this computation against all currently existing participations for consistency.

quantum5 (approved):

PR #1727, opened 2021-06-17
Implement per-problem submission access control

quantum5:
The same setting used to be only available globally, but can now be applied
per-problem.


kiritofeng (approved):
LGTM

PR #1729, opened 2021-06-18
Remove unused `malformed_packet` method

Ninjaclasher:
This method is never used and the parent class does not even have a `malformed_packet`. Probably (?) related to 22cbce731.


quantum5 (approved):
Seems legit.

PR #700, opened 2018-09-12
Display contest options in short form

Xyene:
Addresses #695.

PR #1732, opened 2021-06-19
Implement contest settings short display

### LINKED ISSUE #695: Contest settings short display
We currently allow a lot of customization for contests, and rely on contest authors to write a descriptive-enough contest body to describe the options.

For example, we allow customization of:
* pretests
* virtual/windowed
* comments allowed
* rated

As well as possible future features:

* time penalty
* scoreboard freeze settings
* feedback format

...and so on.

We should have a section on the contest description page where these options are laid out in a standard format.

Ninjaclasher:
This fixes #695 and supercedes #700.

We add an option to turn off this short display so that information is not duplicated for existing contests.

Todo:
 - [x] Add contest format-specific information.


Ninjaclasher:
Screenshot: 
image

judge/contest_format/ecoo.py
```diff
@@ -125,3 +125,25 @@ def display_participation_result(self, participation):
...
+        first_ac_bonus = self.config['first_ac_bonus']
+        if first_ac_bonus:
+            yield _(
+                'There is a **%d bonus** for fully solving on your first non-CE submission.',
+            ) % first_ac_bonus
+
+        time_bonus = self.config['time_bonus']
+        if time_bonus:
+            yield ungettext(
+                'For every **1 minute** you submit before the end of your window, there will be a **1** point bonus.',
```
kiritofeng, line 141: Why do you hardcode the `1` here but no where else?

judge/views/contests.py
```diff
@@ -234,8 +234,23 @@ def get_context_data(self, **kwargs):
...
+        context['metadata'] = {
+            'has_public_editorials': any(
+                problem.is_public and problem.has_public_editorial for problem in context['contest_problems']
+            ),
+        }
+        context['metadata'].update(
+            **self.object.contest_problems
+            .annotate(
+                partials_enabled=F('partial').bitand(F('problem__partial')),
+                pretests_enabled=F('is_pretested').bitand(F('contest__run_pretests_only')),
```
kiritofeng, line 246: Can someone confirm that this will work even after systesting? cc @Xyene @quantum5

templates/contest/contest.html
```diff
@@ -65,6 +65,115 @@
...
+                                {% trans trimmed count=testers|length, link_testers=link_users(testers) %}
+                                    Special thanks to {{ link_testers }} for testing and feedback on the problems.
+                                {% pluralize count %}
+                                    Special thanks to {{ link_testers }} for testing and feedback on the problems.
+                                {% endtrans %}
+                            </li>
+                        {% endif %}
+                    {% endwith %}
+                    <li>
+                        {% if contest.is_rated %}
```
kiritofeng, line 94: Is there a way to clean this up, say in the view using formatted strings?

templates/contest/contest.html
```diff
@@ -65,6 +65,115 @@
...
+                                {% endif %}
+                            {% endif %}
+                        {% else %}
+                            {{ _('This contest will **not** be rated.')|markdown('default') }}
+                        {% endif %}
+                    </li>
+                    <li>
+                        {% filter markdown('default') %}
+                            {% trans trimmed count=metadata.problem_count %}
+                                There is **1** problem in this contest.
```
kiritofeng, line 123: Is hardcoded 1 a good idea? (i18n is pain)

kiritofeng (changes requested):
Potential i18n problems are pain~~

templates/contest/contest.html
Ninjaclasher: fixed

judge/contest_format/ecoo.py
Ninjaclasher: fixed

templates/contest/contest.html
Ninjaclasher: We could break up different segments of that sentence and join them together, but I'm not sure if that will make translations possible/correct. Alternatively, we could use a nested list and list the rating ceiling/floor restrictions there, but I originally opted to not go that route as it may be confusing to a user to see "rated for all participants" followed by some rating restrictions.

judge/views/contests.py
Ninjaclasher: This unfortunately won't be correct after systests, but I don't see any way to correctly grab that information after the settings are unchecked. FWIW this is also how it was done in #700, so it might be fine (?)

templates/contest/contest.html
kiritofeng: i18n is pain...

this gets an "ok fine then" from me, I'll see what others have to say on this

judge/views/contests.py
quantum5: This is a defect in the design of the pretest system. What we really want is two different booleans, but that's out of scope here.

templates/contest/contest.html
quantum5: We definitely must use discrete sentences or colon-style for i18n. Colon-style will be confusing, so I think this is the best we can do.

maxcruickshanks (changes requested):
There is a typo for penalties: ```Each submission before the first maximum score submission with incur a **penalty of %d minute**.```. It should be ```will incur``` instead of ```with incur```.

Markdown is not capitalized: ```Returns a generator of markdown strings```.

A is not supposed to be capitalized: ```:return A generator, where each item is an individual line.```

```The scoreboard will be **hidden** for the duration of the contest.``` seems vague.
Roger's potential replacement: ```the entire duration of the contest```.

Ninjaclasher:
> There is a typo for penalties: `Each submission before the first maximum score submission with incur a **penalty of %d minute**.`. It should be `will incur` instead of `with incur`.

Fixed.

> Markdown is not capitalized: `Returns a generator of markdown strings`.

Fixed. 

> A is not supposed to be capitalized: `:return A generator, where each item is an individual line.`

The `A` is supposed to be captialized (see the other docstrings). However, I was missing a `:` after `:return`, which is now added.

> `The scoreboard will be **hidden** for the duration of the contest.` seems vague.
> Roger's potential replacement: `the entire duration of the contest`.

Updated

kiritofeng (approved):
LGTM

quantum5 (approved):
Looks good, will deploy after current rated contest finishes.

Ninjaclasher:
Rebased migration.

Ninjaclasher:
Rebased migration again.

quantum5 (approved):

PR #1734, opened 2021-06-20
Remove unused ticket CSS "messages" class

Ninjaclasher:
This class is never used, and also conflicts with other styles
for "messages".

For example, there is padding for list items when there shouldn't be: https://github.com/DMOJ/online-judge/blob/master/resources/widgets.scss#L492

Before:
image

After (what it looks like in the preview):
image


kiritofeng (approved):
I believe you

quantum5 (approved):

PR #1735, opened 2021-06-26
Remove duplicated and unused `manifest.json`

Ninjaclasher:
This file is a duplicate of `resources/icons/manifest.json`, and
shouldn't be used. The one in `resources/icons/manifest.json`
should be used, and is the one used if the site setup documentation
is followed.

It looks like this file was misplaced in #411, and it looked as if the file was missing, so was duplicated (in the correct directory) in 653299104fca6d4f45d623a72a00be1dc5211e18.


kiritofeng (approved):
I believe you

quantum5 (approved):

Issue #1742, opened 2021-08-04
Wrong `old_rating` value in /contest API


JoshuaTianYangLiu:
If two rated contests end at the same time, both will show the same `old_rating` in the `/contest` endpoint

The only example where this error can be found is with contest `mockccc15s` and `mockccc15j`

User: https://dmoj.ca/user/quantum

https://dmoj.ca/api/v2/contest/mockccc15j
`"old_rating": 1976, "new_rating": 2058`

https://dmoj.ca/api/v2/contest/mockccc15s
Expected: `"old_rating": 2058, "new_rating": 1987`
What it responds: `"old_rating": 1976, "new_rating": 1987`

PR #1743, opened 2021-08-04
Add consistency to participation in /contest

JoshuaTianYangLiu:
TODOs
`/contest`
- [X] old_volatility
- [X] new_volatility
- [ ] ~virtual_participation_number~ (Always 0)

`/participations`
- [ ] old_rating
- [ ] new_rating
- [ ] old_volatility
- [ ] new_volatility
- [ ] solutions ? (Out of context if placed in api)

`/user`
- [ ] user ? (Redundant)
- [ ] start_time
- [ ] end_time
- [ ] tiebreaker
- [ ] old_rating
- [ ] new_rating
- [ ] is_disqualified
- [ ] solutions
- [ ] old_volatility
- [ ] new_volatility
- [ ] virtual_participation_number (Always 0)

Also, a fix needs to be implemented for #1742 or else the same error will propagate to these proposed changes and cause confusion

~Perhaps increase mccc15s by a microsecond~

### DIFF
--- a/judge/views/api/api_v2.py
+++ b/judge/views/api/api_v2.py
@@ -265,6 +265,8 @@ class APIContestDetail(APIDetailView):
                 username=F('user__user__username'),
                 old_rating=Subquery(old_ratings_subquery.values('rating')[:1]),
                 new_rating=Subquery(new_ratings_subquery.values('rating')[:1]),
+                old_volatility=Subquery(old_ratings_subquery.values('volatility')[:1]),
+                new_volatility=Subquery(new_ratings_subquery.values('volatility')[:1]),
             )
             .order_by('-score', 'cumtime', 'tiebreaker')
         )
@@ -318,6 +320,8 @@ class APIContestDetail(APIDetailView):
                     'tiebreaker': participation.tiebreaker,
                     'old_rating': participation.old_rating,
                     'new_rating': participation.new_rating,
+                    'old_volatility': participation.old_volatility,
+                    'new_volatility': participation.new_volatility,
                     'is_disqualified': participation.is_disqualified,
                     'solutions': contest.format.get_problem_breakdown(participation, problems),
                 } for participation in participations


judge/views/api/api_v2.py
```diff
@@ -527,6 +552,13 @@ def get_object_data(self, profile):
...
+        new_ratings_subquery = Rating.objects.filter(participation=OuterRef('pk'))
+        old_ratings_subquery = (
+            Rating.objects
+            .filter(user=OuterRef('user__pk'), contest__end_time__lt=OuterRef('contest__end_time'))
+            .order_by('-contest__end_time')
+        )
+
         contest_history = []
         participations = (
             ContestParticipation.objects
```
Ninjaclasher, line 564: Do we want to `select_related('contest')`?

judge/views/api/api_v2.py
```diff
@@ -536,17 +568,27 @@ def get_object_data(self, profile):
                 contest__in=Contest.get_visible_contests(self.request.user),
                 contest__end_time__lt=self._now,
             )
+            .annotate(
+                old_rating=Subquery(old_ratings_subquery.values('rating')[:1]),
+                new_rating=Subquery(new_ratings_subquery.values('rating')[:1]),
```
Ninjaclasher, line 573: Not sure if this gets optimized properly, but why not just use `rating__rating` and `rating__volatility` instead of a subquery here?

judge/views/api/api_v2.py
```diff
@@ -352,9 +356,22 @@ def get_unfiltered_queryset(self):
...
 
+        new_ratings_subquery = Rating.objects.filter(participation=OuterRef('pk'))
+        old_ratings_subquery = (
+            Rating.objects
+            .filter(user=OuterRef('user__pk'), contest__end_time__lt=OuterRef('contest__end_time'))
+            .order_by('-contest__end_time')
+        )
+
         return (
             ContestParticipation.objects
```
Ninjaclasher, line 367: Same comments here as above.

Ninjaclasher (commented):
Can you also create a PR to https://github.com/DMOJ/docs?

judge/views/api/api_v2.py
JoshuaTianYangLiu: If so, should https://github.com/JoshuaTianYangLiu/online-judge/blob/api-consistency/judge/views/api/api_v2.py#L256 be changed as well?

judge/views/api/api_v2.py
JoshuaTianYangLiu: That's true, could the same reasoning be applied to https://github.com/JoshuaTianYangLiu/online-judge/blob/api-consistency/judge/views/api/api_v2.py#L255?

judge/views/api/api_v2.py
Ninjaclasher: Yes, I think it should be changed as well.

judge/views/api/api_v2.py
Ninjaclasher: Yes

Ninjaclasher:
Also, perhaps we should hold off on this PR until #1751 .

Ninjaclasher (approved):
LGTM. @int-y1 when you have time, could you take another look?

PR #1747, opened 2021-08-09
Implement user display name override

### LINKED ISSUE #1745: Allow for user display name overrides
Currently we always show the username of a user everywhere:

https://github.com/DMOJ/online-judge/blob/c189db30f5e734f18d6a9b1287004851118af84c/judge/models/profile.py#L151-L154

We could add a new column in `Profile`, e.g. `username_display_override`, that when present overrides the return value of that function. The use case is pre-registered contests were teams have names which cannot be expressed as usernames, and which we currently butcher.

quantum5:
Fixes #1745.


kiritofeng:
Doesn't override problem author, nor https://github.com/DMOJ/online-judge/blob/master/templates/ticket/message.html#L5

quantum5:
@kiritofeng fixed tickets, not sure what you mean about author, since it appears to work on the problem page.

kiritofeng (approved):
LGTM; I guess the thing about problem authors might've been some caching nonsense.

Ninjaclasher (approved):
Code looks good, but do we want to support filtering by the override in the select2 widget and profile admin?

PR #1751, opened 2021-08-14
Migrate rating system from Topcoder to Elo-MMR

int-y1:
[Elo-MMR](https://arxiv.org/abs/2101.00400) is a rating system that has more desirable properties than Topcoder's rating system. For example, Elo-MMR converges faster, prevents users from [farming volatility](https://codeforces.com/blog/entry/87848), and has a more reasonable performance calculation.

Change in the rating curve (ratings rounded to the nearest 100x+50): ratingcurve

Change in each user's rating (i.e. users can expect their rating to increase by +100): 
ratingdelta

Due to the expected +100 rating increase, I modified the rating cutoffs:
- Old: `RATING_VALUES = [1000, 1200, 1500, 1800, 2200, 3000]`
- New: `RATING_VALUES = [1000, 1300, 1600, 1900, 2400, 3000]`

Change in performance (on my virtual machine):
- It takes 8 seconds to rate all 111 contests using Topcoder.
- It takes **12** seconds to rate all 111 contests using Elo-MMR. The biggest bottlenecks are `recalculate_ratings()` (**4.5** seconds), and the query for `historical_p` (**1.5** seconds).

### DIFF
--- a/judge/ratings.py
+++ b/judge/ratings.py
@@ -8,12 +8,13 @@ from django.db.models.functions import Coalesce
 from django.utils import timezone
 
 
-BETA2 = 320. ** 2
+BETA2 = 328.33 ** 2
 VAR_INIT = 350. ** 2 * (BETA2 / 212**2)
 VAR_PER_CONTEST = 1219.047619 * (BETA2 / 212**2)
 VAR_LIM = ((VAR_PER_CONTEST**2 + 4*BETA2*VAR_PER_CONTEST) ** .5 - VAR_PER_CONTEST) / 2
 MEAN_INIT = 1500.
-VALID_RANGE = MEAN_INIT - 20*VAR_INIT, MEAN_INIT + 20*VAR_INIT
+SD_INIT = VAR_INIT ** .5
+VALID_RANGE = MEAN_INIT - 20*SD_INIT, MEAN_INIT + 20*SD_INIT
 TANH_C = 3 ** .5 / math.pi
 
 
@@ -21,13 +22,9 @@ def eval_tanhs(tanh_terms, x):
     return sum((wt / (TANH_C * sd)) * math.tanh((x - mu) / (2 * TANH_C * sd)) for mu, sd, wt in tanh_terms)
 
 
-def eval_tanhs_grad(tanh_terms, x):
-    return sum(0.5 * wt * (TANH_C * sd * math.cosh((x - mu) / (2 * TANH_C * sd))) ** -2 for mu, sd, wt in tanh_terms)
-
-
-def solve(tanh_terms, y_tg=0, lin_factor=0, bounds=VALID_RANGE):
+def solve(tanh_terms, y_tg, lin_factor=0, bounds=VALID_RANGE):
     L,R = bounds
-    while R-L > 1e-5*VAR_INIT:
+    while R-L > 1e-5*SD_INIT:
         x = 0.5 * (L + R)
         y = lin_factor * x + eval_tanhs(tanh_terms, x)
         if y > y_tg:
--- a/judge/ratings.py
+++ b/judge/ratings.py
@@ -96,7 +96,6 @@ def recalculate_ratings(ranking, old_mean, times_ranked, historical_p):
                 y_tg -= 1. / (TANH_C * delta[j])
             # Otherwise, this is a tie that counts as half a win, as per Elo-MMR.
         new_p[i] = solve(tanh_terms, y_tg, bounds=bounds)
-        historical_p[i] = [new_p[i]] + historical_p[i]
 
     # Fill all indices between i and j, inclusive. Use the fact that new_p is non-increasing.
     def divconq(i, j):
@@ -120,7 +119,7 @@ def recalculate_ratings(ranking, old_mean, times_ranked, historical_p):
             tanh_terms = []
             w_prev = 1.
             w_sum = 0.
-            for j, h in enumerate(historical_p[i]):
+            for j, h in enumerate([new_p[i]] + historical_p[i]):
                 gamma2 = (VAR_PER_CONTEST if j > 0 else 0)
                 h_var = get_var(times_ranked[i] + 1 - j)
                 k = h_var / (h_var + gamma2)
--- a/judge/ratings.py
+++ b/judge/ratings.py
@@ -9,13 +9,14 @@ from django.utils import timezone
 
 
 BETA2 = 328.33 ** 2
-VAR_INIT = 350. ** 2 * (BETA2 / 212**2)
-VAR_PER_CONTEST = 1219.047619 * (BETA2 / 212**2)
-VAR_LIM = ((VAR_PER_CONTEST ** 2 + 4 * BETA2 * VAR_PER_CONTEST) ** .5 - VAR_PER_CONTEST) / 2
 MEAN_INIT = 1500.
-SD_INIT = VAR_INIT ** .5
+VAR_INIT = 350**2 * (BETA2 / 212**2)
+SD_INIT = math.sqrt(VAR_INIT)
 VALID_RANGE = MEAN_INIT - 20 * SD_INIT, MEAN_INIT + 20 * SD_INIT
-TANH_C = 3 ** .5 / math.pi
+VAR_PER_CONTEST = 1219.047619 * (BETA2 / 212**2)
+VAR_LIM = (math.sqrt(VAR_PER_CONTEST**2 + 4 * BETA2 * VAR_PER_CONTEST) - VAR_PER_CONTEST) / 2
+SD_LIM = math.sqrt(VAR_LIM)
+TANH_C = math.sqrt(3) / math.pi
 
 
 def eval_tanhs(tanh_terms, x):
@@ -67,7 +68,7 @@ def tie_ranker(iterable, key=attrgetter('points')):
         yield rank + (delta - 1) / 2.0
 
 
-def get_var(times_ranked, cache):
+def get_var(times_ranked, cache=[VAR_INIT]):
     while times_ranked >= len(cache):
         next_var = 1. / (1. / (cache[-1] + VAR_PER_CONTEST) + 1. / BETA2)
         cache.append(next_var)
@@ -76,9 +77,8 @@ def get_var(times_ranked, cache):
 
 def recalculate_ratings(ranking, old_mean, times_ranked, historical_p):
     n = len(ranking)
-    cache = [VAR_INIT]
 
-    delta = [(get_var(t, cache) + VAR_PER_CONTEST + BETA2) ** .5 for t in times_ranked]
+    delta = [math.sqrt(get_var(t) + VAR_PER_CONTEST + BETA2) for t in times_ranked]
 
     new_p = [0.] * n
     new_mean = [0.] * n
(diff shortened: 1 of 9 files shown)


judge/ratings.py
```diff
@@ -28,94 +67,77 @@ def tie_ranker(iterable, key=attrgetter('points')):
...
-
-
-def normal_CDF_inverse(p):
-    assert 0.0 < p < 1
-
-    # See article above for explanation of this section.
-    if p < 0.5:
-        # F^-1(p) = - G^-1(p)
-        return -rational_approximation(math.sqrt(-2.0 * math.log(p)))
+def get_var(times_ranked, cache):
```
quantum5, line 70: Here's a trick: you can hide the cache as a default argument and it'll last for the duration of execution, i.e. do
```suggestion
def get_var(times_ranked, cache=[VAR_INIT]):
```

judge/ratings.py
```diff
@@ -28,94 +67,77 @@ def tie_ranker(iterable, key=attrgetter('points')):
...
+        y_tg = 0
+        for j, s in enumerate(ranking):
+            tanh_terms.append((old_mean[j], delta[j] * TANH_C, 1))
+            if s > r:       # j loses to i
+                y_tg += 1. / (TANH_C * delta[j])
+            elif s < r:     # j beats i
+                y_tg -= 1. / (TANH_C * delta[j])
+            # Otherwise, this is a tie that counts as half a win, as per Elo-MMR.
+        new_p[i] = solve(tanh_terms, y_tg, bounds=bounds)
+        historical_p[i] = [new_p[i]] + historical_p[i]
```
quantum5, line 99: Might be worth considering a `deque` if prepends happen a lot. Although `deque` has its own overhead, so it's probably worth benchmarking.

judge/ratings.py
```diff
@@ -28,94 +67,77 @@ def tie_ranker(iterable, key=attrgetter('points')):
...
+                tanh_terms.append((h, BETA2 ** .5 * TANH_C, w))
+                w_prev = w
+                w_sum += w / BETA2
+            w0 = 1. / get_var(times_ranked[i] + 1, cache) - w_sum
+            p0 = eval_tanhs(tanh_terms[1:], old_mean[i]) / w0 + old_mean[i]
+            new_mean[i] = solve(tanh_terms, w0 * p0, lin_factor=w0)
+
+    # Display a slightly lower rating to incentivize participation.
+    # As times_ranked increases, new_rating converges to new_mean.
+    new_rating = [round(m - (get_var(t + 1, cache) ** .5 - VAR_LIM ** .5)) for m, t in zip(new_mean, times_ranked)]
```
quantum5, line 138: Is there a reason you are using `** .5` other than `sqrt` (`from math import sqrt`)?

```console
$ python3 -m timeit -s 'a = 1337' 'a ** .5'
5000000 loops, best of 5: 82.8 nsec per loop
$ python3 -m timeit -s 'a = 1337; from math import sqrt' 'sqrt(a)'
10000000 loops, best of 5: 26.8 nsec per loop
```

Also since `VAR_LIM` is a constant, might as well define it as one?

judge/ratings.py
```diff
@@ -8,6 +8,45 @@
...
 
 
+BETA2 = 328.33 ** 2
+VAR_INIT = 350. ** 2 * (BETA2 / 212**2)
+VAR_PER_CONTEST = 1219.047619 * (BETA2 / 212**2)
+VAR_LIM = ((VAR_PER_CONTEST ** 2 + 4 * BETA2 * VAR_PER_CONTEST) ** .5 - VAR_PER_CONTEST) / 2
+MEAN_INIT = 1500.
+SD_INIT = VAR_INIT ** .5
+VALID_RANGE = MEAN_INIT - 20 * SD_INIT, MEAN_INIT + 20 * SD_INIT
+TANH_C = 3 ** .5 / math.pi
```
quantum5, line 18: Use `sqrt` instead of `** .5` for better readability? See below.

judge/ratings.py
```diff
@@ -28,94 +67,77 @@ def tie_ranker(iterable, key=attrgetter('points')):
...
+            tanh_terms = []
+            w_prev = 1.
+            w_sum = 0.
+            for j, h in enumerate(historical_p[i]):
+                gamma2 = (VAR_PER_CONTEST if j > 0 else 0)
+                h_var = get_var(times_ranked[i] + 1 - j, cache)
+                k = h_var / (h_var + gamma2)
+                w = w_prev * k**2
+                # Future optimization: If j is around 20, then w < 1e-3 and it is possible to break early.
+                tanh_terms.append((h, BETA2 ** .5 * TANH_C, w))
```
quantum5, line 129: `sqrt`? (see below)

quantum5 (commented):
Otherwise looks pretty good.

judge/ratings.py
fluix-dev: Not exactly sure how timeit works, but I see (consistent) different results. Perhaps in your instance, `timeit` imports `sqrt` every iteration?
```py
>>> timeit.timeit("sqrt(1337)", setup="from math import sqrt", number=1000000)
0.053775883017806336
>>> timeit.timeit("1337**0.5", number=1000000)
0.010119114012923092
```

fluix-dev: Disregard, variable assignment is important.
```py
>>> timeit.timeit("a=1337; sqrt(a)", setup="from math import sqrt", number=1000000)
0.03669788900879212
>>> timeit.timeit("a=1337; a**0.5", number=1000000)
0.07756782200885937
```
```py
>>> timeit.timeit("a=1337; a=sqrt(a)", setup="from math import sqrt", number=1000000)
0.06084884298616089
>>> timeit.timeit("a=1337; a=a**0.5", number=1000000)
0.098016113974154
```

judge/ratings.py
int-y1: Added `sqrt`.

judge/ratings.py
int-y1: Added `sqrt`. I also reordered the constants to improve readability.

judge/ratings.py
```diff
@@ -122,20 +122,20 @@ def divconq(i, j):
...
                 w_sum += w / BETA2
-            w0 = 1. / get_var(times_ranked[i] + 1, cache) - w_sum
+            w0 = 1. / get_var(times_ranked[i] + 1) - w_sum
             p0 = eval_tanhs(tanh_terms[1:], old_mean[i]) / w0 + old_mean[i]
             new_mean[i] = solve(tanh_terms, w0 * p0, lin_factor=w0)
 
     # Display a slightly lower rating to incentivize participation.
     # As times_ranked increases, new_rating converges to new_mean.
-    new_rating = [round(m - (get_var(t + 1, cache) ** .5 - VAR_LIM ** .5)) for m, t in zip(new_mean, times_ranked)]
+    new_rating = [round(m - (math.sqrt(get_var(t + 1)) - SD_LIM)) for m, t in zip(new_mean, times_ranked)]
```
quantum5, line 138: Should really have done `from math import sqrt`:

```console
$ python3 -m timeit -s 'a = 1337; from math import sqrt' 'sqrt(a)'
10000000 loops, best of 5: 26.6 nsec per loop
$ python3 -m timeit -s 'a = 1337; import math' 'math.sqrt(a)'
5000000 loops, best of 5: 46.1 nsec per loop
```

judge/ratings.py
int-y1: Done. This improved the code readability because `solve()` no longer needs to store `cache`.

judge/ratings.py
int-y1: Done. This saved 0.5 seconds, apparently. Rating all contests now takes 13 seconds.

judge/ratings.py
int-y1: I moved this code to where it's needed (see `Make intent of historical_p clearer`).
Each list is involved in 1 prepend, so the `deque` overhead may not be worth it.

judge/ratings.py
```diff
@@ -120,7 +119,7 @@ def divconq(i, j):
             tanh_terms = []
             w_prev = 1.
             w_sum = 0.
-            for j, h in enumerate(historical_p[i]):
+            for j, h in enumerate([new_p[i]] + historical_p[i]):
```
quantum5, line 122: Interesting. If it's for this, it might make sense to use `itertools.chain([new_p[i]], historical_p[i])`.

int-y1: `chain` didn't have an impact on performance. Should I use `chain` anyway?

quantum5: Supposedly it's better as the lists grow larger, so yea.

int-y1: Benchmarking shows that `chain` is slower.
```
$ python3 -m timeit -s 'a=.1;b=[.2,.3,.4]*300' 'sum([a]+b)'
100000 loops, best of 3: 7.48 usec per loop
$ python3 -m timeit -s 'a=.1;b=[.2,.3,.4]*300;from itertools import chain' 'sum(chain([a],b))'
100000 loops, best of 3: 9.37 usec per loop
```

quantum5: I see, then just stick with concatenation.

Ninjaclasher (commented):
Please also create a PR to https://github.com/DMOJ/docs/blob/master/docs/site/api.md with any API changes.

quantum5 (commented):
Code looks fine to me. Has anyone tested the migration? If it works I am going to approve.

judge/ratings.py
```diff
@@ -8,6 +8,46 @@
...
+        return L
+    if Ry is None:
+        Ry = lin_factor * R + eval_tanhs(tanh_terms, R)
+    if y_tg >= Ry:
+        return R
+    ratio = (y_tg - Ly) / (Ry - Ly)
+    return L * (1 - ratio) + R * ratio
+
+
 def tie_ranker(iterable, key=attrgetter('points')):
```
Ninjaclasher, line 51: Nit: could we move this function above or below the functions of computing ratings? Right now, it's in the middle (unless there's a reason for doing this that I missed).

judge/ratings.py
```diff
@@ -125,13 +145,11 @@ def rate_contest(contest):
     rating_sorted = rating_subquery.order_by('-contest__end_time')
     users = contest.users.order_by('is_disqualified', '-score', 'cumtime', 'tiebreaker') \
         .annotate(submissions=Count('submission'),
-                  last_rating=Coalesce(Subquery(rating_sorted.values('rating')[:1]), 1200),
```
Ninjaclasher, line 128: I think we still need to keep this for rating floors and ceilings.

Ninjaclasher: (and if we do keep it, we might want to make `1200` a constant like `MEAN_INIT`).

int-y1: Now that I think about it, `last_rating__lt` might cause an error. I will fix this soon.

int-y1: Fixed, rating floors and ceilings no longer cause an error.

judge/ratings.py
int-y1: I moved this function upwards.

judge/ratings.py
```diff
@@ -125,13 +145,13 @@ def rate_contest(contest):
...
-                  volatility=Coalesce(Subquery(rating_sorted.values('volatility')[:1]), 535),
+                  last_rating=Coalesce(Subquery(rating_sorted.values('rating')[:1]), MEAN_INIT),
```
Ninjaclasher, line 148: What I meant was to set the `1200` as a constant (i.e. `RATING_INIT = 1200`), since right now having a default of `MEAN_INIT` doesn't make sense. Also, we should probably double check that `1200` is a reasonable default rating.

int-y1: Done. 1200 is a reasonable value because in the Elo-MMR rating curve, the median is 1211 and the mean is 1279.

Xyene (approved):
LGTM and applied to prod.

Thanks everyone who worked on this!

PR #1755, opened 2021-08-20
Fix `li` tag in contest short form display

Ninjaclasher:
When it renders correctly with the CSS and it was only caught in prod because the CSS wasn't rebuilt...

(Maybe we should include some kind of HTML + Jinja linter in the build)


kiritofeng (approved):
Thanks!

PR #1762, opened 2021-09-04
Implement hardcore 2FA mode

### LINKED ISSUE #1761: Add hardcore 2FA mode
Due to the potential of bypassing 2FA with social engineering, some installs may opt to reject all 2FA lockout requests. To make this absolutely clear, there should be the option to add a bolded warning on the scratch codes page, reminding the user that losing both the 2FA device and scratch codes means the account is permanently lost and irrecoverable.

This means adding a `settings.py` option `DMOJ_2FA_HARDCORE`, defaulting to `False`.

quantum5:
Example UI:

2FA enable

Also, the message on the 2FA page prompting to contact admins when losing both scratch codes and second factor is removed in this mode.

Fixes #1761.

### DIFF
--- a/dmoj/settings.py
+++ b/dmoj/settings.py
@@ -36,6 +36,8 @@ SITE_LONG_NAME = 'DMOJ: Modern Online Judge'
 SITE_ADMIN_EMAIL = False
 
 DMOJ_REQUIRE_STAFF_2FA = True
+# Display warnings that admins will not perform 2FA recovery.
+DMOJ_2FA_HARDCORE = False
 
 # Set to 1 to use HTTPS if request was made to https://
 # Set to 2 to always use HTTPS for links
--- a/judge/views/two_factor.py
+++ b/judge/views/two_factor.py
@@ -78,6 +78,7 @@ class TOTPEnableView(TOTPView):
         context['scratch_codes'] = [] if self.is_refresh else json.loads(self.profile.scratch_codes)
         context['qr_code'] = self.render_qr_code(self.request.user.username, context['totp_key'])
         context['is_refresh'] = self.is_refresh
+        context['is_hardcore'] = settings.DMOJ_2FA_HARDCORE
         return context
 
     def form_valid(self, form):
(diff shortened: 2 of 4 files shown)


templates/registration/totp_enable.html
```diff
@@ -103,6 +103,15 @@
                         {{ _('These codes will never be shown again after you enable two-factor authentication.') }}</label>
                     <pre><code><span class="fullwidth">{{ scratch_codes|join('\n') }}</span></code></pre>
                 </div>
+                {% if hardcore %}
+                    <div class="alert alert-danger">
+                        <b>{{ _('Important') }}</b>:
+                        {% trans trimmed site_name=SITE_NAME %}
+                            If you lose both your phone and scratch codes, {{ site_name }} admins will NOT be able to recover your account.
```
WallE256, line 110: Maybe change the wording of "phone and scratch codes" to be similar to the one in `two_factor_auth.html`.

quantum5: Fixed.

dmoj/settings.py
```diff
@@ -36,6 +36,7 @@
 SITE_ADMIN_EMAIL = False
 
 DMOJ_REQUIRE_STAFF_2FA = True
+DMOJ_2FA_HARDCORE = False
```
Xyene, line 39: Let's add a comment for what this here

judge/views/two_factor.py
```diff
@@ -78,6 +78,7 @@ def get_context_data(self, **kwargs):
         context['scratch_codes'] = [] if self.is_refresh else json.loads(self.profile.scratch_codes)
         context['qr_code'] = self.render_qr_code(self.request.user.username, context['totp_key'])
         context['is_refresh'] = self.is_refresh
+        context['hardcore'] = settings.DMOJ_2FA_HARDCORE
```
Xyene, line 81: is_hardcore to match is_refresh?

Xyene (approved):
otherwise LGTM

dmoj/settings.py
quantum5: Done.

judge/views/two_factor.py
quantum5: Done.

PR #1764, opened 2021-09-08
add black and flake8-black to the linter

Riolku:
Also run black in the root, automatically formatting all files


Riolku:
Well that's awkward

Riolku:
This conflicts with `flake8-commas` drastically. What should we do?

Riolku:
I guess the question is whether we prefer the auto-formatting that Black provides us with, or we use flake8-commas to have slightly better diffs? Kinda surprising, given that Black advertises small diff sizes. Maybe we can find a different auto-formatter?

Riolku:
Updated to remove flake8-quotes. Won't be able to pass unit-tests in this state, unless the PR was split into two: one to remove the quote check, and another to add black.

dmoj/celery.py
```diff
@@ -23,5 +24,4 @@
 
 @task_failure.connect()
 def celery_failure_log(sender, task_id, exception, traceback, *args, **kwargs):
-    logger.error('Celery Task %s: %s on %s', sender.name, task_id, socket.gethostname(),  # noqa: G201
-                 exc_info=(type(exception), exception, traceback))
+    logger.exception('Celery Task %s: %s on %s', sender.name, task_id, socket.gethostname())  # noqa: G201
```
quantum5, line 27: Pretty sure this is wrong since it discards the exception we are given. Please revert and put the `# noqa` comment in the correct place.

Riolku:
The comment about linting seems to just... not be true. Flake8-commas is the proper package to remove. There was a pending error with logging-format, presumably only raised by black now.

dmoj/celery.py
Riolku: this was raised by `flake8-logging-format`

Riolku:
```(dmojsite) riolku@midas:~/workspace/site/site$ flake8
./dmoj/celery.py:27:5: G201 Logging: .exception(...) should be used instead of .error(..., exc_info=True)
(dmojsite) riolku@midas:~/workspace/site/site$ flake8
./dmoj/celery.py:27:5: G202 Logging statement has redundant exc_info
```

Riolku: Oh. I see. Let me fix that.

Riolku: Had to move `# noqa` around to avoid the linter error, because of black's formatting.

Riolku:
Closing in favour of #1766

PR #1765, opened 2021-09-08
Fix join/spectate button in contest listing

### LINKED ISSUE #544: Contest listing displays "Join" where it should say "Spectate"
Contest listing displays "Join" even after finishing contest.

Clicking on the "Join" button results in an alert.

int-y1:
Fixes #544.

Fix the contest listing button for normal users (i.e. user not involved in making the contest). The behaviour is now:

- Didn't start contest: `Join`
- In contest: (column is omitted)
- In contest and left, but the timer didn't count to 0 yet: `Join`
- Timer counted to 0: `Spectate`
- While spectating: (column is omitted)
- Done spectating: `Spectate`

### DIFF
--- a/templates/contest/list.html
+++ b/templates/contest/list.html
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() or request.profile in contest.testers.all() %}
+            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() or request.profile in contest.testers.all() or contest.has_completed_contest(request.user) %}
                 <form action="{{ url('contest_join', contest.key) }}" method="post">
                     {% csrf_token %}
                     <input type="submit" class="unselectable button full participate-button"
--- a/templates/contest/list.html
+++ b/templates/contest/list.html
@@ -121,10 +121,10 @@
     {% endif %}
 {% endmacro %}
 
-{% macro contest_join(contest, request) %}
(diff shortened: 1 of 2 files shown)


templates/contest/list.html
```diff
@@ -124,7 +124,7 @@
 {% macro contest_join(contest, request) %}
     {% if not request.in_contest %}
         <td>
-            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() or request.profile in contest.testers.all() %}
+            {% if request.profile in contest.authors.all() or request.profile in contest.curators.all() or request.profile in contest.testers.all() or contest.has_completed_contest(request.user) %}
```
quantum5, line 127: This runs O(n) queries, where `n` is the number of active contests. I guess it's not that bad.

Riolku: I suppose the other option is to run a few SQL queries at the beginning so that it's just one SQL query that produces O(n) results. More complex obviously.

Ninjaclasher: The way participation end time is calculated is quite complex to do in a batch query (https://github.com/DMOJ/online-judge/blob/master/judge/models/contest.py#L471), so I think doing O(n) queries is the best option we have.

Riolku: Like Quantum said, it's probably not too bad. If the query is complex and the number of active contests is expected to be low, it's not worth changing this.

int-y1: > The way participation end time is calculated is quite complex to do in a batch query

`contests.py` already calculates `participation.ended` O(n) times. I pushed a commit that uses this calculation. There's no need to optimize `participation.ended`.

quantum5 (approved):

PR #1769, opened 2021-09-22
Return a plaintext 404 response for comment vote view

Ninjaclasher:
Fixes 
image

Credits to @int-y1 for finding this.


kiritofeng (approved):
LGTM

PR #1772, opened 2021-09-25
Implement dumb full text search for CJK

quantum5:
This works around MariaDB's inability to tokenize CJK text, which would
force the user to search whole sentences.


kiritofeng (approved):
LGTM

PR #1775, opened 2021-09-27
Allow <details> and <summary> tags

quantum5:
Currently patched on prod.


kiritofeng (approved):
Well it's patched into prod...

PR #1778, opened 2021-10-01
Allow joining a new contest while a participation is already active

### LINKED ISSUE #1749: Allow joining a new contest while a participation is already active
This occasionally causes contestant confusion. We can just pop up an extra alert if they're already in a contest upon clicking _Join_, then move to the new participation.

Ninjaclasher:
Fixes #1749.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -330,9 +330,6 @@ class ContestJoin(LoginRequiredMixin, ContestMixin, BaseDetailView):
                                    _('"%s" is not currently ongoing.') % contest.name)
 
         profile = request.profile
-        if profile.current_contest is not None:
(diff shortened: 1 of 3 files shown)


quantum5 (commented):
Code looks fine, but what if the user joins from the contest list page?

Ninjaclasher:
When a user is in contest, the third column on the contest list page is hidden, so it's not possible for them to join from that page. I'm not sure if we want to keep this behaviour or if we want to show the third column unconditionally.

quantum5 (approved):
This looks fine as is, I think. Users could always join from the contest page, and joining a new contest while already in a contest is not really something we want to encourage. If it does become a problem we can just write a few more lines of code.

PR #1781, opened 2021-10-02
Remove CachedContestCalendar view

quantum5:
This view is not used and can never be used due to the existence of
private contests. Therefore it should be removed.


kiritofeng (approved):
LGTM

Issue #1649, opened 2021-04-20
Upgrade to Django 3.2 LTS


kiritofeng:
Security support for Django 2.2 ends April 2022, we should upgrade to Django 3.2 LTS before then.

int-y1:
Possibly blocked on #961. See https://github.com/DMOJ/online-judge/pull/1811#issuecomment-1156055058.

Xyene:
This is complete.

PR #1783, opened 2021-10-02
Remove dependency on django.utils.six

quantum5:
This is done to prepare for upgrade to Django 3.x, which drops six.

Refs #1649


kiritofeng (approved):
LGTM

PR #1784, opened 2021-10-02
Replace force_text with force_str

quantum5:
These two are aliases on Python 3.x and Django 3.x is deprecating the `force_text`.


kiritofeng (approved):
LGTM

PR #1785, opened 2021-10-02
Explicitly exclude current user from rank by point query

### LINKED ISSUE #1460: User ranking does not rank users correctly
There are two (possibly due to the same issue) bugs:
 1. The "rank by points" field on the user about page and the user list page are different. Example: [List](https://dmoj.ca/users/) vs [crackersamdjam About](https://dmoj.ca/user/crackersamdjam).
image
image

 2. Two users have the same "rank by points". Example: [JoshuaL](https://dmoj.ca/user/JoshuaL) vs [Devine](https://dmoj.ca/user/Devine). Note that although they both look like they have the same amount of points, JoshuaL has 534.96 points while Devine has 534.58 points.
image
image

Ninjaclasher:
Fixes #1460.

### DIFF
--- a/judge/views/user.py
+++ b/judge/views/user.py
@@ -110,7 +110,7 @@ class UserPage(TitleMixin, UserMixin, DetailView):
 
         context['rank'] = Profile.objects.filter(
             is_unlisted=False, performance_points__gt=self.object.performance_points,
-        ).count() + 1
+        ).exclude(id=self.object.id).count() + 1
 
         if rating:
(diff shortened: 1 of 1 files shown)


quantum5 (commented):
Why would this be necessary? `performance_points__gt=self.object.performance_points` should exclude the current user already?

Ninjaclasher:
According to https://github.com/DMOJ/online-judge/issues/1460#issuecomment-832918397, it seems that the current user is being included even though we use `>` due to precision issues.

PR #1790, opened 2021-10-03
Display rating level names as hover text 

quantum5:
Also made the level names translatable.


judge/ratings.py
```diff
@@ -6,6 +6,7 @@
 from django.db.models import Count, OuterRef, Subquery
 from django.db.models.functions import Coalesce
 from django.utils import timezone
+from django.utils.translation import gettext_lazy
```
quantum5, line 9: Would use `as  _` here, except flake8 freaks out over a `for _` below.

PR #1792, opened 2021-10-04
Copy supported languages from judge-server to README

int-y1:
Changes to the README:
- Whitespace cleanup
- Copy+paste supported languages from judge-server. This mostly affects Java.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,11 +105,12 @@ The judge can also grade in the languages listed below. These languages are less
 * Forth
 * Go
 * Groovy
+* GAS x86/x64/ARM
 * Haskell
 * INTERCAL
 * Kotlin
 * Lua
-* NASM
+* NASM x86/x64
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -121,12 +105,11 @@ The judge can also grade in the languages listed below. These languages are less
 * Forth
 * Go
 * Groovy
-* GAS x86/x64/ARM
```
quantum5, line 124: Interesting that GAS isn't listed. This is probably an oversight on the judge side.

quantum5 (commented):
Otherwise seems okay.

README.md
int-y1: I agree, `GAS x86/x64/ARM` and `NASM x86/x64` seem to be more correct.

quantum5 (approved):

PR #1793, opened 2021-10-05
Add checkmark to hide closed tickets

### LINKED ISSUE #1610: Add option to filter /tickets/ by open...
...so admins know which issues still need to be addressed

int-y1:
Closes #1610. Here is a screenshot of this PR. Note that `dummy 2` is closed, and doesn't show up.

Capture

There is weird spacing because the notification checkbox is not in the `<form>`. This spacing issue is irrelevant to this PR.

### DIFF
--- a/judge/views/ticket.py
+++ b/judge/views/ticket.py
@@ -236,6 +236,8 @@ class TicketList(LoginRequiredMixin, ListView):
 
     def get_queryset(self):
         queryset = self._get_queryset()
+        if self.GET_with_session('hide'):
+            queryset = queryset.filter(is_open=True)
         if self.GET_with_session('own'):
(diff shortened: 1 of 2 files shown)


judge/views/ticket.py
```diff
@@ -236,6 +236,8 @@ def _get_queryset(self):
 
     def get_queryset(self):
         queryset = self._get_queryset()
+        if self.GET_with_session('hide'):
```
Ninjaclasher, line 239: `hide` isn't very descriptive in this context, perhaps `open`?

int-y1: Fixed. I think `open` is a better word.

quantum5 (approved):

PR #1801, opened 2021-10-14
Remove raw problem view

kiritofeng:


int-y1:
Other places in the codebase use `problem/raw.html`

Issue #1776, opened 2021-09-27
Preventing duplicate form submissions


leduythuccs:
According to https://github.com/DMOJ/online-judge/issues/565 & https://github.com/DMOJ/online-judge/commit/0e880ccc26c75b50ddbd835472c5435bf4e6a9d4, users will not able to submit a form more than once time USING the submit button.

However, if users use enter button to submit the form (for example: registration form, edit organization form), they can submit multiple forms if they press the enter fast enough.

This is a really big thing for the registration form because:
- If the user is registering for a new account and they press the enter 2 time
- Both forms will pass the check for the unique username (because the site haven't added that username into the db yet)
- The first form will create the account peacefully 
- The second form will raise an internal error because of a duplicate username (instead of a form error) -> This will show the user the "Site died" view, which is very confusing for the user (because the account is created successfully).

This could be solved easily by overriding the submit method. I could submit a PR.

PR #1809, opened 2021-10-19
Properly prevent multiple form submissions

leduythuccs:
Due to my bad in #1777, the Random Problem button was not able to click more than once time.

This PR comes with another way to prevent multiple form submissions. Instead of disabling the submit function, it disables submit buttons (button & input with type=submit). This also prevents multiple submissions via pressing the enter key (as I mentioned in #1776)

Because random button using `a` to submit the form so it will be fine.

### DIFF
--- a/resources/common.js
+++ b/resources/common.js
@@ -247,10 +247,7 @@ $(function () {
 
     $('form').submit(function (evt) {
         // Prevent multiple submissions of forms, see #565, #1776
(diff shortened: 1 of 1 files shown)


quantum5:
Because pressing enter triggers the submit event on the form directly? The proper solution would be to either change the random problem button, or even better, add a CSS class that exempts the form.

PR #1814, opened 2021-11-15
Use different colour for the first letter of target

quantum5:
This should help visually distinguish targets from grandmasters.


kiritofeng (approved):
LGTM

PR #1817, opened 2021-11-24
Add space in rating chart

int-y1:
Works in chrome.

Capture

### DIFF
--- a/templates/user/user-about.html
+++ b/templates/user/user-about.html
@@ -276,6 +276,7 @@
                 var getItem = function (index) {
                     return rating_history[index];
                 };
+                var ratings = rating_history.map(function (x) {return x.rating;});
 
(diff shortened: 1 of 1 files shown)


templates/user/user-about.html
```diff
@@ -276,6 +276,7 @@ <h4>{{ _('Rating history') }}</h4>
                 var getItem = function (index) {
                     return rating_history[index];
                 };
+                var ratings = rating_history.map(function (x) {return x.rating;});
```
quantum5, line 279:
```suggestion
                var ratings = rating_history.map(function (x) {
                    return x.rating;
                });
```

quantum5 (approved):

quantum5 (approved):

Issue #1816, opened 2021-11-23
Add a button to check the entered password.


k906506:
In now, user cannot see the entered password In the login page.
Add a visible button so that user can see the entered password.

Diwakar-Gupta:
I have created a button on the bottom of password field.
I wonder if this is fine or a floating button on the right end of password field will be better.

<img width="267" alt="Screenshot 2022-11-04 at 9 45 11 AM" src="https://user-images.githubusercontent.com/39624018/199885828-7de04218-2265-4870-a03c-56d39e62b47d.png">

<img width="272" alt="Screenshot 2022-11-04 at 9 39 17 AM" src="https://user-images.githubusercontent.com/39624018/199884212-e39f30b7-d44d-40e6-a8a3-b9ef8a447306.png">

quantum5:
This is a terrible idea because it's almost impossible to get right. Any mistake could result in the password being stored as plaintext autocomplete, which is a terrible security situation. While it may be possible to coerce all known browsers into not storing the password as plaintext, the autocomplete is just a blackbox from the webapp's perspective and there is no guarantee that future browser versions will work the same way. The only possible secure implementation is one by the browser itself.

PR #1819, opened 2021-11-25
Add visible password button

k906506:
Add visible password button
work on Chrome

ezgif com-gif-maker

### DIFF
--- a/templates/registration/login.html
+++ b/templates/registration/login.html
@@ -1,79 +1,95 @@
 {% extends "base.html" %}
 
 {% block media %}
-    <style>
-        h4 {
-            padding-top: 1em;
-        }
+<style>
+    h4 {
+        padding-top: 1em;
+    }
 
-        .social {
-            display: inline;
-            font-size: 2.3em;
-            float: none;
-        }
+    .social {
+        display: inline;
+        font-size: 2.3em;
+        float: none;
+    }
 
-        .google-icon i {
-            color: #DD4B38;
-        }
+    .google-icon i {
+        color: #DD4B38;
+    }
 
-        .facebook-icon i {
-            color: #133783;
-        }
+    .facebook-icon i {
+        color: #133783;
+    }
 
-        .github-icon i {
-            color: black;
-        }
-    </style>
+    .github-icon i {
+        color: black;
+    }
+</style>
 {% endblock %}
 
 {% block body %}
-    <div class="auth-flow-form">
-        <form action="" method="post" class="form-area">
-            {% csrf_token %}
-            {% if form.errors %}
-                <div id="form-errors">
-                    <p class="error">{{ _('Invalid username or password.') }}</p>
-                </div>
-            {% endif %}
-            <table border="0" style="text-align:left">
-                <tr>
-                    <th><i class="fa fa-user fa-fw"></i>
-                    </th>
-                    <td>{{ form.username }}
-                    </td>
-                </tr>
-                <tr>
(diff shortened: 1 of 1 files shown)


k906506:
@quantum5 May I ask you to review it?

templates/registration/login.html
```diff
@@ -1,79 +1,95 @@
...
+                <td>{{ form.username }}
+                </td>
+            </tr>
+            <tr>
+                <th><i class="fa fa-key fa-fw"></i>
+                </th>
+                <td>{{ form.password }}
+                </td>
+                <td><i class="fa fa-eye fa-lg">
+                    <script>
```
fluix-dev, line 51: Javascript should be placed in the `js_media` block and the `script` tag should have `type="text/javascript"`.

templates/registration/login.html
```diff
@@ -1,79 +1,95 @@
...
+                        $(document).ready(function () {
+                            $('.form-area table tr td i').on('click', function () {
```
fluix-dev, line 53: All of these selectors are unclear. Select based on `id` attributes instead. You'll have to add one for the icon but can use `{{ form.password.auto_id }}` for the password field.

templates/registration/login.html
```diff
@@ -1,79 +1,95 @@
...
+                                $('.form-area table tr:nth-child(2) td:nth-child(2) input').toggleClass('active');
+                                if ($('.form-area table tr:nth-child(2) td:nth-child(2) input').hasClass('active')) {
+                                    $(this).attr('class', "fa fa-eye-slash fa-lg")
```
fluix-dev, line 56: It may be better to use the jQuery's `addClass`, `removeClass`, or `toggleClass` instead of setting all of the classes each time.

templates/registration/login.html
```diff
@@ -1,79 +1,95 @@
...
+                </th>
+                <td>{{ form.username }}
+                </td>
+            </tr>
+            <tr>
+                <th><i class="fa fa-key fa-fw"></i>
+                </th>
+                <td>{{ form.password }}
+                </td>
+                <td><i class="fa fa-eye fa-lg">
```
fluix-dev, line 50: I think this lacks a bit when it comes to accessibility. We should do one of the following:

1. Use `aria-role="button"` alongside an appropriate `aria-label` (something like "Make password visible."). The `aria-pressed` attribute can be toggled too, [see the MDN docs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role) and [the FontAwesome docs](https://fontawesome.com/v4.7/accessibility/).

2. Alternatively, just hiding with `aria-hidden="true"` may be just as acceptable because a screen reader could probably just read the password out loud.

Not sure what's best.

fluix-dev (changes requested):
I don't personally think this is really necessary, but I'll go ahead and give an honest review nonetheless, if only to help you improve :)

> Work on Chrome

Let's test this out on multiple browsers; at least Firefox.

PS: in general, you should wait at least a week or two before pinging contributors.

templates/registration/login.html
k906506: @fluix-dev Thank you for reply even though it was a problematic code. After supplement the problems, conducting various tests, May I send the PR again?

fluix-dev: You can amend your previous commit with:
```
git commit --amend
```
and then force push it:
```
git push -f
```

Though no promises this gets accepted of course.

kiritofeng:
Closing as #1816 is closed.

PR #1820, opened 2021-11-28
Fix problems list not showing all problems a user has access to

Ninjaclasher:
If a problem was organization-private, and the user was marked as
tester/curator/author without being in the organization, the
problem would not show up in the problems list. This commit will
fix this.


judge/views/problem.py
```diff
@@ -354,17 +354,18 @@ def apply_full_text(queryset, query):
         return queryset.search(query, queryset.BOOLEAN).extra(order_by=['-relevance'])
 
     def get_normal_queryset(self):
-        filter = Q(is_public=True)
-        if self.profile is not None:
-            filter |= Q(authors=self.profile)
-            filter |= Q(curators=self.profile)
-            filter |= Q(testers=self.profile)
-        queryset = Problem.objects.filter(filter).select_related('group').defer('description', 'summary')
+        q = Q(is_public=True)
```
kiritofeng, line 357: I would personally prefer if you renamed `q` to `filter`, and changed the current `filter` to something else (`org_filter` perhaps?) Otherwise LGTM

Ninjaclasher: Done

kiritofeng (approved):
LGTM, thanks!

Issue #1138, opened 2019-10-23
Upgrade Mistune to version 3


Xyene:
[Mistune](https://github.com/lepture/mistune/) has a new version that supports more fancy formatting (for instance, alignment on Markdown tables). It is a breaking upgrade, but we should upgrade nonetheless.

Ninjaclasher:
One thing to watch out for: `parse_inline_html` and `parse_inline_html` flags no longer exist. The relevant issue is https://github.com/lepture/mistune/issues/256. We should be careful when upgrading so that we don't break markdown inside HTML. The package author describes a workaround (https://github.com/lepture/mistune/issues/256#issuecomment-731139529) that we can probably use for most cases but we'll need to figure something out for inline elements.

kiritofeng:
Mistune v3 has been out for ~5mo by now, perhaps we should consider that instead of v2.

PR #1825, opened 2021-12-07
Fix mistune

UnknowName01:
DMOJ not work with mistune 2.0


fluix-dev:
Would it be difficult to make DMOJ work with mistune 2.0.0?

Xyene (approved):
Relevant: https://github.com/DMOJ/online-judge/issues/1138

In the meantime LGTM, thanks!

PR #1828, opened 2021-12-10
Add support for classes

quantum5:
This PR adds a new subdivision for some organizations, allowing class-wide contests and organization join request approval to be delegated to the class admins.


Xyene (approved):
LGTM, thanks!

PR #1830, opened 2021-12-15
Infer private booleans; fixes #1829

### LINKED ISSUE #1829: Infer "Private to organizations" problem admin field from length of "Organizations"
This would make the interaction between the two less confusing. "Private to organizations" can be updated at model save time.

kiritofeng:


judge/admin/contest.py
```diff
@@ -182,6 +181,9 @@ def save_model(self, request, obj, form, change):
                 raise PermissionDenied
 
         super().save_model(request, obj, form, change)
+        obj.is_private = bool(form.cleaned_data['private_contestants'])
+        obj.is_organization_private = bool(form.cleaned_data['organizations'])
```
quantum5, line 185: Do this before the super call to avoid the extra save.

judge/admin/problem.py
```diff
@@ -235,9 +235,11 @@ def get_form(self, *args, **kwargs):
 
     def save_model(self, request, obj, form, change):
         super(ProblemAdmin, self).save_model(request, obj, form, change)
+        obj.is_organization_private = bool(form.cleaned_data['organizations'])
```
quantum5, line 238: Same here.

quantum5 (approved):

judge/migrations/0125_infer_private_bools.py
```diff
@@ -0,0 +1,29 @@
...
+from django.db import migrations
+
+
+def sync_private_booleans(apps, schema_editor):
+    Problem = apps.get_model('judge', 'Problem')
+    Contest = apps.get_model('judge', 'Contest')
+    db_alias = schema_editor.connection.alias
+
+    for problem in Problem.objects.using(db_alias).all():
+        problem.is_organization_private = problem.organizations.exists()
```
Ninjaclasher, line 12: Should we do this instead?
 - If `is_organization_private` is false, remove all organizations
 - If `is_organization_private` is true and there are no organizations, set `is_organization_private` to false

There are quite a few problems that have `is_organization_private` set to false, but have a non-empty `organizations` field. Of the top of my head, most of the WCIPEG migrated problems still have `WCIPEG Migration` in `organizations`. If we keep this code as-is, all of those problems will be set to organization-private, and we need to track down which ones to make public.

quantum5 (approved):

PR #1843, opened 2022-02-21
Full hidden scoreboard & Post-window own scoreboard

Riolku:
This adds support for a fully-hidden scoreboard type, i.e., one that is hidden even after the contest finishes.

In such a case, it is expected that it would be nice for users to be able to see their own participation, so the second commit allows users to see their own scoreboard if they have completed the contest.

### DIFF
--- a/judge/views/submission.py
+++ b/judge/views/submission.py
@@ -251,7 +251,11 @@ class SubmissionsListBase(DiggPaginatorMixin, TitleMixin, ListView):
                 contest_queryset = Contest.objects.filter(Q(authors=self.request.profile) |
                                                           Q(curators=self.request.profile) |
                                                           Q(scoreboard_visibility=Contest.SCOREBOARD_VISIBLE) |
-                                                          Q(end_time__lt=timezone.now())).distinct()
+                                                          Q(end_time__lt=timezone.now()), scoreboard_visibility__in=(
+                                                          Contest.SCOREBOARD_AFTER_PARTICIPATION,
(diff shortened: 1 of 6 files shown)


judge/views/submission.py
```diff
@@ -251,7 +251,9 @@ def _get_queryset(self):
                 contest_queryset = Contest.objects.filter(Q(authors=self.request.profile) |
                                                           Q(curators=self.request.profile) |
                                                           Q(scoreboard_visibility=Contest.SCOREBOARD_VISIBLE) |
-                                                          Q(end_time__lt=timezone.now())).distinct()
+                                                          Q(end_time__lt=timezone.now())) \
```
Ninjaclasher, line 254: Doesn't this prevent contest authors/curators from seeing submissions to their own contests? I think the correct query would be something like:
```suggestion
                                                          Q(end_time__lt=timezone.now(), scoreboard_visibility__in=(Contest.SCOREBOARD_AFTER_PARTICIPATION, Contest.SCOREBOARD_AFTER_CONTEST))) \
```
(and remove the `.exclude()`.

Ninjaclasher (commented):
Otherwise LGTM.

judge/views/submission.py
Riolku: Ah, you're right. It might be worth noting that this query doesn't handle AFTER_PARTICIPATION properly.

Ninjaclasher (approved):

quantum5 (approved):

Issue #1445, opened 2020-06-12
Problems list ordering isn't consistent


Ninjaclasher:
### Reproduction
Go to a problems list link with a custom sort such as https://dmoj.ca/problems/?order=-points

### Expected behaviour
The ordering of the problems does not change when refreshing the page.

### Actual behaviour
The ordering of the problems is random when refreshing the page.

----

This is usually not a problem, but it can become confusing when going between pages and the same problem shows up twice. Potential fix: add `id` or `code` as a secondary order.

PR #1844, opened 2022-02-21
Add editorial column and filtering

Riolku:
The problem page now has an editorial column. The search form can now be 
used to only show problems with an editorial.

The only extraneous change is adding a method to `Problem`, 
`has_public_editorial`.

Here's some screenshots of the UI I get locally:

Editorial Column, plus sorting:

image

Filtering:

image

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -10,13 +10,13 @@ from django.conf import settings
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.db import transaction
-from django.db.models import Count, F, Prefetch, Q
+from django.db.models import Case, Count, F, IntegerField, Prefetch, Q, Sum, When
 from django.db.utils import ProgrammingError
 from django.http import Http404, HttpResponse, HttpResponseForbidden, HttpResponseRedirect
 from django.shortcuts import get_object_or_404
 from django.template.loader import get_template
 from django.urls import reverse
-from django.utils import translation
+from django.utils import timezone, translation
 from django.utils.functional import cached_property
 from django.utils.html import escape, format_html
 from django.utils.safestring import mark_safe
@@ -370,6 +370,13 @@ class ProblemList(QueryStringSortMixin, TitleMixin, SolvedProblemMixin, ListView
                                         .values_list('problem__id', flat=True))
         if self.show_types:
             queryset = queryset.prefetch_related('types')
+        queryset = queryset.annotate(has_public_editorial=Sum(Case(
+            When(solution__is_public=True, solution__publish_on__lte=timezone.now(), then=1),
+            default=0,
+            output_field=IntegerField(),
+        )))
+        if self.has_public_editorial:
+            queryset = queryset.filter(solution__is_public=True, solution__publish_on__lte=timezone.now())
         if self.category is not None:
             queryset = queryset.filter(group__id=self.category)
         if self.selected_types:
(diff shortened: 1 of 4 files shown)


Riolku:
image

Fixed casing thanks to Xyene, refer to updated image for the casing on the sidebar.

judge/views/problem.py
```diff
@@ -309,6 +309,12 @@ def _solved_sort_order(problem):
 
                     queryset = list(queryset)
                     queryset.sort(key=_solved_sort_order, reverse=self.order.startswith('-'))
+            elif sort_key == 'editorial':
+                def _editorial_sort_order(problem):
+                    return int(problem.has_public_editorial())
```
Ninjaclasher, line 314: This will do O(# of problems) database queries. I would recommend doing something similar to https://github.com/DMOJ/online-judge/blob/master/judge/views/contests.py#L240 and doing the ordering with `order_by`. Also, please add a secondary order so #1445 doesn't happen.

Also, I'm not too sure the usefulness of allowing sorting by a boolean when we also add a toggle.

templates/problem/list.html
```diff
@@ -287,6 +290,15 @@ <h3>{{ _('Hot problems') }} <i class="fa fa-fire"></i></h3>
                         {% if not request.in_contest %}
                             <td class="ac-rate">{{ problem.ac_rate|floatformat(1) }}%</td>
                         {% endif %}
+                        {% if problem.has_public_editorial() %}
```
Ninjaclasher, line 293: This will do O(pagination size) number of queries. See the above comment for a possible solution. Also, what happens when the user is in contest?

judge/views/problem.py
Riolku: I wasn't sure about this either, but one can sort by solved problems. What do we want to do?

Riolku:
image
image

Here are screenshots showing what happens when other editorial situations arise.

judge/views/problem.py
Riolku: Removed sorting entirely.

Riolku:
Sorting was removed entirely. `has_public_editorial`, as a method, was also removed, because it is no longer called.

templates/problem/list.html
```diff
@@ -287,6 +290,15 @@ <h3>{{ _('Hot problems') }} <i class="fa fa-fire"></i></h3>
                         {% if not request.in_contest %}
                             <td class="ac-rate">{{ problem.ac_rate|floatformat(1) }}%</td>
                         {% endif %}
+                        {% if problem.has_public_editorial and not request.in_contest %}
```
Ninjaclasher, line 293: I think we should hide this column outright in contest (similar to AC rate).

Riolku: Yeah that was my intent, I'm just bad at logic.

quantum5 (dismissed):

templates/problem/list.html
```diff
@@ -286,6 +289,15 @@ <h3>{{ _('Hot problems') }} <i class="fa fa-fire"></i></h3>
                         <td class="p">{{ problem.points|floatformat }}{% if problem.partial %}p{% endif %}</td>
                         {% if not request.in_contest %}
                             <td class="ac-rate">{{ problem.ac_rate|floatformat(1) }}%</td>
+                            {% if problem.has_public_editorial %}
+                            <td editorial="1">
```
Xyene, line 293: Nit: indent

templates/problem/list.html
```diff
@@ -232,6 +232,9 @@ <h3>{{ _('Hot problems') }} <i class="fa fa-fire"></i></h3>
                         <th class="ac-rate">
                             <a href="{{ sort_links.ac_rate }}">{{ _('AC %%') }}{{ sort_order.ac_rate }}</a>
                         </th>
+                        <th class="editorial">
+                            {{ _('Editorial') }}
```
Xyene, line 236: Doesn't this create the column even when in-contest?

Riolku: No, it's inside an `else`

quantum5 (approved):

resources/problem.scss
```diff
@@ -151,6 +157,14 @@ ul.problem-list {
     color: orange;
 }
 
+.has-editorial-color {
```
Xyene, line 160: Can avoid the extra class since you're already tagging with an attribute

```suggestion
td[editorial=1] i {
```

Xyene: But w/e, the rest of the code is written like this already so sure

Xyene (approved):

PR #1850, opened 2022-02-22
fix submission visibility in contests

Riolku:
After #1843, because of a misplaced bracket, users can't see submissions from contests with visible scoreboards.


kiritofeng (approved):
LGTM

Issue #1744, opened 2021-08-09
Allow for organization-private, public scoreboard contests


Xyene:
This would enable us to run public-scoreboard contests that require ahead-of-time, out-of-band registration without having to disseminate access codes.

Proposal: turn `is_organization_private : bool` into `visibility : Visible_to_all | Visible_to_organizations | Visible_to_organization_public_scoreboard`. `is_organization_private = true` would then map to `Visible_to_organizations`, and `false` to `Visible_to_all`. `Visible_to_organization_public_scoreboard` would be the proposed new mode.

PR #1854, opened 2022-02-23
Can join organizations

Riolku:
This allows contests to limit which organizations can join a contest.
Closes https://github.com/DMOJ/online-judge/issues/1744.

This also renames `can_join` to `started` in order to avoid confusion,
and implements methods for checking if a contest can be joined by a
user.

Also changes `test_contest` so that `normal` and `normal_open_org` are 
two different users.

Here are UI screenshots of the "cannot join" behaviour:

contest page
contest page hover
contest list
contest list hover


Riolku:
CI seems to be broken, see #1855. The PR can still be reviewed for now: tests pass locally, so hopefully they will pass when CI is fixed.

Riolku:
Lua is now fixed, and tests pass.

Riolku:
@Xyene , since my proposal differs from your proposal in the issue, can I get you to review this, or at least, what you think of my proposal?

I wrote this before realizing what your proposal was, but I feel like having `visible` as a dropdown that changes the meaning of `organizations` is confusing, and then `visibility` is kind of an odd name. 

This PR instead just adds a field `join_organizations`, which specifies which organizations can join the contest (and has an inferred boolean). The implementation of this PR is also simpler, but I'm not sure if this is acceptable, given what you had in mind.

kiritofeng (approved):
This seems fine, just rebase the migration depending on which is pulled first

PR #1857, opened 2022-02-24
add editorial book and sort

Riolku:
Change Editorial column header to a book icon, and implement editorial sorting. Also clean up Sum() clauses for editorial queries, since problems can have at most one editorial.

problems list

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -10,7 +10,7 @@ from django.conf import settings
 from django.contrib.auth.mixins import LoginRequiredMixin, PermissionRequiredMixin
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied
 from django.db import transaction
-from django.db.models import Case, Count, F, IntegerField, Prefetch, Q, Sum, When
+from django.db.models import BooleanField, Case, Count, F, Prefetch, Q, When
 from django.db.utils import ProgrammingError
 from django.http import Http404, HttpResponse, HttpResponseForbidden, HttpResponseRedirect
 from django.shortcuts import get_object_or_404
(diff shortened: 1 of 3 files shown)


templates/problem/list.html
```diff
@@ -233,7 +233,8 @@ <h3>{{ _('Hot problems') }} <i class="fa fa-fire"></i></h3>
                             <a href="{{ sort_links.ac_rate }}">{{ _('AC %%') }}{{ sort_order.ac_rate }}</a>
                         </th>
                         <th class="editorial">
-                            {{ _('Editorial') }}
+                            <a href="{{ sort_links.editorial }}"><i class="fa fa-book"></i>{{ sort_order.editorial }}
```
Ninjaclasher, line 236:
```suggestion
                            <a href="{{ sort_links.editorial }}">
                                <i class="fa fa-book"></i>{{ sort_order.editorial }}
```

judge/views/problem.py
```diff
@@ -294,6 +294,13 @@ def get_paginator(self, queryset, per_page, orphans=0,
                 queryset = queryset.order_by(self.order.replace('name', 'i18n_name'), 'id')
             elif sort_key == 'group':
                 queryset = queryset.order_by(self.order + '__name', 'id')
+            elif sort_key == 'editorial':
+                queryset.annotate(has_public_editorial=Case(
```
Ninjaclasher, line 298: Is this annotation necessary? We already annotate unconditionally in `get_normal_queryset`.

templates/problem/list.html
Riolku: Fixed, and also updated the formatting of the place a few lines above from which I copied it.

judge/views/problem.py
Riolku: Fixed, good catch, thanks.

quantum5 (approved):

PR #1858, opened 2022-02-24
add submission 403 page

### LINKED ISSUE #894: Redesign of unauthorized submission page
Instead of having a SIGSEGV: unauthorized_access page for submissions to problems user has not solved, have a UI that informs user they are unable to view the submission as they have not solved the problem, as well as providing submission's author and problem.

Riolku:
Closes #894. Only shows the problem link if the problem has Solved submission access, and the problem is accessible by the user.

UI:

can see problem
can't see problem

### DIFF
--- a/judge/views/submission.py
+++ b/judge/views/submission.py
@@ -22,10 +22,11 @@ from django.views.generic import DetailView, ListView
 from judge import event_poster as event
 from judge.highlight_code import highlight_code
 from judge.models import Contest, Language, Problem, ProblemTranslation, Profile, Submission
+from judge.models.problem import SubmissionSourceAccess
 from judge.utils.infinite_paginator import InfinitePaginationMixin
 from judge.utils.problems import get_result_data, user_completed_ids, user_editable_ids, user_tester_ids
 from judge.utils.raw_sql import join_sql_subquery, use_straight_join
-from judge.utils.views import DiggPaginatorMixin, TitleMixin
+from judge.utils.views import DiggPaginatorMixin, TitleMixin, generic_message
 
 
 def submission_related(queryset):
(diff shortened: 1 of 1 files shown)


judge/views/submission.py
```diff
@@ -47,9 +53,29 @@ class SubmissionDetailBase(LoginRequiredMixin, TitleMixin, SubmissionMixin, Deta
     def get_object(self, queryset=None):
         submission = super(SubmissionDetailBase, self).get_object(queryset)
         if not submission.can_see_detail(self.request.user):
-            raise PermissionDenied()
+            raise SubmissionPermissionDenied(submission)
         return submission
 
+    def get(self, request, *args, **kwargs):
+        try:
+            return super(SubmissionDetailBase, self).get(request, *args, **kwargs)
```
Ninjaclasher, line 61:
```suggestion
            return super().get(request, *args, **kwargs)
```
This is Python 2 syntax. We're slowly migrating to Python 3 syntax, so any new code should just use `super()`.

Riolku: Alright, pushed.

judge/views/submission.py
```diff
@@ -37,6 +38,11 @@ def submission_related(queryset):
         .prefetch_related('contest_object__authors', 'contest_object__curators')
 
 
+class SubmissionPermissionDenied(Exception):
```
quantum5, line 41: Would it make more sense semantically to inherit from `PermissionDenied`?

quantum5 (commented):
Otherwise LGTM.

judge/views/submission.py
Riolku: Yes, it definitely would, updated.

quantum5 (approved):

PR #1860, opened 2022-02-24
add `view_contest_submissions` field

### LINKED ISSUE #1695: Extend `can_see_scoreboard` to also see in-contest submissions

Riolku:
Overrides and closes #1695. It is likely better to have a dedicated field for this instead of reusing an existing field.


Riolku:
Re-ordered query parameters in submission queryset, unsure if it matters.

judge/models/submission.py
```diff
@@ -152,8 +152,13 @@ def can_see_detail(self, user):
...
             return True
 
-        # If user is an author or curator of the contest the submission was made in
-        if self.contest_object is not None and user.profile.id in self.contest_object.editor_ids:
+        contest = self.contest_object
+        # If user is an author or curator of the contest the submission was made in, or they can see in-contest subs
+        if contest is not None and (
+            user.profile.id in contest.editor_ids or
+            contest.view_contest_submissions.filter(id=user.profile.id).exists() or
+            (user.profile.id in contest.tester_ids and contest.tester_see_submissions)
```
kiritofeng, line 160: Should we switch the order of these two? I think the potential short-circuit should save us time in the majority of cases.

kiritofeng (changes requested):

judge/models/submission.py
Riolku: Definitely. Fixed.

kiritofeng (approved):

judge/models/contest.py
```diff
@@ -91,6 +93,10 @@ class Contest(models.Model):
     view_contest_scoreboard = models.ManyToManyField(Profile, verbose_name=_('view contest scoreboard'), blank=True,
                                                      related_name='view_contest_scoreboard',
                                                      help_text=_('These users will be able to view the scoreboard.'))
+    view_contest_submissions = models.ManyToManyField(Profile, verbose_name=('can see contest submissions'), blank=True,
```
quantum5, line 96: Where is the `_`?

PR #1864, opened 2022-02-24
hide contest links on scoreboard

### LINKED ISSUE #1758: Contest Links Expose Problem Codes to non-contestants
On the contest ranking page, the links to the problems as well as the links in the cells give a 404, but still give away the problem codes used in the contest.

Proposed fixes: remove links for users who can't view the problem (because they aren't in the contest and they don't have view access to the problem).

Riolku:
Closes #1758. Hides problem links on the scoreboard if the problem is inaccessible to the user.

### DIFF
--- a/templates/contest/ranking-table.html
+++ b/templates/contest/ranking-table.html
@@ -54,10 +54,18 @@
 
 {% block before_point_head %}
     {% for problem in problems %}
-        <th class="points header"><a href="{{ url('contest_ranked_submissions', contest.key, problem.problem.code) }}">
-            {{- contest.get_label_for_problem(loop.index0) }}
-            <div class="point-denominator">{{ problem.points }}</div>
-        </a></th>
+        <th class="points header">
+            {# problem is a ContestProblem, so get the real problem #}
+            {% if problem.problem.is_accessible_by(request.user) %}
+                <a href="{{ url('contest_ranked_submissions', contest.key, problem.problem.code) }}">
+                    {{- contest.get_label_for_problem(loop.index0) }}
+                    <div class="point-denominator">{{ problem.points }}</div>
+                </a>
+            {% else %}
+                {{- contest.get_label_for_problem(loop.index0) }}
+                <div class="point-denominator">{{ problem.points }}</div>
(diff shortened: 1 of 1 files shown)


Riolku:
Add a comment to clarify `problem.problem`

templates/contest/ranking-table.html
```diff
@@ -54,10 +54,18 @@
 
 {% block before_point_head %}
     {% for problem in problems %}
-        <th class="points header"><a href="{{ url('contest_ranked_submissions', contest.key, problem.problem.code) }}">
-            {{- contest.get_label_for_problem(loop.index0) }}
-            <div class="point-denominator">{{ problem.points }}</div>
-        </a></th>
+        <th class="points header">
+            {# problem is a ContestProblem, so get the real problem #}
+            {% if problem.problem.is_accessible_by(request.user) %}
```
Ninjaclasher, line 59: I wonder if we should use a `Problem.get_visible_problems` and filter by the contest problems so we do 1 database query instead of O(# of contest problem) queries. Also, we should probably give a similar treatment each participation row since they also have links that lead to 404s.

Riolku: Ugh. Yes, we need to give the same treatment to each row.

Riolku: Unfortunately the code that shows that URL is deeply tangled with, of all things, the ContestFormat code.

Riolku: I think `problems = list(contest.contest_problems.select_related('problem').defer('problem__description').order_by('order'))` the line that queries for the problems already gets the problem object, so we aren't hitting the database.

Riolku: https://github.com/DMOJ/online-judge/blob/master/judge/views/contests.py#L646 here I mean

Riolku: Although we don't get the authors, so we might be in an even worth position.

Riolku: Not sure, comments more than welcome.

PR #1868, opened 2022-02-25
cmd+enter to submit

y0105w49:


Riolku (commented):
Would it be better to have both ctrl+enter and cmd+enter work on mac? Also, how does this code behave on, say, linux?

Riolku (approved):
Seems that the rest of the code is written this way without issue, so whatever.

Ninjaclasher (approved):
LGTM

PR #1870, opened 2022-02-25
add change visibility blog post perm

Riolku:
This allows us to have two groups: users who can draft blog posts, that only admins can public, and of course, admins. It works much the same as problems and contests do.


Riolku:
Prefix perm checks with `judge.`, the code is wrong otherwise. In fact, it silently returns false.

Riolku:
Use a list instead of a tuple in `get_readonly_fields` for better consistency with the rest of the codebase.

judge/admin/interface.py
```diff
@@ -79,11 +79,27 @@ class BlogPostAdmin(VersionAdmin):
     form = BlogPostForm
     date_hierarchy = 'publish_on'
 
+    def has_view_permission(self, request, obj=None):
```
Ninjaclasher, line 82: Is this necessary? None of the other classes override `has_view_permission`.

judge/admin/interface.py
```diff
@@ -79,11 +79,27 @@ class BlogPostAdmin(VersionAdmin):
...
 
+    def get_readonly_fields(self, request, obj=None):
+        if not request.user.has_perm('judge.change_post_visibility'):
+            return ['visible']
+        return []
+
+    def get_queryset(self, request):
+        queryset = BlogPost.objects.all()
+        if not request.user.has_perm('judge.edit_all_post'):
+            queryset = queryset.filter(authors=request.user.profile)
```
Ninjaclasher, line 100:
```suggestion
            queryset = queryset.filter(authors=request.profile)
```

Ninjaclasher (commented):
Otherwise LGTM

judge/admin/interface.py
Riolku: Might as well document this: after a long search, it seems that `get_queryset` is filtered to fetch the model instance, so as long as we write that method write, we don't have to write `has_view_permission`.

judge/admin/interface.py
Riolku: Did not know request had a profile field, thanks.

Riolku:
Alright, updated, thanks.

Riolku:
(forgot to push)

Riolku:
Rename migration name to something more useful.

kiritofeng (approved):
Please rebase migration, probably

PR #1871, opened 2022-02-26
add toggle for testers to see scoreboard

Riolku:
Add a toggle to contests, if true, lets testers see the scoreboard without adding them to the explicit allow list.


kiritofeng (approved):
LGTM

quantum5 (approved):

PR #1876, opened 2022-02-26
fix contest join logic

Riolku:
In particular, users should be able to spectate any contest that is over (even limit_join_organization ones), and nobody should be allowed to join a contest that has ended. Also swap the text on the join buttons.


Riolku:
Closing, want to change how join logic works, fixing visual bug in own PR.

PR #1883, opened 2022-03-05
RFC: Don't strip whitespace from submissions

quantum5:
This could be beneficial to `TEXT` submissions and would have been useful for certain languages that required line numbers to be indented and aligned. However, this is a significant behaviour change and deserves some scrutiny.

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -124,7 +124,7 @@ class DownloadDataForm(Form):
 
 
 class ProblemSubmitForm(ModelForm):
-    source = CharField(max_length=65536, widget=AceWidget(theme='twilight', no_ace_media=True))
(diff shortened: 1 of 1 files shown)


Riolku:
Maybe a silly suggestion, but could we fix a smaller issue and only not strip the trailing NL?

quantum5:
> Maybe a silly suggestion, but could we fix a smaller issue and only not strip the trailing NL?

No. You don't want COBOL users to fail to compile if they insist on writing old style with the line numbers, do you?

kiritofeng (commented):
Are there any problems that rely on this old behaviour? There are some problems with custom graders where we've had to override the behaviour of TEXT so that it includes the trailing newline, but I can't think of any were we want this current behaviour.

PR #1884, opened 2022-03-06
Compare i18n changes with correct branch

int-y1:
I suspect this workflow compares with `master`. This led to meaningless changes such as
https://github.com/DMOJ/online-judge/compare/6b6f3d1..f47a6ab

### DIFF
--- a/.github/workflows/updatemessages.yml
+++ b/.github/workflows/updatemessages.yml
@@ -64,12 +64,13 @@ jobs:
     - name: Remove useless changes
       run: |
         set -eu
+        i18n_head=$(git rev-parse --verify -q origin/update-i18n || echo "HEAD")
         while IFS= read -r -d $'\0' file; do
           if git diff -s --exit-code "$file"; then
             continue
           fi
 
-          if git diff origin/update-i18n -- "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
+          if git diff $(i18n_head) -- "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
             echo "Changed: $file"
           else
             echo "No material change: $file"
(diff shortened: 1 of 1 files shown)


.github/workflows/updatemessages.yml
```diff
@@ -69,7 +69,7 @@ jobs:
             continue
           fi
 
-          if git diff "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
+          if git diff origin/update-i18n -- "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
```
quantum5, line 72: The problem here is what happens after you merge the PR and the branch is deleted? I agree that this approach should solve the problem of daily updates, but better logic is needed to avoid that special case.

int-y1: Done. This should work.
```bash
$ echo $(git rev-parse --verify -q origin/update-i18n || echo "HEAD")
9e610dd9dd5f0c7d302c9939a3f698d95fb7c2d0
$ echo $(git rev-parse --verify -q origin/update-i18nn || echo "HEAD")
HEAD
```

.github/workflows/updatemessages.yml
```diff
@@ -64,12 +64,13 @@ jobs:
...
       run: |
         set -eu
+        i18n_head=$(git rev-parse --verify -q origin/update-i18n || echo "HEAD")
         while IFS= read -r -d $'\0' file; do
           if git diff -s --exit-code "$file"; then
             continue
           fi
 
-          if git diff "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
+          if git diff $i18n_head -- "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
```
quantum5, line 73: Triggered by missing quotes.
```suggestion
          if git diff "$i18n_head" -- "$file" | tail -n +5 2>/dev/null | grep -qP '^[-+](?!"PO-Revision-Date:|"POT-Creation-Date:)'; then
```

quantum5 (approved):
Only one way to find out if this actually works.

PR #1885, opened 2022-03-07
Fetch all branches in GitHub action

int-y1:
`origin/update-i18n` wasn't fetched, so the action went with `i18n_head=HEAD`. To fix this, [fetch all branches](https://github.com/actions/checkout/tree/v2.4.0#fetch-all-history-for-all-tags-and-branches).

### DIFF
--- a/.github/workflows/updatemessages.yml
+++ b/.github/workflows/updatemessages.yml
@@ -8,6 +8,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


Ninjaclasher (approved):
LGTM

quantum5 (approved):
Only one way to find out. I suspect this is not the end of it: it'll probably close the PR if no file has material change. I think you know what to do in that case: if no file went down the changed path, `git reset --hard "$i18n_head"`.

PR #1886, opened 2022-03-07
Refactor contest join checks

Riolku:
As suggested, merge the access checks into one. Some tests had to be changed, because the old checks didn't check the dates.


Riolku:
@Ninjaclasher since you suggested this can I get your review?

judge/views/contests.py
```diff
@@ -380,13 +385,9 @@ def join_contest(self, request, access_code=None):
...
         else:
             SPECTATE = ContestParticipation.SPECTATE
-            LIVE = ContestParticipation.LIVE
 
-            if contest.is_live_joinable_by(request.user):
-                participation_type = LIVE
-            elif contest.is_spectatable_by(request.user):
-                participation_type = SPECTATE
-            else:
+            participation_type = contest.get_join_type(request.user)
```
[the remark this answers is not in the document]
Riolku: I've deleted the code. This means that on spectate creation, an access code is required. Is that intended? This might be a regression.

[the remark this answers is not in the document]
Riolku: Regressed in 8b04126. Let me fix.

Riolku:
As mentioned in the comment under Ninjaclasher's review, it was noted that there's a regression, introduced in 8b04126, that forced people to enter access codes when spectating. It has been fixed here. Some useless code also deleted.

Ninjaclasher (approved):

judge/views/contests.py
```diff
@@ -192,6 +194,11 @@ def get_context_data(self, **kwargs):
         context['is_spectator'] = self.is_spectator
         context['can_edit'] = self.can_edit
 
+        participation_type = self.object.get_join_type(self.request.user)
+        context['can_test'] = participation_type == ContestParticipation.TESTING
```
int-y1, line 198: Does `TESTING` exist? Also, `can_test` is not currently used in any templates.
```py
>>> from judge.models import ContestParticipation
>>> ContestParticipation.LIVE
0
>>> ContestParticipation.TESTING
  File "<console>", line 1, in <module>
AttributeError: type object 'ContestParticipation' has no attribute 'TESTING'
```

judge/views/contests.py
```diff
@@ -380,13 +387,9 @@ def join_contest(self, request, access_code=None):
                     break
         else:
             SPECTATE = ContestParticipation.SPECTATE
```
int-y1, line 389: Nit: Delete this, and replace `virtual=SPECTATE` with `virtual=ContestParticipation.SPECTATE`.

int-y1 (changes requested):

judge/views/contests.py
Riolku: Yeah this is a mixup I made with #1941.

judge/models/tests/test_contest.py
```diff
@@ -660,8 +645,6 @@ def test_private_contest_methods(self):
             'normal_open_org': {
                 'can_see_own_scoreboard': self.assertTrue,
                 'can_see_full_scoreboard': self.assertTrue,
-                'is_live_joinable_by': self.assertTrue,
-                'is_spectatable_by': self.assertTrue,
```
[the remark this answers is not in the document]
Riolku: No. See the comment: the contest is over.

judge/models/tests/test_contest.py
```diff
@@ -717,8 +707,7 @@ def test_organization_private_contest_methods(self):
             'normal': {
                 'can_see_own_scoreboard': self.assertTrue,
                 'can_see_full_scoreboard': self.assertTrue,
-                'is_live_joinable_by': self.assertTrue,
-                'is_spectatable_by': self.assertTrue,
+                'get_join_type': self.assertLive,
```
int-y1, line 710: `organization_private_contest` is ongoing. Add `get_join_type` to the other items.

int-y1 (approved):
lgtm

judge/views/contests.py
```diff
@@ -122,6 +122,8 @@ def get_context_data(self, **kwargs):
         context['current_contests'] = present
         context['future_contests'] = future
         context['finished_contests'] = finished
+        context['LIVE'] = ContestParticipation.LIVE
+        context['SPECTATE'] = ContestParticipation.SPECTATE
```
Xyene, line 126: Can we make this `context['is_spectator'] = ...` instead? Passing constants like this into the template feels a little icky.

Riolku: Nevermind, we can't do this because uh, there are multiple participations...

So we should make a method on the object.

Riolku: There's a world in which we refactor this out to an enum to make this all a little nicer, but the fix implemented here is just way easier.

PR #1892, opened 2022-03-12
Update Ruby key in language fixture

Ninjaclasher:
There is also an instance of `RUBY2` in a migration (https://github.com/DMOJ/online-judge/blob/master/judge/migrations/0112_language_extensions.py#L57), but I'm not sure if we should be updating old migrations.


quantum5:
There is no need to update the migration.

quantum5 (approved):

PR #1894, opened 2022-03-18
Fix i18n messages for pwned passwords

int-y1:
Previously, the translation was done once, at init. Make sure the translation is done once per request.

Also, remove `__init__` because its arguments are not being used.

Before:
c1

After:
Capture


Ninjaclasher (approved):
LGTM

Riolku:
Unit tests seem to be broken, something about an unsupported git version.

quantum5 (approved):

PR #1895, opened 2022-03-18
Add `implicit_problem_perms` field to contest

Riolku:
Contest testers (and potentially curators and authors) cannot see problems if they are not spectating the contest. This commit adds a checkbox which if on, allows testers and editors to see the problems outside of the contest. 

This is especially helpful in combination with `testers_see_submissions`, because otherwise, although testers can see detailed verdicts, they cannot see the problem itself if they are not spectating.

Furthermore, because only visible can be selected in contest admin, this feature cannot be used for privilege escalation.


Riolku:
Let's not do this. It's just going to hurt the database.

PR #1897, opened 2022-03-19
Fix i18n in fields

int-y1:
```
git grep "verbose_name=[^_]"
git grep "help_text=[^_]"
```

### DIFF
--- a/judge/models/runtime.py
+++ b/judge/models/runtime.py
@@ -134,7 +134,7 @@ class Judge(models.Model):
     load = models.FloatField(verbose_name=_('system load'), null=True,
                              help_text=_('Load for the last minute, divided by processors to be fair.'))
     description = models.TextField(blank=True, verbose_name=_('description'))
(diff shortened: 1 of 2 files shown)


Ninjaclasher (approved):
LGTM

judge/models/runtime.py
```diff
@@ -134,7 +134,7 @@ class Judge(models.Model):
     load = models.FloatField(verbose_name=_('system load'), null=True,
                              help_text=_('Load for the last minute, divided by processors to be fair.'))
     description = models.TextField(blank=True, verbose_name=_('description'))
-    last_ip = models.GenericIPAddressField(verbose_name='Last connected IP', blank=True, null=True)
+    last_ip = models.GenericIPAddressField(verbose_name=_('Last connected IP'), blank=True, null=True)
```
quantum5, line 137:
```suggestion
    last_ip = models.GenericIPAddressField(verbose_name=_('last connected IP'), blank=True, null=True)
```

Let's fix this while we are at it.

quantum5 (approved):

PR #1898, opened 2022-03-19
Add i18n to a line of text

int-y1:


templates/base.html
```diff
@@ -291,7 +291,7 @@ <h2 style="color:#393630; display:inline">
     <footer>
         <span id="footer-content">
             <br>
-            <a style="color: #808080" href="https://dmoj.ca">proudly powered by <b>DMOJ</b></a> |
+            <a style="color: #808080" href="https://dmoj.ca">{{ _('proudly powered by <b>DMOJ</b>') }}</a> |
```
Ninjaclasher, line 294: I don't think HTML tags work with translations. We should probably switch to markdown ([Example](https://github.com/DMOJ/online-judge/blob/master/templates/contest/contest.html#L97)).

int-y1: HTML tags worked here.

I didn't get markdown to work, because `|markdown('default')` wraps the text in a `<p>` and creates unwanted whitespace.

quantum5: You will be able to do `|markdown('default', strip_paragraphs=True)` after #1899.

int-y1: Done. Markdown renders well.
```html
            <br>
            <a style="color: #808080" href="https://dmoj.ca">proudly powered by <strong>DMOJ</strong>
</a> |
```

PR #1899, opened 2022-03-20
Use new safe translation module for Jinja2

quantum5:
This prevents translators from injecting HTML into the template via translation tags.

### DIFF
--- a/templates/user/pp-row.html
+++ b/templates/user/pp-row.html
@@ -21,10 +21,7 @@
         <a href="{{ url('submission_status', breakdown.sub_id) }}">{{ breakdown.points|floatformat(0) }}pp</a>
     </div>
     <div class="pp-weighted">
-        {% trans weight=breakdown.weight|floatformat(0) %}
-            weighted <b>{{ weight }}%</b>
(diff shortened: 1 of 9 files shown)


templates/user/pp-row.html
```diff
@@ -21,10 +21,7 @@
         <a href="{{ url('submission_status', breakdown.sub_id) }}">{{ breakdown.points|floatformat(0) }}pp</a>
     </div>
     <div class="pp-weighted">
-        {% trans weight=breakdown.weight|floatformat(0) %}
-            weighted <b>{{ weight }}%</b>
-        {% endtrans %}
-
+        {{ _('weighted %(percent)', percent='<b>%s%%</b>'|safe|format(breakdown.weight|floatformat(0))) }}
```
int-y1, line 24: Should this be `%(percent)s`?

quantum5: Yes, fixed.

Riolku: We should resolve this.

Xyene (approved):

PR #1900, opened 2022-03-20
Remove unsafe translations in Python code

### LINKED ISSUE #1494: Remove HTML tags in i18n strings
We do not want translators to be able to change the HTML in potentially malicious ways. Instead, we should follow the example in https://github.com/DMOJ/online-judge/blob/master/judge/views/ticket.py#L87-L90 and do it correctly:

```python
    def get_content_title(self):
        return mark_safe(escape(_('New ticket for %s')) %
                         format_html('<a href="{0}">{1}</a>', reverse('problem_detail', args=[self.object.code]),
                                     self.object.translated_name(self.request.LANGUAGE_CODE)))
```

In fact, we might want to introduce a new helper function for this purpose alone.

quantum5:
This prevents translators from injecting HTML via messages translated in Python.

Closes #1494.

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -114,8 +114,9 @@ class ProblemSolution(SolvedProblemMixin, ProblemMixin, TitleMixin, CommentedDet
         return _('Editorial for {0}').format(self.object.name)
 
     def get_content_title(self):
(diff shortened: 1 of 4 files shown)


Ninjaclasher:
I believe this will close #1494.

Xyene (approved):
LGTM

Issue #1318, opened 2020-04-10
Batch queueing functionality in bridge


Xyene:
Currently if you wish to rejudge 2000 submissions, a Celery task is created that operates at ~10 submissions/s.

PR #1901, opened 2022-03-21
Implement batch scheduling bridge call

### LINKED ISSUE #1501: Batch scheduling bridge call
Currently, if the site wishes to schedule _N_ submissions for rejudging, it must make _N_ bridge API calls. This can be very slow when _N_ is large (enough so that there is a Celery screen for it), and can be alleviated by exposing batch scheduling functionality.

Ninjaclasher:
Closes #1501.


Ninjaclasher:
Oops, actually closes #1318.

quantum5:
Did you know that you can edit the description?

judge/models/submission.py
```diff
@@ -121,18 +121,47 @@ def long_status(self):
...
+    def batch_judge(cls, submissions, *, rejudge=False, force_judge=False, rejudge_user=None, batch_size=100, **kwargs):
+        # Don't trust the caller to follow related objects that we need.
+        submissions = submissions.select_related('problem', 'language', 'source')
+
+        if not force_judge:
+            submissions = submissions.exclude(locked_after__lt=timezone.now())
+
+        # Force evaluate the queryset here so we have a fixed set of submissions that we are working with
+        # rather than a lazy queryset that could change.
+        submissions = list(submissions)
```
Ninjaclasher, line 146: Thought about using `.iterator()` instead of forcefully evaluating and caching the entire queryset so we don't use too much memory, but not sure if it's worth it given that MySQL doesn't support streaming results and so the entire result set would be loaded anyways ([Source](https://docs.djangoproject.com/en/2.2/ref/models/querysets/#without-server-side-cursors)). The benefit is that we wouldn't be storing expensive Python objects, but it would complicate this code more.

quantum5: No, let's use `.iterator()`. Let me create a quick chunk helper function and you can use it then.

Ninjaclasher: Pulled in #1903 temporarily for completeness.

Riolku:
I scheduled 56 submissions for rejudging and nothing broke :+1: .

judge/admin/submission.py
```diff
@@ -160,7 +160,6 @@ def judge(self, request, queryset):
             self.message_user(request, gettext('You do not have the permission to rejudge submissions.'),
                               level=messages.ERROR)
             return
-        queryset = queryset.order_by('id')
```
Riolku, line 163: Is this unrelated to the main focus of this PR? Just curious.

judge/models/submission.py
```diff
@@ -121,18 +122,41 @@ def long_status(self):
...
                         revisions.set_user(rejudge_user)
                     revisions.set_comment('Rejudged')
                     revisions.add_to_revision(self)
-            judge_submission(self, *args, rejudge=rejudge, **kwargs)
+            judge_submission(self, rejudge=rejudge, **kwargs)
 
     judge.alters_data = True
 
+    @classmethod
+    def batch_judge(cls, submissions, *, rejudge=False, force_judge=False, rejudge_user=None, chunk_size=100, **kwargs):
```
quantum5, line 138: Would it be less clunky to put this on the QuerySet given how you are using it?

judge/judgeapi.py
```diff
@@ -100,6 +104,65 @@ def judge_submission(submission, rejudge=False, batch_rejudge=False, judge_id=No
...
+                .annotate(pretested=F('contest_object__run_pretests_only').bitand(F('contest__problem__is_pretested')))
+                .values_list('pretested')[:1],
+                output_field=BooleanField(),
+            ),
+            False,
+        ),
+    }
+
+    with transaction.atomic():
+        submission_queryset = Submission.objects.filter(id__in=map(attrgetter('id'), submissions)) \
```
quantum5, line 124: If the argument is a queryset, then you wouldn't need to do this hack and could just do `.exclude` on the queryset directly. And then you can just load this queryset to get the `submissions` instead of filtering in Python.

judge/judgeapi.py
```diff
@@ -100,6 +104,65 @@ def judge_submission(submission, rejudge=False, batch_rejudge=False, judge_id=No
...
+                                                .exclude(status__in=('P', 'G'))
+        submission_queryset.update(**updates)
+
+        ids = set(submission_queryset.values_list('id', flat=True))
+        # Do the filtering using the new set of IDs rather than submission_queryset
+        # because submission_queryset itself takes a list of IDs anyways.
+        SubmissionTestCase.objects.filter(submission_id__in=ids).delete()
```
quantum5, line 131: And then you can just pass `submission_queryset` as the argument here.

judge/judgeapi.py
```diff
@@ -100,6 +104,65 @@ def judge_submission(submission, rejudge=False, batch_rejudge=False, judge_id=No
...
+                'language': submission.language.key,
+                'source': submission.source.source,
+            } for submission in submissions],
+            'judge-id': judge_id,
+            # This is technically incorrect when rejudge is False.
+            # Some submissions might be contest submissions and should have priority "CONTEST_SUBMISSION_PRIORITY".
+            # It's not worth further complicating this code by separating contest submissions and
+            # normal submissions as that case is not even being used.
+            # For the moment, let's just set all submissions to have "DEFAULT_PRIORITY" if we're not rejudging.
+            'priority': BATCH_REJUDGE_PRIORITY if rejudge else DEFAULT_PRIORITY,
```
quantum5, line 150: Why not just move the `priority` into the list? Then you can set per-submission priorities. In fact priority is a per-submission priority and should be in there anyway...

judge/judgeapi.py
Ninjaclasher: If the argument is a queryset, that means it will be lazy. We would be doing the splicing at the database layer, so the slice may change between batches. For example, consider we were trying to rejudge `Submission.objects.exclude(status='QU')`. As we're processing each batch, [0:batch_size] would always contain a not judged queryset since the previous batch's submissions would become `QU` (and thus filtered out). We would only process the [0:batch_size] splice once, and so we would be missing some submissions.

From what I could tell, there's 2 solutions:
 - Use a single batch. This wouldn't handle large querysets well, so I opted not to do this.
 - Evaluate the queryset so we have a fixed set of submissions to judge in each batch, which is what is currently done via `.iterator()`.

Other solutions are welcome.

PR #1904, opened 2022-03-23
Make submission history chart timezone aware

Ninjaclasher:


judge/views/user.py
```diff
@@ -195,7 +203,7 @@ def get_context_data(self, **kwargs):
         context['submission_metadata'] = mark_safe(json.dumps({
             'min_year': (
                 self.object.submission_set
-                .annotate(year_only=ExtractYear('date'))
+                .annotate(year_only=ExtractYear('date', tzinfo=timezone.get_current_timezone()))
```
quantum5, line 206: This seems unnecessary — [the documentation](https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#extract) states that the current timezone is automatically used, and the code confirms that just like above.

judge/views/user.py
```diff
@@ -183,9 +183,17 @@ def get_context_data(self, **kwargs):
...
 
+        # Django 2.2 doesn't like truncating directly into a DateField with a timezone,
+        # so we have to truncate to a DateTimeField and cast to a DateField.
+        # In Django 3+, we can switch to TruncDate().
         submissions = (
             self.object.submission_set
-            .annotate(date_only=Cast('date', DateField()))
+            .annotate(
+                date_only=Cast(
+                    Trunc('date', 'date', output_field=DateTimeField(), tzinfo=timezone.get_current_timezone()),
```
quantum5, line 193: The second argument to `Trunc` is `kind` and should probably be passed as a keyword argument. The [documentation](https://docs.djangoproject.com/en/2.2/ref/models/database-functions/#trunc) seems to suggest that you should use `kind='day'` and the active timezone is automatically used:
> If a different timezone like `Australia/Melbourne` is active in Django, then the datetime is converted to the new timezone before the value is truncated.
This is confirmed by the [`TimezoneMixin`](https://docs.djangoproject.com/en/2.2/_modules/django/db/models/functions/datetime/) source code.

judge/views/user.py
Ninjaclasher: Reverted.

judge/views/user.py
Ninjaclasher: It looks like we don't actually need any of this. Since all of these functions default to the current timezone including `TruncDate`, we can just use `TruncDate` instead.

quantum5 (approved):

PR #1909, opened 2022-03-25
Import Markup from MarkupSafe instead of jinja2

Ninjaclasher:
From the changelog:
```
Version 3.1.0
Released 2022-03-24
...
Markup and escape should be imported from MarkupSafe.
```
`MarkupSafe` is already a dependency for `jinja2`, so no new dependencies need to be added here.


kiritofeng (approved):
LGTM

PR #1923, opened 2022-04-16
Remove dead templates

int-y1:
- `blog/dashboard.html` is not used, so remove it.
- No styles in `blog/media-css.html` are being used, so remove it.


templates/blog/dashboard.html
```diff
@@ -1,34 +0,0 @@
...
-            <img src="{{ gravatar(request.user, 135) }}"
-                 alt="gravatar" width="135px" height="135px">
-        </div>
-        <div class="recently-attempted">
-            <h4>Recently attempted problems</h4>
-            <ul>
-                {% for code, name, problem_points, user_points, s_date in recently_attempted_problems %}
-                    <li>
-                        <a href="{{ url('problem_detail', code) }}">{{ name }}</a>
-                        [<a href="{{ url('user_submissions', code, request.user.username }}">
```
int-y1, line 15: Fun fact: This causes an error because `url(` doesn't have a matching `)`

quantum5 (approved):

PR #1927, opened 2022-04-22
Disallow voting on own comments

### LINKED ISSUE #1517: do not allow users to vote on their own comments
probably a prerequisite before #839 can be implemented?

int-y1:
Fixes #1517. New self-votes cannot be created. The migration deletes all self-votes, and recalculates all comment scores.

### DIFF
--- a/judge/migrations/0132_no_self_vote.py
+++ b/judge/migrations/0132_no_self_vote.py
@@ -11,11 +11,11 @@ def delete_self_votes(apps, schema_editor):
     Comment.objects.update(
         score=Subquery(
             Comment.objects.filter(
-                id=OuterRef('id')
+                id=OuterRef('id'),
             ).annotate(
-                foo=Coalesce(Sum('votes__score'), 0)
-            ).values('foo')[:1]
-        )
+                newscore=Coalesce(Sum('votes__score'), 0),
+            ).values('newscore')[:1],
+        ),
     )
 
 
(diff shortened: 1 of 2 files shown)


quantum5:
I think you can do something like `Comment.objects.update(score=Sum('votes__score'))`. If you can, I think you should make a migration.

int-y1:
This gave me `django.core.exceptions.FieldError: Joined field references are not permitted in this query`

judge/migrations/0132_no_self_vote.py
```diff
@@ -0,0 +1,30 @@
...
+    CommentVote.objects.filter(voter=F('comment__author')).delete()
+    Comment.objects.update(
+        score=Subquery(
+            Comment.objects.filter(
+                id=OuterRef('id')
+            ).annotate(
+                foo=Coalesce(Sum('votes__score'), 0)
+            ).values('foo')[:1]
+        )
+    )
```
quantum5, line 19: Nice, but I probably wouldn't call it `foo`.

quantum5 (approved):

judge/migrations/0132_no_self_vote.py
int-y1: Renamed to `newscore`

int-y1:
The answer was https://docs.djangoproject.com/en/4.0/ref/models/expressions/#using-aggregates-within-a-subquery-expression

Ninjaclasher (approved):
LGTM

PR #1928, opened 2022-04-22
Fix countdown space

int-y1:
During the first second of page load, the countdown says `Starting in 365384 days 23:01:56 .`

The new HTML fragment is:
```html
<span class="time">Starting in <span data-secs="31569260516.83524" class="time-remaining">365384 days 23:01:56</span>.</span>
```


kiritofeng (approved):
LGTM, thanks!

PR #1933, opened 2022-05-11
Add file submission to problem page

alexwaeseperlman:
- Changed the button labelled "Submit!" on the problem page to say "Go to editor"
- Added a small form beneath the "Go to editor" button, initially hidden behind a chevron, that allows users to submit a file. I tried to make the changes as unobtrusive as possible.

Notes:
- The file submission form goes to `/problem/<problem-code>/submit`, which means that error messages, for example when submitting a file with length greater than 65535, show up in the editor. This isn't amazing UX but it might be acceptable. 
- It might be worth adding a button to submit files from the editor page
- If a user selects a file and then clicks "Go to editor", should it load that file into their editor?

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -9,7 +9,7 @@ from django.contrib.auth.forms import AuthenticationForm
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.db.models import Q
-from django.forms import BooleanField, CharField, ChoiceField, Form, ModelForm, MultipleChoiceField
+from django.forms import BooleanField, CharField, ChoiceField, Form, ModelForm, MultipleChoiceField, FileField
 from django.urls import reverse_lazy
 from django.utils.translation import gettext_lazy as _
 
@@ -124,7 +124,8 @@ class DownloadDataForm(Form):
 
 
 class ProblemSubmitForm(ModelForm):
-    source = CharField(max_length=65536, widget=AceWidget(theme='twilight', no_ace_media=True))
+    source = CharField(max_length=65536, widget=AceWidget(theme='twilight', no_ace_media=True), required=False)
+    file = FileField(required=False, widget=forms.FileInput)
     judge = ChoiceField(choices=(), widget=forms.HiddenInput(), required=False)
 
     def __init__(self, *args, judge_choices=(), **kwargs):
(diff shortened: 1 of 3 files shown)


ehhthing:
Perhaps it would be be much less code if this was just implemented client side with JS so there would be no need for any kind of server code changes.

alexwaeseperlman:
I added 16 lines of server side code. It's possible that using client side JS would take less but it won't be a huge difference. Unless there is a better reason to switch I think we can leave things as they are.

templates/problem/problem.html
```diff
@@ -108,14 +116,40 @@ <h2 style="color:#393630; display: inline-block">{{ title }}</h2>
...
+            <a class="unselectable button full disabled">{{ _('Go to editor') }}</a>
             <div class="no-submissions-left submissions-left">{{ _('0 submissions left') }}</div>
         {% endif %}
     {% else %}
         <a href="{{ url('problem_submit', problem.code) }}" class="unselectable button full">
-            {{ _('Submit solution') }}
+            {{ _('Go to editor') }}
         </a>
     {% endif %}
+	{% if request.user.is_authenticated and (not request.in_contest or submission_limit) %}
```
Ninjaclasher, line 127: Nit: please use spaces instead of tabs.

judge/forms.py
```diff
@@ -139,11 +140,21 @@ def __init__(self, *args, judge_choices=(), **kwargs):
...
+                if len(source) >= 65536: raise ValidationError(_('Source file is too long.'))
+                cleaned_data['source'] = source
+        elif cleaned_data['source'] is None: raise ValidationError(_('Source and file are both empty.'))
+        return cleaned_data
+
     class Meta:
         model = Submission
         fields = ['language']
 
-
```
Ninjaclasher, line 146: Nit: don't delete newline.

Ninjaclasher (commented):
Just a general comment: I think we should move this to the `Submit Solution` rather than at the bottom of the problem statement page. This centralizes the location where a user can submit a solution.

As a consequence of this, we can revert all the changes in `judge/views/problem.py` and revert `Go to editor` to `Submit solution`.

Also, could you run `flake8` locally to fix all the linting issues?

ehhthing:
> I added 16 lines of server side code. It's possible that using client side JS would take less but it won't be a huge difference. Unless there is a better reason to switch I think we can leave things as they are.

The main issue is that handling file uploads on the server side is much more prone to issues than simply using a text field. 

For example, your code is vulnerable to DoS attacks if the user uploads a very large file composed of only 0s since it decodes line-by-line. In addition, it would cause a 500 if you were to upload a file that included invalid UTF-8 because it tries to decode without a try-except. 

Typically, Django does most of the validation work for you if you use a normal text field, but it provides much less built in protection against these bugs if you want to handle file uploads.

Moving the processing to the client side significantly reduces the bug surface simply because handling file uploads on the server side is nontrivial.

alexwaeseperlman:
> The main issue is that handling file uploads on the server side is much more prone to issues than simply using a text field.

I removed all modifications to server side code. Thanks for looking at this.

> Just a general comment: I think we should move this to the Submit Solution page rather than at the bottom of the problem statement page. This centralizes the location where a user can submit a solution.

In my opinion the main benefit of this feature is that it creates a shortcut for submitting from the problem page. I'm open to moving it around though.

Here are some screenshots of the "Submit solution"/"Go to editor" button in case anyone wants to look at the changes without setting them up:

Old:
image

New with dropdown closed:
image

New with dropdown open:
image

PR #1939, opened 2022-05-21
Fix filters that break i18n

int-y1:
These issues were found with `git grep "_(.*=.*|"`

- Change `as_countdown` to a `registry.function`. Fix issues with `|timedelta`. Fix pp messages.
- Test that all affected pages still work properly.
- Test that makemessages can find the i18n strings:
```
python manage.py makemessages -l en -e py,html,txt
git diff upstream/update-i18n -- locale/en/LC_MESSAGES/django.po
```

### DIFF
--- a/judge/jinja2/timedelta.py
+++ b/judge/jinja2/timedelta.py
@@ -22,7 +22,7 @@ def seconds(timedelta):
     return timedelta.total_seconds()
 
 
-@registry.filter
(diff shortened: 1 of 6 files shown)


Ninjaclasher (approved):
LGTM

Xyene (approved):
LGTM, thanks!

Issue #1670, opened 2021-05-07
Add TESTING ContestParticipation type


kiritofeng:
One of the things that some contests (i.e. WAC) likes to do is to throw their testers in a windowed contest to see how well they perform in that window. With spectating, contest authors can't watch how their testers are performing (since it's untimed), and would need to calculate how long it takes the tester to solve each problem manually (which was how it was done before). I think it would be more useful for contest authors to be able to watch their testers through participations. 

Ideally, we would use a new virtual number (e.g `-2` for "testing"), and display testing participations in the `Participations` tab (maybe hide it so only contest authors and curators can see), but that's a lot of refactoring, so for now, we can probably use the built in virtual join system (perhaps something like "If the contest hasn't begun and the user is a tester, use virtual join. If the contest is ongoing and the user is a tester, use spectate").

Some other opinions on this would be good.

_Originally posted by @Ninjaclasher in https://github.com/DMOJ/online-judge/pull/1668#r627553068_

Riolku:
I wonder if it's really best to use a number to indicate participation type. I guess it's not the end of the world, and I can't think of any better ways to refactor this if we are to keep the current virtual join system.

I am against putting the data in the Participation tab. I feel like that should show your participations, as it currently does. Instead, we can add another tab, and reuse a large amount of the current code, and throw all the testers onto one board.

PR #1941, opened 2022-05-23
Testing participation

Riolku:
Depends on #1886. Lots of code was refactored in order for this change to not be horrible to implement.

This is a hefty change, so testing from others would be appreciated.


Riolku:
Closes #1670.

Riolku:
This PR needs some more unit tests for the new methods. I will do this soon.

PR #1948, opened 2022-05-29
Translate the problem types in a dropdown

int-y1:
Go to `/problems/`. `ru` contains [a translation for `String Algorithms`](https://github.com/DMOJ/online-judge/blob/966b37d52fe88ec28fbd845cf31164a56d900c17/locale/ru/LC_MESSAGES/dmoj-user.po#L88-L89), so the dropdown looks like:

Capture

### DIFF
--- a/templates/problem/search-form.html
+++ b/templates/problem/search-form.html
@@ -52,7 +52,7 @@
                     <select id="types" name="type" multiple>
                         {% for type in problem_types %}
                             <option value="{{ type.id }}"{% if type.id in selected_types %} selected{% endif %}>
-                                {{ type.full_name }}
+                                {{ user_trans(type.full_name) }}
                             </option>
                         {% endfor %}
                     </select>


templates/problem/search-form.html
```diff
@@ -52,7 +52,7 @@ <h3>{{ _('Problem search') }} <i class="fa fa-search"></i>
                     <select id="types" name="type" multiple>
                         {% for type in problem_types %}
                             <option value="{{ type.id }}"{% if type.id in selected_types %} selected{% endif %}>
-                                {{ type.full_name }}
+                                {{ user_trans(type.full_name) }}
```
Ninjaclasher, line 55: Would it make sense to add a similar change for problem categories?

int-y1: Problem categories are out of scope. In particular, they require these steps:

- Modify `makedmojmessages` to get all problem categories. Example: https://github.com/DMOJ/online-judge/blob/8eed5e5278f68c35a0b46bcce17c1a64fae5dd78/judge/management/commands/makedmojmessages.py#L127-L130
- Wait for Crowdin translators to translate those strings, and an admin to get those translations.
- Use `user_trans` in the appropriate templates.

Ninjaclasher: I see, this is out of scope.

Ninjaclasher (approved):

Xyene (approved):
LGTM, thanks!

PR #1953, opened 2022-06-02
i18n plurals in error messages

int-y1:
- Fix i18n issues with `TOTP_CODE_LENGTH`. Note: `TOTP_CODE_LENGTH` was introduced in #1708. I don't see why this variable is needed (mostly because 6 is still hardcoded in tons of other places). Feel free to suggest its removal.
- Support plurals in the max org error message.
- Fix plurals in this error modal:

Capture

## Testing

`makemessages` can detect all i18n strings properly.

Go to `~/site/locale/ru/LC_MESSAGES/django.po`. Keep the header and delete everything else. Insert:
```po
#: judge/forms.py:28
#, python-brace-format
msgid "Two-factor authentication tokens must be {count} decimal digit."
msgid_plural "Two-factor authentication tokens must be {count} decimal digits."
msgstr[0] "p0 {count}"
msgstr[1] "p1 {count}"
msgstr[2] "p2 {count}"
msgstr[3] "p3 {count}"

#: judge/forms.py:83 judge/views/organization.py:147
#, python-brace-format
msgid "You may not be part of more than {count} public organization."
msgid_plural "You may not be part of more than {count} public organizations."
msgstr[0] "A {count}"
msgstr[1] "B {count}"
msgstr[2] "C {count}"
msgstr[3] "D {count}"
```

Run `python manage.py compilemessages`.

- Go to `/accounts/2fa/enable/`. Test `_lazy` by switching between `en` and `ru`. Play around with `TOTP_CODE_LENGTH = 6` to see the different plural forms.
- Go to `/edit/profile/`. Join organizations until you reach the max limit.
- Go to `/organization/<num>-<slug>`. Join until you reach the max limit.

### DIFF
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -11,7 +11,8 @@ from django.core.validators import RegexValidator
 from django.db.models import Q
 from django.forms import BooleanField, CharField, ChoiceField, Form, ModelForm, MultipleChoiceField
 from django.urls import reverse_lazy
-from django.utils.translation import gettext_lazy as _
+from django.utils.text import format_lazy
+from django.utils.translation import gettext_lazy as _, ngettext_lazy
 
 from django_ace import AceWidget
 from judge.models import Contest, Language, Organization, Problem, Profile, Submission, WebAuthnCredential
@@ -24,7 +25,9 @@ two_factor_validators_by_length = {
     TOTP_CODE_LENGTH: {
         'regex_validator': RegexValidator(
             f'^[0-9]{{{TOTP_CODE_LENGTH}}}$',
-            _(f'Two-factor authentication tokens must be {TOTP_CODE_LENGTH} decimal digits.'),
+            format_lazy(ngettext_lazy('Two-factor authentication tokens must be {count} decimal digit.',
+                                      'Two-factor authentication tokens must be {count} decimal digits.',
+                                      TOTP_CODE_LENGTH), count=TOTP_CODE_LENGTH),
         ),
         'verify': lambda code, profile: not profile.check_totp_code(code),
         'err': _('Invalid two-factor authentication token.'),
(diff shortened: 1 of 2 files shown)


judge/forms.py
```diff
@@ -24,7 +25,7 @@
     TOTP_CODE_LENGTH: {
         'regex_validator': RegexValidator(
             f'^[0-9]{{{TOTP_CODE_LENGTH}}}$',
-            _(f'Two-factor authentication tokens must be {TOTP_CODE_LENGTH} decimal digits.'),
+            format_lazy(_('Two-factor authentication tokens must be {count} decimal digits.'), count=TOTP_CODE_LENGTH),
```
quantum5, line 28: While you are at it, why not use `ngettext`? Imagine changing the code length to 8 and then having to do eclipsis when translating to Irish...

judge/forms.py
```diff
@@ -78,9 +80,9 @@ def clean(self):
         max_orgs = settings.DMOJ_USER_MAX_ORGANIZATION_COUNT
 
         if sum(org.is_open for org in organizations) > max_orgs:
-            raise ValidationError(ngettext('You may not be part of more than {count} public organization.',
-                                           'You may not be part of more than {count} public organizations.',
-                                           max_orgs).format(count=max_orgs))
+            raise ValidationError(ngettext_lazy('You may not be part of more than {count} public organization.',
```
quantum5, line 83: What's the point of doing this lazily?

judge/forms.py
int-y1: Done

judge/forms.py
int-y1: To avoid an extra import

quantum5:
By the way, the original point of `TOTP_CODE_LENGTH` was to make it obvious that 6 is TOTP in `two_factor_validators_by_length`.

judge/forms.py
quantum5: In theory, imports are a lot cheaper than doing the extra work every invocation with laziness, but whatever.

quantum5 (approved):

PR #1954, opened 2022-06-04
Refactor user submission table's js

int-y1:
JS's `ngettext` doesn't work in `templates/`, so move it to `resources/`.

i18n: `python manage.py makemessages -l en -d djangojs` detected all 5 i18n strings and put them in `djangojs.po`.

JS compatibility: The submission table works in IE11. (This required a lot of small changes)

Capture

### DIFF
--- a/resources/user_profile.js
+++ b/resources/user_profile.js
@@ -1,0 +1,121 @@
+var active_tooltip = null;
+
+function display_tooltip(where) {
+    if (active_tooltip !== null) {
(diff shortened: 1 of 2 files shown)


kiritofeng:
I feel like "not working in IE" is totally fine (iirc MS is going to deprecate IE mode soon:tm:)

Xyene (approved):
LGTM, thanks!

PR #1956, opened 2022-06-12
Fix 500 error on /contest/<code>/leave

int-y1:
The 500 error: visit https://dmoj.ca/contest/cheerio1j/leave while logged in.

The issue is that a GET request goes to `BaseDetailView.get` then crashes at `self.render_to_response`. https://github.com/django/django/blob/5c3300027b30c1b498d99010f7d618316f685045/django/views/generic/detail.py#L103-L108

In the new version, GET requests receive error 405.

Also, `ContestJoin` doesn't need `BaseDetailView` because `ContestJoin` implements its own `get` method.


Ninjaclasher (approved):
LGTM

Xyene (approved):
LGTM, thanks!

PR #1960, opened 2022-06-24
Ban pointless Unicode characters from problem statements

int-y1:
- i saw `ﬀ` in a problem statement
- i saw a problem that said ``Output ... `−1` if it is not possible.``


Xyene (approved):
:pray:

PR #1976, opened 2022-07-20
Fix regression caused by 42ee283; fixes #1970

### LINKED ISSUE #1970: Fix CSS on contest pages
On contest pages with many problems, the table is no longer scrollable; this is due to 42ee283e3d4124985d29dc6d0cbfada3f5d83d97.

kiritofeng:
Turns out `width: 100%;` is in fact necessary to prevent the contest ranking page from overflowing.

This patch is currently applied in prod.

### DIFF
--- a/resources/content-description.scss
+++ b/resources/content-description.scss
@@ -19,6 +19,7 @@
 
     #content-left {
         flex: 5;
+        width: 100%;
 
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks!

PR #1533, opened 2020-09-11
Allow individual organisations' user list to be sortable

slightlyskepticalpotat:
Closes #1483.

PR #1977, opened 2022-07-27
Allow sorting of organization's user list

### LINKED ISSUE #1483: Add Sorting For Organisations
Currently the organisation user list is only sorted by the point count. 

image

int-y1:
Closes #1483. Makes #1533 redundant.

Allows sorting for the organization user list, and the class user list.

Also, clean up the css so that the headers are displayed as white.


Xyene (approved):
LGTM, thanks!

PR #1979, opened 2022-07-27
Fix rare bug on copying text

int-y1:
### The bug

When a code block has a specified language, but this language is not in Pygments, the copy button is missing. e.g. Turing code for [aplusb](https://dmoj.ca/problem/aplusb)


Xyene (approved):
LGTM, thanks!

PR #1980, opened 2022-07-27
Add i18n to server-side messages

int-y1:
Affected pages:

- `/admin/judge/contest/1/change/` > Problems > Rejudge column
- `/admin/judge/judge/add/` > Regenerate button
- `/admin/judge/submission/` > Rejudge button (and Locked button, if sub is locked)
- `/contest/c1/participations/<user>` page's title
- Too many running submissions
- Submit to an unviewable problem

### DIFF
--- a/judge/admin/contest.py
+++ b/judge/admin/contest.py
@@ -73,8 +73,8 @@ class ContestProblemInline(SortableInlineAdminMixin, admin.TabularInline):
     def rejudge_column(self, obj):
         if obj.id is None:
             return ''
-        return format_html('<a class="button rejudge-link" href="{}">Rejudge</a>',
-                           reverse('admin:judge_contest_rejudge', args=(obj.contest.id, obj.id)))
+        return format_html('<a class="button rejudge-link" href="{0}">{1}</a>',
+                           reverse('admin:judge_contest_rejudge', args=(obj.contest.id, obj.id)), _('Rejudge'))
     rejudge_column.short_description = ''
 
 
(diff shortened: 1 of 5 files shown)


judge/admin/submission.py
```diff
@@ -244,9 +244,10 @@ def language_column(self, obj):
 
     def judge_column(self, obj):
         if obj.is_locked:
-            return format_html('<input type="button" disabled value="Locked"/>')
+            return format_html('<input type="button" disabled value="{0}"/>', _('Locked'))
         else:
-            return format_html('<input type="button" value="Rejudge" onclick="location.href=\'{}/judge/\'" />', obj.id)
+            return format_html('<input type="button" value="{0}" onclick="location.href=\'{1}/judge/\'" />',
```
Ninjaclasher, line 249: While we're changing this code, could we update this to reverse the URL instead of this hack?

int-y1: Done.

judge/admin/contest.py
```diff
@@ -73,7 +73,7 @@ class ContestProblemInline(SortableInlineAdminMixin, admin.TabularInline):
     def rejudge_column(self, obj):
         if obj.id is None:
             return ''
-        return format_html('<a class="button rejudge-link" href="{}">Rejudge</a>',
+        return format_html('<a class="button rejudge-link" href="{1}">{0}</a>', _('Rejudge'),
```
Xyene, line 76: Could we flip the arg order here?

int-y1: Done. The button is still good.

Capture

Xyene (approved):
LGTM, thanks!

PR #1981, opened 2022-07-27
Added dmoj-user.po for the Greek language

ptheoch:


quantum5:
`dmoj-user.po` reflects the state of dmoj.ca and cannot be generalized to other deployments, so it's handled separately and not checked into source control here.

PR #1982, opened 2022-07-27
Fix scratch code's copy button

int-y1:
Fix the scratch code's copy button on the edit profile page. Also, mark some i18n strings as translatable.

Capture


Xyene (approved):
LGTM, thanks!

PR #1985, opened 2022-07-28
Add i18n to js messages

int-y1:
Translate these messages:

- error when voting, e.g. attempt to self-vote
- warning popup when hiding comment
- error when hiding comment
- "updated" message after editing a comment
- `/problem/aplusb/test_data` > Checker > select Floats > hover over input box
- error when closing/opening ticket

### DIFF
--- a/templates/comments/media-js.html
+++ b/templates/comments/media-js.html
@@ -90,7 +90,7 @@
                             on_success();
                     },
                     error: function (data, textStatus, jqXHR) {
-                        alert('Could not vote: ' + data.responseText);
+                        alert('{{ _('Could not vote: {error}') }}'.replace('{error}', data.responseText));
                     }
                 });
             }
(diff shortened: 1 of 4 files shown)


templates/comments/media-js.html
```diff
@@ -169,9 +169,9 @@
                                 if (window.add_code_copy_buttons)
                                     window.add_code_copy_buttons($area);
                                 var $edits = $comment.find('.comment-edits').first();
-                                $edits.text('updated');
+                                $edits.text('{{ _('updated') }}');
                             }).fail(function () {
-                                console.log('Failed to update comment:' + id);
+                                console.log('Failed to update comment: ' + id);
```
Ninjaclasher, line 174: Should this receive the same treatment?

int-y1: I left it out because the user never sees `console.log`'s output.

Ninjaclasher: I think we should change this to an `alert`.

int-y1: Done. The user will see `Failed to update comment body.`

templates/problem/data.html
```diff
@@ -1,15 +1,10 @@
 {% extends "base.html" %}
 
 {% block js_media %}
-    <script type="text/javascript">
-        window.valid_files = {{valid_files_json}};
-
-        $(function () {
-        });
```
Xyene, line 8: A+

Xyene (approved):
LGTM, thanks!

PR #1988, opened 2022-08-07
Use full sentences in help_text

int-y1:
This makes the help text more readable. For example:

Capture

Also, change `base32` to `Base32`.


Xyene (approved):
LGTM, thanks!

PR #1990, opened 2022-08-18
Increase precision on relative time offsets

### LINKED ISSUE #1846: Increase Precision on Relative Time Offsets
On things like contests and submissions, relative times are given. Unfortunately, they aren't that precise. "2 hours ago" can mean anything from 1:30 to 2:30.

I was going to PR this, but it seems that the library in use, [Moment.js](https://momentjs.com) has no real good solution. We use `fromNow()`, here's the [relevant documentation](https://momentjs.com/docs/#/displaying/fromnow/).

What can be done? Would we find another library? Would we try to hack moment.js? What other options do we have?

int-y1:
Fixes #1846. Makes moment.js redundant. The new relative time messages will be:

- `1 second` to `119 seconds`
- `2 minutes` to `179 minutes`
- `3 hours` to `47 hours`
- `2 days` to `99 days`

This diff is friendly towards i18n. German example:

Capture

As a failsafe, always use absolute time if the browser doesn't support `Math.trunc` and `Intl.RelativeTimeFormat`. For example, IE11.

### DIFF
--- a/resources/common.js
+++ b/resources/common.js
@@ -212,12 +212,32 @@ function count_down(label) {
     }, 1000);
 }
 
-function register_time(elems, limit) {
-    limit = limit || 300;
+function set_date_locale(language_code) {
+    if (typeof Intl !== 'undefined' && !!Intl.RelativeTimeFormat && !!Math.trunc) {
+        var rtf = new Intl.RelativeTimeFormat(language_code);
+        window.format_ms = function (amount) {
+            amount = Math.trunc(amount / 1000); // seconds
+            if (Math.abs(amount) < 120) return rtf.format(amount, 'second');
+            amount = Math.trunc(amount / 60);   // minutes
+            if (Math.abs(amount) < 180) return rtf.format(amount, 'minute');
+            amount = Math.trunc(amount / 60);   // hours
+            if (Math.abs(amount) < 48) return rtf.format(amount, 'hour');
+            amount = Math.trunc(amount / 24);   // days
+            if (Math.abs(amount) < 100) return rtf.format(amount, 'day');
+            return '';          // beyond 100 days, use absolute time
+        };
+    } else {
+        window.format_ms = function () {
+            return '';
+        };
+    }
+}
+
+function register_time(elems) {
     elems.each(function () {
         var outdated = false;
         var $this = $(this);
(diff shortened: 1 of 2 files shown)


int-y1:
Fun fact: `Intl.RelativeTimeFormat` supports fractional amounts, e.g. `rtf.format(-3.14159, 'minute')` becomes `'3.142 minutes ago'`.

I only used integer amounts in this pr, because there's no need for fractional amounts right now.

Xyene:
> 2 minutes to 179 minutes

👀

resources/common.js
```diff
@@ -212,24 +212,46 @@ function count_down(label) {
...
+            if (Math.abs(amount) < 120) return rtf.format(amount, 'second');
+            amount = Math.trunc(amount / 60);   // minutes
+            if (Math.abs(amount) < 180) return rtf.format(amount, 'minute');
+            amount = Math.trunc(amount / 60);   // hours
+            if (Math.abs(amount) < 48) return rtf.format(amount, 'hour');
+            amount = Math.trunc(amount / 24);   // days
+            if (Math.abs(amount) < 100) return rtf.format(amount, 'day');
+            return '';          // beyond 100 days, use absolute time
+        };
+        window.format_ms(0);    // check for errors
```
Xyene, line 229: Should the check for support be something like

```js
typeof Intl !== "undefined" && !!Intl.RelativeTimeFormat
```

so we don't rely on errors?

templates/base.html
```diff
@@ -79,12 +79,11 @@
         <script src="{{ static('libs/jquery-cookie.js') }}"></script>
         <script src="{{ static('libs/jquery-taphold.js') }}"></script>
         <script src="{{ static('libs/jquery.unveil.js') }}"></script>
-        <script src="{{ static('libs/moment.js') }}"></script>
```
Ninjaclasher, line 82: Since we no longer need `moment.js`, we should probably delete https://github.com/DMOJ/site-assets/blob/master/moment.js

resources/common.js
int-y1: Done. Chrome and Edge go to the `if` block. IE11 goes to the `else` block.

Xyene (approved):
LGTM, thanks!

PR #1991, opened 2022-08-27
Fix pwned passwords throwing `ValueError`

Ninjaclasher:
Relevant upstream issue: https://github.com/ubernostrum/pwned-passwords-django/issues/35


int-y1 (changes requested):
There is no reason to call `int()`.

I think `results` can be a set. Please change it to a set to simplify the code.

Ninjaclasher:
This code is copied from https://github.com/ubernostrum/pwned-passwords-django. I think we should match upstream, but other opinions are welcome.

kiritofeng (approved):
LGTM, thanks!

PR #2005, opened 2022-10-06
Implement disabling a judge

WallE256:
This feature removes a disabled judge from the queue handling, such that normal or rejudged submissions will not be allocated to it for grading. However, it will still receive submissions intended for it specifically.


judge/bridge/judge_list.py
```diff
@@ -52,6 +52,9 @@ def _handle_free_judge(self, judge):
                         break
                 node = node.next
 
+    def count_not_disabled(self):
+        return len(self.judges) - sum(judge.is_disabled for judge in self.judges) > 1
```
kiritofeng, line 56: You're treating this like an `int` on line 36, but returning a `bool` here

judge/models/runtime.py
```diff
@@ -147,6 +149,13 @@ def disconnect(self, force=False):
 
     disconnect.alters_data = True
 
+    def toggle_disabled(self):
+        self.is_disabled = not self.is_disabled
+        self.save(update_fields=['is_disabled'])
+        disconnect_judge(self, force=True)
```
kiritofeng, line 155: Should this send a forced disconnect if you're re-enabling the judge?

In particular, it looks like this function is used less like *toggle* disable, and more like *disable*.

kiritofeng (changes requested):

judge/bridge/judge_list.py
WallE256: Oh true, the `>1` must've slipped when writing the function.

judge/models/runtime.py
WallE256: The purpose of forcing a disconnect is to update the `is_disabled` cached property in the bridge, which helps to not access the Judge DB object for every submission and queue removal.

kiritofeng (approved):
Tentative LGTM, would like @quantum5 or @Xyene to look over it before merging though.

Xyene (commented):
LGTM, but it looks like there are merge conflicts.

judge/models/runtime.py
Xyene: I think there should be a bridge RPC added rather than hacking around it this way. The judge we run on dmoj.ca will auto-reconnect, but it's not guaranteed that all judges will.

judge/bridge/judge_list.py
```diff
@@ -52,6 +52,9 @@ def _handle_free_judge(self, judge):
                         break
                 node = node.next
 
+    def count_not_disabled(self):
+        return len(self.judges) - sum(judge.is_disabled for judge in self.judges)
```
quantum5, line 56:
```suggestion
        return sum(not judge.is_disabled for judge in self.judges)
```

templates/admin/judge/judge/change_form.html
```diff
@@ -22,5 +23,10 @@
             <i class="fa fa-lg fa-plug"></i>
             <span class="text">{% trans "Terminate" %}</span>
         </a>
+        <a style="display: none" title="{% trans "Disable" %}" href="{% url 'admin:judge_judge_disable' original.pk %}"
+           class="button disable-link">
+            <i class="fa fa-lg fa-ban"></i>
+            <span class="text">{% trans "Disable" %}</span>
```
quantum5, line 29: Don't know how I feel about this being a toggle but always showing `Disable`.

judge/models/runtime.py
WallE256: Added a bridge RPC instead of the disconnect method.

templates/admin/judge/judge/change_form.html
WallE256: This has been addressed.

judge/bridge/django_handler.py
```diff
@@ -18,6 +18,7 @@ def __init__(self, request, client_address, server, judges):
             'submission-request': self.on_submission,
             'terminate-submission': self.on_termination,
             'disconnect-judge': self.on_disconnect_request,
+            'disable-judge': self.update_disable_judge,
```
Xyene, line 21:
```suggestion
            'disable-judge': self.on_disable_judge,
```

for consistency

judge/models/runtime.py
```diff
@@ -147,6 +149,13 @@ def disconnect(self, force=False):
 
     disconnect.alters_data = True
 
+    def toggle_disabled(self):
+        self.is_disabled = not self.is_disabled
+        self.save(update_fields=['is_disabled'])
```
Xyene, line 154: I think this `save` should come after the RPC, so we don't end up (in case of an error) with a bridge that does not agree in state with the site.

templates/admin/judge/judge/change_form.html
```diff
@@ -22,5 +23,18 @@
             <i class="fa fa-lg fa-plug"></i>
             <span class="text">{% trans "Terminate" %}</span>
         </a>
+        {% if not original.is_disabled %}
+        <a style="display: none" title="{% trans "Disable" %}" href="{% url 'admin:judge_judge_disable' original.pk %}"
```
Xyene, line 27: Missing indentation within the `if`/`else`.

PR #2007, opened 2022-10-10
Add `linematches` checker to site-managed problems.

kiritofeng:
Site-side component of DMOJ/judge-server#844


quantum5 (approved):

kiritofeng:
Giving up for now, because I realized this is a lot harder than I gave it credit for.

Issue #1881, opened 2022-03-01
Fix rendering of target colour


kiritofeng:
Target colour is broken on:
- [ ] Tickets
- [ ] Contest scoreboards

On the former, this is because the image is included in the `a`, and on the latter, this is because the `display: inline-block` is being overrode.

PR #2008, opened 2022-10-11
Change how username is displayed in tickets

int-y1:
Partially do #1881.

* In existing message, move username from gravatar image to info header. This looks similar to comments.
* In new message tool, remove username from gravatar.

Capture


Xyene (approved):
LGTM, thanks!

PR #2009, opened 2022-10-11
Remove reply button on old comments

### LINKED ISSUE #823: Block commenting on old threads
It's pretty ridiculous when people bump up threads from years ago, adding nothing of value to the conversation.

int-y1:
fixes #823

* remove reply button when comment is too old and user doesn't have `judge.change_comment`.
* users can now reply to their own comment (previously, the backend allowed it, but the frontend didn't show the reply button)

Capture


Riolku:
Do other people have opinions on this being a pure UI change?

int-y1:
> It looks to be a trivial change in the backend, so let's do it instead of letting people deliberately necropost to "show their 1337 h4x0r skills".

done

dmoj/settings.py
```diff
@@ -85,6 +85,7 @@
 DMOJ_USER_DATA_CACHE = ''
 DMOJ_USER_DATA_DOWNLOAD_RATELIMIT = datetime.timedelta(days=1)
 DMOJ_COMMENT_VOTE_HIDE_THRESHOLD = -5
+DMOJ_COMMENT_REPLY_TIMEFRAME = datetime.timedelta(days=1000)
```
Riolku, line 88: Maybe 365 is better because it's less arbitrary.

judge/comments.py
```diff
@@ -78,10 +79,14 @@ def post(self, request, *args, **kwargs):
...
+                return HttpResponseBadRequest()
+            try:
+                parent_comment = Comment.objects.get(hidden=False, id=parent, page=page)
+            except Comment.DoesNotExist:
                 return HttpResponseNotFound()
-            else:
-                if not Comment.objects.filter(hidden=False, id=parent, page=page).exists():
-                    return HttpResponseNotFound()
+            if not (self.request.user.has_perm('judge.change_comment') or
+                    parent_comment.time > timezone.now() - settings.DMOJ_COMMENT_REPLY_TIMEFRAME):
```
Riolku, line 88: I might be the only one, but I feel like

```
if timezone.now() - parent_comment.time > settings.DMOJ_COMMENT_REPLY_TIMEFRAME and not self.request.user.has_perm('judge.change_comment'):
```

Is the more readable way of writing this.

templates/comments/list.html
```diff
@@ -66,39 +66,32 @@ <h2><i style="padding-right: 0.3em" class="fa fa-comments"></i>{{ _('Comments')
                                 {% else %}
                                     <span class="comment-edits"></span>
                                 {% endif %}
-                                    <a href="#comment-{{ node.id }}" title="{{ _('Link') }}" class="comment-link">
-                                    <i class="fa fa-link fa-fw"></i>
-                                </a>
+                                    <a href="#comment-{{ node.id }}" title="{{ _('Link') }}"
+                                       class="comment-link"><i class="fa fa-link fa-fw"></i></a>
```
Riolku, line 70: I think `</a>` should be on its own line.

judge/comments.py
```diff
@@ -118,5 +123,6 @@ def get_context_data(self, **kwargs):
             context['is_new_user'] = not self.request.user.is_staff and not profile.has_any_solves
         context['comment_list'] = queryset
         context['vote_hide_threshold'] = settings.DMOJ_COMMENT_VOTE_HIDE_THRESHOLD
+        context['reply_timeframe'] = timezone.now() - settings.DMOJ_COMMENT_REPLY_TIMEFRAME
```
Riolku, line 126: Let's call this something like `can_reply_before`.

Riolku (commented):
Overall looks fine.

templates/comments/list.html
int-y1: `</a>` appears multiple times, and moving each `</a>` into a new line hurts readability

judge/comments.py
int-y1: `{% if node.time > can_reply_before %}` is too confusing for me

dmoj/settings.py
int-y1: i should collect statistics to see if 365 is a good cutoff

judge/comments.py
Riolku: what about `reply_cutoff`

int-y1: sure

kiritofeng (approved):
LGTM, thanks!

Xyene (approved):
LGTM, thanks!

PR #2010, opened 2022-10-12
Confirm rejudging on submission page

Riolku:
If a contest sub is rejudged and the verdict doesn't change, did the organizer still mess up?

### DIFF
--- a/templates/submission/status.html
+++ b/templates/submission/status.html
@@ -74,10 +74,19 @@
         <div><a href="{{ url('problem_submit', submission.problem.code, submission.id) }}">{{ _('Resubmit') }}</a></div>
     {% endif %}
     {% if perms.judge.rejudge_submission and not submission.is_locked %}
+        {% compress js %}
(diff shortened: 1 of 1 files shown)


templates/submission/status.html
```diff
@@ -74,10 +74,19 @@
         <div><a href="{{ url('problem_submit', submission.problem.code, submission.id) }}">{{ _('Resubmit') }}</a></div>
     {% endif %}
     {% if perms.judge.rejudge_submission and not submission.is_locked %}
+        {% compress js %}
```
quantum5, line 77: Why are we shoving this into a separate js file?

Riolku: Where should it go? The JS blocks in this file are in unrelated if blocks.

Xyene: Eh, seems fine, it'll get compressed.

Xyene (approved):
LGTM, thanks!

Issue #1572, opened 2020-11-03
color test case AC differently on /submission page when AC does not award full marks


xiaowuc1:
This can parallel the different coloring on /submissions.

PR #1802, opened 2021-10-14
color test case AC differently on /submission page when AC does not a…

kiritofeng:
…ward full marks; fixes #1572

PR #2021, opened 2022-10-31
Highlight partial AC testcases differently

Xyene:
Supersedes https://github.com/DMOJ/online-judge/pull/1802, fixes https://github.com/DMOJ/online-judge/issues/1572.

image

### DIFF
--- a/resources/submission.scss
+++ b/resources/submission.scss
@@ -288,7 +288,7 @@ label[for="language"], label[for="status"] {
 }
 
 .case-_AC {
-    color: red;
+    color: #BBCC00;
     font-weight: bold;
 }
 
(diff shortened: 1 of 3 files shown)


resources/submission.scss
```diff
@@ -288,7 +288,7 @@ label[for="language"], label[for="status"] {
 }
 
 .case-_AC {
```
quantum5, line 290: Why did this even exist in the first place?

quantum5 (approved):

quantum5:
Actually the colour may be a bit hard to see on a white background.

kiritofeng:
image
Agree with @quantum5, maybe we should pick something closer to the full AC colour

Xyene:
You're late, I've already updated the PR with a @quantum5-approved color.

PR #2023, opened 2022-10-31
Replace unique_together_left_join with FilteredRelation

quantum5:
Django supports `FilteredRelation` now, so there is no need to use that ugly hack.


Xyene (approved):
LGTM, thanks!

PR #2030, opened 2022-11-01
Fix submission table for 0 submissions

int-y1:
In the old code, `obj.activity / max_activity` can divide by 0, and break the ui.

New screenshot:

Capture


Xyene (approved):
LGTM, thanks!

PR #2032, opened 2022-11-03
UI: Button to Toggle Password Visiblity in login page

### LINKED ISSUE #1816: Add a button to check the entered password.

Diwakar-Gupta:
this pr fixes #1816 

Screenshot 2022-11-04 at 9 39 17 AM

Screenshot 2022-11-04 at 9 39 53 AM

### DIFF
--- a/templates/registration/login.html
+++ b/templates/registration/login.html
@@ -48,6 +48,22 @@
                     <td>{{ form.password }}
                     </td>
                 </tr>
+                <tr>
+                    <th></th>
+                    <th>
+                        <input type="checkbox" onclick="togglePasswordVisiblity()">Show Password
+                        <script>
+                            function togglePasswordVisiblity(){
+                                var source = document.getElementsByName('password')[0];
+                                if (source.type === "password") {
+                                    source.type = "text";
+                                } else {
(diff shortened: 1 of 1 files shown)


templates/registration/login.html
```diff
@@ -48,6 +48,22 @@
...
+                        <script>
+                            function togglePasswordVisiblity(){
+                                var source = document.getElementsByName('password')[0];
+                                if (source.type === "password") {
+                                    source.type = "text";
+                                } else {
+                                    source.type = "password";
+                                }
+                            }
+                        </script>
```
quantum5, line 64: We do not write JavaScript like this, especially with `onclick`.

quantum5 (changes requested):
This feature is almost certainly a bad idea. Any mistake, and the password could be remembered as text. Here, if you login with the box checked, the browser most certainly will remember it for autocompletion as normal text. While this is possible to fix, I simply do not trust any fix to always work.

In conclusion, this feature shouldn't exist, and even if it does, it should be the browser's responsibility.

PR #2036, opened 2022-11-14
Clean up ContestCalendar css

int-y1:
* `weekday_classes` is useless. It only affects `<th>`, but no `<th>` uses it.
* No need for `border-right/left/bottom`. `border` is enough because of `border-collapse: collapse;`.

New version of ContestCalendar:

Capture


Xyene (approved):
LGTM, thanks!

PR #2038, opened 2022-11-14
Clean up page tab's css

int-y1:
* `class="page-title"` is dead (and I have no idea why its css was in `templates/user/user-base.html`)
* `<span class="spacer"></span>` became redundant due to `justify-content: space-between;`
* Replace `tabs-no-flex` nonsense with `flex-wrap: wrap;`

The page tabs in `/contests/2022/11/` should behave a little more nicely now.

Example of `flex-wrap` doing something: Capture

### DIFF
--- a/resources/base.scss
+++ b/resources/base.scss
@@ -498,11 +498,6 @@ noscript #noscript {
     display: inline-block;
 }
 
-.spacer {
-    display: inline-block;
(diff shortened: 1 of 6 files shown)


Xyene (approved):
LGTM, thanks!

PR #2039, opened 2022-11-15
Don't allow for infinite and nan values in `safe_float_or_none`

kiritofeng:
Thanks to @int-y1 for finding this bug.

### DIFF
--- a/judge/utils/strings.py
+++ b/judge/utils/strings.py
@@ -1,3 +1,6 @@
+from math import isfinite
+
+
 def safe_int_or_none(value):
     try:
         return int(value)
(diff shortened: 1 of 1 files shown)


int-y1 (approved):

judge/utils/strings.py
```diff
@@ -1,12 +1,19 @@
...
 
-def safe_float_or_none(value):
+def safe_float_or_none(value, force_finite=True):
     try:
-        return float(value)
+        num = float(value)
+        if force_finite and not isfinite(num):
+            return None
+
+        return num
```
quantum5, line 17: Move these out of the `try`.

quantum5 (approved):
Get rid of the trailing period in the commit+PR title.

Issue #2035, opened 2022-11-13
Dark mode


int-y1:
This is for the many users who requested dark mode. This issue is possibly a subset of #578.

* Clean up `resources/vars.scss`. Variable names like `$background_light_gray` are not acceptable.
* Images need `background-color: white;` because the image may have transparency.
* Find page-specific `<style>` with `git grep -n "<style"`. At a minimum, migrate the color-specific properties into scss file(s). (Warning: this scss migration may need a lot of pr's)

int-y1:
This issue has reached a milestone.

The site will serve dark mode if the user specifies `?dark=1` in the URL. For example: https://dmoj.ca/problem/aplusb?dark=1
This new feature is intended for developers who want to test in prod, and work on this issue.

Xyene:
* [ ] SVG/PNG math isn't color-inverted, though MathJax is. The PNG may fall under image color inversion.

Xyene:
* [ ] User profile images should have a background set, see e.g. comments on https://dmoj.ca/problem/fibonacci

jdabtieu:
This might be a little nitpicky, but purple text can be hard to read against the dark background
![image](https://user-images.githubusercontent.com/62577178/210495219-2b0069d0-7c39-48b6-9d47-acd955a34ea3.png)
Lightening the color in dark mode to something like `#a600a6` (from `#800080`) would help
![image](https://user-images.githubusercontent.com/62577178/210495616-89125b4c-d41a-4bbb-9d12-ffa5d7c9607c.png)
Same for blue users, though I don't have a better suggestions for it
![image](https://user-images.githubusercontent.com/62577178/210495399-d98588dc-cf2d-46e9-bef9-d2bc726e14a1.png)

Xyene:
Here's a CSS export of pygment's `github-dark` theme: https://gist.github.com/Xyene/e0cac053b28507cd5e65478ae44eba22

This could be used in place of https://github.com/DMOJ/online-judge/blob/master/resources/pygment-github.css

![image](https://user-images.githubusercontent.com/1403503/211182855-e1c80da5-10b7-476a-a827-c482d500ea9d.png)

Xyene:
* [ ] Grading / processing submission status rows are unreadable

![Screenshot_20230108-162850~2](https://user-images.githubusercontent.com/1403503/211220106-253f1fa6-3002-4036-a8df-70c9b17eca76.png)
#

[...]

PR #2041, opened 2022-11-22
Introduce primary color in scss

int-y1:
Part of #2035. Replace color names in `vars.scss` with names convenient for dark mode. The convention I'm using is `$color_<base name><percentage>`.

In a future pr, migrate hardcoded grays to `$color_primaryXX`.


Xyene (approved):
LGTM, thanks!

PR #2043, opened 2022-11-27
Standardize SCSS gray colors

int-y1:
Part of #2035.

* Update scss files with the proper grays. Lots of grays were rounded to the nearest `$color_primaryXX`.
* Don't invert colors in `nav`. (navbar's colors are already ok for dark mode)
* This pr shouldn't break anything. The original theme should continue to work.

Inverting the `$color_primaryXX` results in:
Capture

There are a few styling errors. Probably good enough for now.

### DIFF
--- a/resources/base-description.scss
+++ b/resources/base-description.scss
@@ -17,7 +17,7 @@
 
     h1, h2, h3, h4, h5, h6 {
         font-weight: normal;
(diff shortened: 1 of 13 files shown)


Xyene (approved):
LGTM, thanks!

PR #2044, opened 2022-12-05
Clean up widget scss

int-y1:
Loosely related to #2035

* in `/contest/asdf`, fix style of facebook icon.
* clean up button styles. note that `:hover` is redundant because of line 47.
* in `/accounts/register/`, fix style of email field.
* remove redundant styles in `ul.pagination` (technically, `margin-left: 0` is in use, but it causes a 2px border, so i'm removing it)
* select2 uses a white bg, so add `color: black`

### DIFF
--- a/resources/widgets.scss
+++ b/resources/widgets.scss
@@ -13,7 +13,7 @@
     color: #55ACEE;
 }
 
-.facebook-this it {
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks!

PR #2045, opened 2022-12-05
Remove dead AdminPagedownWidget

int-y1:
Admin migrated from markdown to martor, so the `AdminPagedownWidget` classes became dead. They're also dead according to `git grep -n "AdminPage[Dd]ownWidget"`.

Clean up `make_style.sh` and `.gitignore`.

(the code has both `Pagedown` and `PageDown`, but this inconsistency isn't relevant)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,5 @@
 *.swp
 dmoj/local_settings.py
 resources/style.css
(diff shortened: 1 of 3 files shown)


Xyene (approved):
LGTM, thanks!

PR #2048, opened 2022-12-06
Migrate text editor css to scss

int-y1:
Part of #2035. When `$color_primaryXX` is inverted, text editors will look like this:

Capture

(the new scss files are super untidy, but that's a problem for another time)


Xyene (approved):
This LGTM, thanks. As an aside, we really should switch the comment editor to use martor at some point, so we don't have to maintain two sets of stylesheets...

PR #2050, opened 2022-12-10
Fix editor buttons when replying to comment

### LINKED ISSUE #1486: Markdown helpers don't work when replying to a comment.
None of these buttons work when replying to a comment:

image

This only happens when replying to comments.

int-y1:
i.e. change enough ids to make `window.DjangoPagedown.createEditor` happy

also, this js looks even more trash than before 🚮

fixes #1486


Xyene (approved):
♻️

PR #2051, opened 2022-12-10
Make problem search more sane

int-y1:
fixes these 2 issues:

1. go to `/problems/`. search for `a20`. get 1 result. clear `a20` and search again. see `point_start=5.0&point_end=5.0` in querystring.
2. go to `/problems/`. select all 4 checkboxes. click "PROBLEMS" in site header. see "Has editorial" get unselected.

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -443,8 +443,9 @@ class ProblemList(QueryStringSortMixin, TitleMixin, SolvedProblemMixin, ListView
         if not points:
             return 0, 0, {}
         if len(points) == 1:
-            return points[0], points[0], {
+            return points[0] - 1, points[0] + 1, {
                 'min': points[0] - 1,
+                '50%': points[0],
                 'max': points[0] + 1,
             }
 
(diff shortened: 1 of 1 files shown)


judge/views/problem.py
```diff
@@ -443,8 +443,9 @@ def get_noui_slider_points(self):
         if not points:
             return 0, 0, {}
         if len(points) == 1:
-            return points[0], points[0], {
+            return points[0] - 1, points[0] + 1, {
```
Xyene, line 446: Why does this fix things?

int-y1: If the problem list only has 5p problems, the server will give `point_start=5` and `point_values.min=4`. This runs into trouble with

https://github.com/DMOJ/online-judge/blob/97f2c450e88814012d19cbcfbec3ae74d14415cb/templates/problem/list.html#L121

Disabled inputs aren't put in the querystring.

Same goes for `point_end=5` and `point_values.max=6`.

Xyene (approved):
👍

PR #2053, opened 2022-12-11
Add querystring option to enable dark mode

int-y1:
part of #2035. this enables dark mode whenever the user adds `?dark=1` to the url. this makes it easy to investigate dark mode on the live site.

1. run `./make_style.sh`
2. visit `/problem/aplusb?dark=1`
3. despair at the page-specific css (see h2 title, see comments)

Capture


make_style.sh
```diff
@@ -14,10 +14,14 @@ if ! [ -x "$(command -v autoprefixer)" ]; then
   exit 1
 fi
 
-FILES=(sass_processed/style.css sass_processed/martor-description.css)
-
 cd "$(dirname "$0")" || exit
+
+echo 'Creating default style...'
+cp resources/vars_default.scss resources/vars.scss
```
Xyene, line 20: Could we write this as a for loop? There's a lot in common between the light/dark themes.

resources/vars_dark.scss
```diff
@@ -0,0 +1,25 @@
...
+$color_link75: #38e;    // default
+$color_link100: #59f;   // hover
+$color_link200: #960;   // active
+
+$color_pageBg: #222;    // #222 because some elements should be darker than pageBg
+
+$highlight_blue: #2980B9;
+$announcement_red: #ae0000;
+
+$base_font_size: 14px;
```
Xyene, line 20: Could we have a `vars_common` for these?

make_style.sh
int-y1: Done. The generated files are now:

```
resources/style.css
resources/martor-description.css (used for admin's martor preview)
resources/dark/style.css
resources/dark/martor-description.css (useless, because admin doesn't do dark mode)
```

resources/vars_dark.scss
int-y1: Done.

resources/vars_dark.scss
```diff
@@ -0,0 +1,17 @@
+@import "vars_common";
```
Xyene, line 1: Nit: use - instead of _ for consistency with other existing scss files.

int-y1: Done.

.gitignore
```diff
@@ -7,6 +7,8 @@
 *~
 *.swp
 dmoj/local_settings.py
-resources/style.css
+resources/dark
 resources/martor-description.css
+resources/style.css
+resources/vars.scss
```
Xyene, line 13: Why this line?

int-y1: Because of `cp resources/vars-$1.scss resources/vars.scss`. Also, the user can kill the script while `vars.scss` exists, and adding `vars.scss` to the gitignore covers this case.

Xyene (approved):
LGTM, thanks!

PR #2055, opened 2022-12-12
Allow unicode and nobigmath hints to be used with site data

FherStk:
A new field called "binary data" has been added to the "problem_data" model, so this field will be stored into the "init.yaml" file when storing the test cases for a problem.


FherStk (commented):
Hints also added (code including accents will work properly).

judge/migrations/0134_problemdata_hints.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.2.16 on 2022-12-12 16:25
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0133_auto_20221212_1427'),
```
kiritofeng, line 9: You'll need to rename this now

judge/models/problem_data.py
```diff
@@ -30,6 +30,11 @@ def problem_directory_file(data, filename):
     ('linecount', _('Line-by-line')),
 )
 
+HINTS = (
+    ('unicode', _('Unicode')),
+    ('none', _('NONE'))
```
kiritofeng, line 35: We also have `nobigmath` (https://github.com/DMOJ/java-sandbox-agent); `nobuf` isn't necessary because there's a language-agnostic `unbuffered` option that should be used instead.

judge/migrations/0133_add_binary_data.py
```diff
@@ -0,0 +1,457 @@
...
+import django.core.validators
+import django.db.models.deletion
+from django.db import migrations, models
+
+import judge.models.contest
+import judge.models.problem
+import judge.models.profile
+
+
+class Migration(migrations.Migration):
```
kiritofeng, line 12: @Xyene @quantum5 we need to do the cleanup migration at some point, should we write a separate migration or just hide it here?

judge/migrations/0134_problemdata_hints.py
FherStk: Done!

Xyene:
Thanks for the contribution. While I'm generally in favour of exposing this functionality in the site interface, I'd rather we didn't expose it as `hints` directly. I think the implementation of hints on the judge side was a mistake and these should have just been fields, e.g.,

```yaml
require_unicode_output: true
disable_bigmath_libraries: true
```

instead of the more arcane

```yaml
hints: [unicode, nobigmath]
```

We might still go back and clean this up on the judge side at some point.

With that in mind, I think we should be exposing three boolean fields in this PR, and only internally outputting the hints field.

quantum5:
Yes, these should be boolean fields. Also, since binary data does not solve your problem, I would rather that one is not exposed unless the need truly arises.

FherStk:
Pull request updated :)

.gitignore
```diff
@@ -10,3 +10,8 @@ dmoj/local_settings.py
 resources/style.css
 resources/martor-description.css
 sass_processed
+dmojsite
+dmojsite/*
+problems
+problems/*
+*.log
```
quantum5, line 19: These lines seem specific to your install. It's recommended to not store unrelated files inside the repository anyways. Would remove this from the PR.

FherStk:
Sorry, I forgot to pull from master, my branch is now updated and so the pull request.

judge/models/problem_data.py
```diff
@@ -91,4 +92,4 @@ class ProblemTestCase(models.Model):
     output_limit = models.IntegerField(verbose_name=_('output limit length'), blank=True, null=True)
     checker = models.CharField(max_length=10, verbose_name=_('checker'), choices=CHECKERS, blank=True)
     checker_args = models.TextField(verbose_name=_('checker arguments'), blank=True,
-                                    help_text=_('Checker arguments as a JSON object.'))
+                                    help_text=_('Checker arguments as a JSON object.'))
```
quantum5, line 95: Can you add the trailing `\n` so that git doesn't complain?

judge/utils/problem_data.py
```diff
@@ -178,10 +180,18 @@ def make_checker(case):
...
+        if self.data.unicode:
+            hints.append("unicode")            
+        if self.data.nobigmath:
+            hints.append("nobigmath")                     
         if self.data.checker:
             init['checker'] = make_checker(self.data)
         else:
             self.data.checker_args = ''
+
+        if len(hints) > 0:
```
quantum5, line 192:
```suggestion
        if hints:
```

quantum5 (commented):
Can we squash all the migrations together? Otherwise LGTM except for the minor nits.

.gitignore
FherStk: Sure! Removing.

judge/models/problem_data.py
FherStk: Sorry, do you mean at the end of the file?

quantum5: Yes.

judge/utils/problem_data.py
FherStk: Done!

judge/models/problem_data.py
FherStk: Done!

FherStk:
PR updated :)

.gitignore
```diff
@@ -11,4 +11,4 @@ resources/dark
 resources/martor-description.css
 resources/style.css
 resources/vars.scss
-sass_processed
+sass_processed
```
quantum5, line 14: Can we not change this file at all?

judge/migrations/0133_add_binary_data.py
```diff
@@ -0,0 +1,457 @@
+# Generated by Django 3.2.16 on 2022-12-12 14:27
```
quantum5, line 1: Can we squash all three migrations?

quantum5 (commented):
Thanks for making the changes! Unfortunately, there are some linter errors and the migrations still need to be squashed.

.gitignore
FherStk: I apologize, downloading the original file again.

judge/migrations/0133_add_binary_data.py
FherStk: Sorry, it was my first squash. Done!

FherStk:
PR updated again, thanks @quantum5 for your infinite patience with me :)

quantum5 (commented):
Changes look good, but you are still failing the linter. Otherwise it looks good.

FherStk:
I'm not sure about how to proceed, should I fix this?
image

quantum5:
> I'm not sure about how to proceed, should I fix this? image

Yes. Once you are done, you should run flake8 locally to make sure all errors disappear.

FherStk:
Done, I've installed fake8 locally with all the needed plugins: no errors on my local side.

quantum5 (approved):

judge/migrations/0133_add_binary_data.py
FherStk: Done!

PR #2068, opened 2022-12-17
Create dark mode for Featherlight

int-y1:
Part of #2035. Move `featherlight.min.css` and inline css into a single scss file.

Old:
Capture

New:
Capture

### DIFF
--- a/resources/featherlight.scss
+++ b/resources/featherlight.scss
@@ -6,7 +6,7 @@
     display: none;
     position: fixed;
     top: 0; right: 0; bottom: 0; left: 0;
-    z-index: 2147483647;
+    z-index: 1000;
     text-align: center;
     white-space: nowrap;
     cursor: pointer;
--- a/resources/featherlight.scss
+++ b/resources/featherlight.scss
@@ -1,0 +1,91 @@
+@import "vars";
+
+// Based on Featherlight 1.7.14
+
+.featherlight {
+    display: none;
+    position: fixed;
+    top: 0; right: 0; bottom: 0; left: 0;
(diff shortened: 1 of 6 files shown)


resources/featherlight.scss
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    &:before {
+        content: '';
+        display: inline-block;
+        height: 100%;
+        vertical-align: middle;
+    }
+
+    & .featherlight-content {
```
quantum5, line 26: Why do you need `& .`?

int-y1: Featherlight used `.featherlight .featherlight-content`, and I wrote it here as `& .featherlight-content`.

quantum5: This is redundant. `.featherlight-content` works just fine. It nests by default and `&` is only needed when it's on the same level.

int-y1: Ok, cleaned up.

resources/featherlight.scss
```diff
@@ -0,0 +1,91 @@
+@import "vars";
+
+// Based on Featherlight 1.7.14
+
+.featherlight {
+    display: none;
+    position: fixed;
+    top: 0; right: 0; bottom: 0; left: 0;
+    z-index: 2147483647;
```
quantum5, line 9:
```suggestion
    z-index: 1000;
```

Let's not go straight to the maximum. What if we wanted to display something else on top?

quantum5 (approved):
Doesn't seem broken.

PR #2071, opened 2022-12-18
Use striped tables

int-y1:
Increase the number of striped tables from 5 to 16. List of affected tables:

1. `/problem/aplusb` > comment > view votes (needs comment moderation perms)
2. `/contest/<key>` > Problems (also, I deleted `#contest-problems` because there's no css for it)
3. `/contest/<key>/moss` (requires moss results)
4. `/organization/<slug>` > class table (requires at least 1 class)
5. `/organizations/`
6. `/status/`
7. `/runtimes/`
8. `/tickets/`
9. `/users/`
10. `/user/<username>/solved` author table
11. `/user/<username>/solved` problem group table

### DIFF
--- a/templates/comments/votes.html
+++ b/templates/comments/votes.html
@@ -1,5 +1,5 @@
 <h2 style="text-align: center">{{ _('Votes') }}</h2>
-<table class="table" style="display: table">
+<table class="table striped">
     <thead>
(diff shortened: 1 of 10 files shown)


Xyene (approved):
LGTM, thanks!

PR #2072, opened 2022-12-19
Remove IE8 code

int-y1:


quantum5 (approved):
Cloudflare seems to have banned IE8 anyways...

PR #2079, opened 2022-12-27
Change style for bad user mentions

int-y1:
Capture

Side note: `<del>` is also used by Markdown's `~~strikethrough~~`.

### DIFF
--- a/judge/jinja2/reference.py
+++ b/judge/jinja2/reference.py
@@ -18,7 +18,7 @@ rereference = re.compile(r'\[(r?user):(\w+)\]')
 
 def get_user(username, data):
     if not data:
-        element = Element('span')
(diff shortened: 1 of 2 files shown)


judge/jinja2/reference.py
```diff
@@ -18,8 +18,10 @@
 
 def get_user(username, data):
     if not data:
-        element = Element('span')
-        element.text = username
+        element = Element('del')
```
Xyene, line 21: Could we implement this as a CSS class rather a tag?

int-y1: Done. Created `.deleted-user`, and checked that the screenshot is still accurate.

Xyene (approved):
LGTM, thanks!

PR #2080, opened 2022-12-27
Fix bridge database connection leaks

quantum5:


kiritofeng (approved):
Thank you for firefighting.

PR #2083, opened 2022-12-27
Fix CSS for rating circles

int-y1:
* `/users/?dark` Support dark mode for the rating circle in the table header.
* `/contest/<key>/ranking/` Vertically align the rating circle.


Xyene (approved):
LGTM, thanks!

PR #2085, opened 2022-12-28
Replace deprecated `django.conf.url` with `django.urls.path`

kiritofeng:


quantum5 (commented):
Literally everything here can be replaced with `path`, so let's use that.

Xyene (approved):
LGTM, thanks, will defer merging to @quantum5.

quantum5 (approved):

PR #2086, opened 2022-12-28
Fix blank submission activity colour

Ninjaclasher:
Part of #2035. Non-existent days (used only for aligning with day of the week) should match the page background colour.

Before:
image

After:
image


Xyene (approved):
LGTM, thanks!

PR #2089, opened 2022-12-30
Fix the check for has_any_solves

int-y1:
~~I have no idea if I should keep `@cached_property`. Or delete `@cached_property`. Or delete all of `has_any_solves()`, and replace every `profile.has_any_solves` with `profile.problem_count > 0`.~~

Change the check to result = AC, case_points >= case_total.

Capture
Capture

### DIFF
--- a/judge/comments.py
+++ b/judge/comments.py
@@ -46,8 +46,7 @@ class CommentForm(ModelForm):
             if profile.mute:
                 raise ValidationError(_('Your part is silent, little toad.'))
             elif not self.request.user.is_staff and not profile.has_any_solves:
-                raise ValidationError(_('You need to have solved at least one problem '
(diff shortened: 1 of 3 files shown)


Ninjaclasher (commented):
There is a logic change here -- with this change, organization-private problems do not count anymore.

Probably not too big of a deal, but something to keep in mind.

Xyene:
I think org-private solutions should continue to count, otherwise students of private orgs may not be able to ask questions in comments.

int-y1:
I updated the check. It searches through all the user's submissions.

Ninjaclasher (approved):

Xyene (approved):
LGTM, thanks!

PR #2090, opened 2022-12-31
Remove dead variables for rating chart

int-y1:
`max_graph` and `min_graph` are dead. Looks like they were intended for the rating chart in `/user/<username>`.


Xyene (approved):
LGTM, thanks!

PR #2093, opened 2023-01-02
Fix SVG/PNG math in dark mode

int-y1:
Part of #2035. This fixes `` math that use black text.

I got the relevant `` classes from [mathoid.py](https://github.com/DMOJ/online-judge/blob/6c5dd7bc217295870a60d2f6cc3b44b18de4e626/judge/utils/mathoid.py). (I ignored `output_msp` because I think this function is dead)

### DIFF
--- a/resources/base-description.scss
+++ b/resources/base-description.scss
@@ -15,7 +15,7 @@
         height: auto;
     }
 
-    img.tex-full {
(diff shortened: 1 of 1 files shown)


Xyene (approved):
LGTM, thanks!

PR #2094, opened 2023-01-02
Move ticket CSS into SCSS file

int-y1:
Part of #2035. A few notes:

* I saw `#44AD41` and `#DE2121` being used for other icons, so I used them here.
* I redid the ticket title.
* I changed these selectors: `#container`->`#ticket-list-container`, `.message`->`.ticket-message`, `.container`->`.ticket-container`

Before and after:
Capture

### DIFF
--- a/resources/ticket.scss
+++ b/resources/ticket.scss
@@ -1,0 +1,217 @@
+@import "vars";
+
+form#ticket-form {
+    display: block;
+    margin: 0 auto;
+    width: 100%;
+    max-width: 750px;
+    padding-top: 1em;
+
+    #id_title {
+        width: 100%;
+    }
+
+    .submit {
+        margin: 10px 0 0 auto;
+    }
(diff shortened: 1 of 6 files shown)


Xyene:
LGTM. While we're at it, can we swap the colors of close / open?

resources/ticket.scss
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    .submit {
+        margin: 10px 0 0 auto;
+    }
+}
+
+#ticket-list {
+    .fa-check-circle-o {
+        color: #44AD41;
```
Xyene, line 21: Maybe we should have a var in this file for this and the other color, since they're used twice.

Xyene: I don't feel strongly about this.

Xyene (approved):

PR #2096, opened 2023-01-02
Fix ticket form's width

int-y1:
`/problem/aplusb/tickets/new`
On narrow screens, `width: 100%;` causes overflow and looks bad.

### DIFF
--- a/resources/ticket.scss
+++ b/resources/ticket.scss
@@ -3,7 +3,6 @@
 form#ticket-form {
     display: block;
     margin: 0 auto;
-    width: 100%;
     max-width: 750px;
     padding-top: 1em;
 


Xyene (approved):
LGTM, but it looks like with this change + narrow screen, there is more padding on the left side than the right?

int-y1:
> LGTM, but it looks like with this change + narrow screen, there is more padding on the left side than the right?

This affects other forms too. Forms have 15px left pad, and 10px right pad, because of https://github.com/DMOJ/online-judge/blob/a5eb1820bda4719083d21d8037471127411cf88c/resources/base.scss#L86-L92

I suspect that this is a typo that wasn't corrected.

PR #2100, opened 2023-01-03
Clean up user icon in top-right corner

int-y1:
Delete `<header>` because it's dead. Clean up `#user-links` css.

Old version:
Capture

New version:
Capture


Xyene (approved):
LGTM, thanks!

PR #2101, opened 2023-01-03
Remove negative margins in comments

int-y1:
Biggest visual changes:

* Comment lock message was moved from the bottom to the top.
Capture
* Spacing between title and no comment message has changed from 5px to 10px.


Xyene (approved):
LGTM, thanks!

PR #2104, opened 2023-01-04
Fix small bug with problem attempt icon

int-y1:
This ensures that 100/100 WA shows up as a problem attempt.

`*_attempted_ids` are used in these 3 places:

* `/problems/` to display orange icon in table column
* `/problems/?order=solved` for solve order
* `/problem/aplusb` to display orange icon in title

In all of these, a `set` works well enough, and a `dict` is unnecessary.

### DIFF
--- a/judge/utils/problems.py
+++ b/judge/utils/problems.py
@@ -2,7 +2,7 @@ from collections import defaultdict
 from math import e
 
 from django.core.cache import cache
-from django.db.models import Case, Count, ExpressionWrapper, F, Max, When
+from django.db.models import Case, Count, ExpressionWrapper, F, When
 from django.db.models.fields import FloatField
 from django.utils import timezone
 from django.utils.translation import gettext_noop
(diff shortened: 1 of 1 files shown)


judge/utils/problems.py
```diff
@@ -44,11 +44,7 @@ def contest_attempted_ids(participation):
     key = 'contest_attempted:%s' % participation.id
     result = cache.get(key)
     if result is None:
-        result = {id: {'achieved_points': points, 'max_points': max_points}
-                  for id, max_points, points in (participation.submissions
-                                                 .values_list('problem__problem__id', 'problem__points')
-                                                 .annotate(points=Max('points'))
-                                                 .filter(points__lt=F('problem__points')))}
+        result = set(participation.submissions.values_list('problem__problem__id', flat=True).distinct())
```
quantum5, line 47: Does this work? Would save a join.

```suggestion
        result = set(participation.submissions.values_list('problem__problem_id', flat=True).distinct())
```

int-y1: Yes, `problem_id` exists. Done.

quantum5 (approved):

PR #2106, opened 2023-01-04
Clean up css for users

int-y1:
There's one visible change:

* In `/users/` and similar tables, the "Rank" header's right padding went from 10px to 5px. This makes the left/right padding symmetrical.


Xyene (approved):
LGTM, thanks!

PR #2107, opened 2023-01-04
Clean up css on contest list page

int-y1:
A few notes:

* `.floating-time*` is dead
* `!important` in `.content-description ul { padding: 0 0 0 2em !important; }`
I couldn't find a page where `!important` had an effect, so I deleted `!important`.

### DIFF
--- a/templates/contest/list.html
+++ b/templates/contest/list.html
@@ -3,20 +3,6 @@
     <meta name="description" content="The {{ SITE_NAME }}'s contest list - past, present, and future.">
 {% endblock %}
 
-{% block media %}
-    <style>
-        .time-left {
-            text-align: left;
-            color: #777;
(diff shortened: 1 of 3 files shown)


templates/contest/list.html
```diff
@@ -3,19 +3,7 @@
...
-            color: #777;
-            padding-top: 0.5em;
-        }
-
-        .content-description ul {
-            padding: 0 !important;
-        }
-    </style>
-{% endblock %}
+{% block media %}{% endblock %}
```
Ninjaclasher, line 6: We can probably delete this empty block.

int-y1: Done. I saw no change in the html source.

Xyene (approved):
LGTM, thanks!

PR #2109, opened 2023-01-07
Simplify AC query for /stats/language/

int-y1:
Improve readability. Page looks the same locally, and query speed is roughly the same.

`Count` `filter` was [introduced](https://docs.djangoproject.com/en/2.0/ref/models/conditional-expressions/#conditional-aggregation) in Django 2.0. This appears to generate the same SQL.


Xyene (approved):
LGTM, thanks!

PR #2111, opened 2023-01-07
Clean up org edit page

int-y1:
The edit page still behaves normally.

Before and after:
Capture


Ninjaclasher (approved):

templates/organization/edit.html
```diff
@@ -1,43 +1,7 @@
 {% extends "base.html" %}
 
-{% block js_media %}
-    {{ form.media.js }}
-    <script type="text/javascript">
-        window.django = {jQuery: $};
-
-        function pluralidx(count) {
-            return (count == 1) ? 0 : 1;
```
Ninjaclasher, line 9: questionable

Xyene: @quantum5 do you remember what this was for?

quantum5: Digging through the git history, it seems like I pulled in an admin widget at some point, which expected this thing. Since DMOJ was English-only back then, I shoved in this shim to make the code work.

Xyene (approved):
LGTM, thanks!

PR #2112, opened 2023-01-07
Clean up pagination css in user table

int-y1:
No visual changes expected.

Delete `.user-pagination` and another `<style>`. Other places use `div.top-pagination-bar`, but they have 1 child, and they should be unaffected.


Xyene (approved):
LGTM, thanks!

PR #2113, opened 2023-01-08
Clean up blog css

int-y1:
Visible changes:

* Link to comments went from `#555` to `primary75`. The number went from 12px to default size.

Capture


Xyene:
The comment link color is pretty jarring, but otherwise this LGTM.

Xyene (approved):
LGTM, thanks!

PR #2117, opened 2023-01-08
Simplify CSS in judge version table

int-y1:
only the `tr:first-child` css did anything important

page is still mobile friendly
Capture


Xyene (approved):
LGTM, thanks!

PR #2118, opened 2023-01-09
Fix filtering problems by full AC

int-y1:
* Give a green icon to >100% AC subs
* Fix filtering solved problems

### DIFF
--- a/judge/utils/problems.py
+++ b/judge/utils/problems.py
@@ -24,8 +24,8 @@ def contest_completed_ids(participation):
     key = 'contest_complete:%d' % participation.id
     result = cache.get(key)
     if result is None:
-        result = set(participation.submissions.filter(submission__result='AC', points=F('problem__points'))
-                     .values_list('problem__problem__id', flat=True).distinct())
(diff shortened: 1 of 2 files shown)


judge/utils/problems.py
```diff
@@ -24,8 +24,8 @@ def contest_completed_ids(participation):
     key = 'contest_complete:%d' % participation.id
     result = cache.get(key)
     if result is None:
-        result = set(participation.submissions.filter(submission__result='AC', points=F('problem__points'))
-                     .values_list('problem__problem__id', flat=True).distinct())
+        result = set(participation.submissions.filter(submission__result='AC', points__gte=F('problem__points'))
```
Ninjaclasher, line 27: Could we perhaps use `case_points` / `case_total` here as well?
```suggestion
        result = set(participation.submissions.filter(submission__result='AC', submission__case_points=F('submission__case_total'))
```

int-y1: This is a ContestSubmission. Is it still ok to check `case_points`/`case_total`?

Ninjaclasher (commented):
LGTM

kiritofeng (approved):
LGTM, thanks!

PR #2119, opened 2023-01-09
Slightly clean up navbar css

int-y1:
* Remove useless styles
* Introduce `.show-list` so that I can safely delete `display: block !important`

No visible changes expected.

Fun fact: On narrow screens, `margin: 0 0 0 -5px !important` was covering `border-left: 4px solid $highlight_blue`

### DIFF
--- a/resources/base.scss
+++ b/resources/base.scss
@@ -227,30 +227,17 @@ h4 {
     height: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


Xyene (approved):
LGTM, thanks!

PR #2120, opened 2023-01-10
Move navbar CSS into separate file

int-y1:
Delete `#nav-placeholder` because it's dead. Then copy and paste the remaining css.


Xyene (approved):
LGTM, thanks!

PR #2124, opened 2023-01-11
Fix rendering of target color on contest page

### LINKED ISSUE #1881: Fix rendering of target colour

Balint-R:
Fixes #1881.

### DIFF
--- a/templates/contest/ranking.html
+++ b/templates/contest/ranking.html
@@ -26,7 +26,7 @@
         }
 
         .userinfo a, .user-name a, .user-name form {
-            display: inline !important;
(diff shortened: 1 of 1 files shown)


int-y1 (approved):
Looks like `!important` is still needed to display the disqualify icon correctly.

Xyene (approved):
LGTM, thanks!

PR #2126, opened 2023-01-15
Remove msp math renderer

int-y1:
msp is dead because users can't pick msp as a math engine (see https://github.com/DMOJ/online-judge/commit/037c745fb936d9a240c3f4532d4cdb1bb06e18d7).


Xyene (approved):
LGTM, thanks!

PR #2121, opened 2023-01-10
Clean up blog page css

int-y1:
Visible changes:
* Blog sidebar became slightly narrower.
* "RSS / Atom" is moved up 0.25em.

PR #2128, opened 2023-01-15
Fix hot problems css

int-y1:
#2121 broke the hot problems css. Unbreak it.

Also, `padding: 0 !important` results in too little padding, so delete it.


kiritofeng (approved):
Thanks!

Issue #2129, opened 2023-01-16
Enable MathML support for Chrome 109+


Xyene:
*This feature tracks everything that must be done in order to enable MathML support on Chrome.*

---

Chrome 109 added support for MathML Core.

* https://www.w3.org/TR/mathml-core/
* https://caniuse.com/?search=mathml

Nominally, this means we can use MathML to render equations in Chrome, after merging https://github.com/DMOJ/online-judge/pull/1663.

@int-y1 did some early testing and discovered that spacing does not work as expected in Chrome 109. For example, in https://dmoj.ca/problem/fibonacci:

### Chrome 109

![image](https://user-images.githubusercontent.com/1403503/212603818-4a9ec533-9ea2-45e4-b973-38e594090771.png)

### Firefox

![image](https://user-images.githubusercontent.com/1403503/212603777-017680a9-8d88-405d-bcf4-c62b28df5f3e.png)

Notably, Chrome isn't rendering `\,` as a thin space. This is because Mathoid (through MathJax) is generating markup that looks like

```xml
<mrow>
<mn>1</mn>
<mspace width="thinmathspace"></mspace>
<mn>000</mn>
<mspace width="thinmathspace"></mspace>
<mn>000</mn>
<mspace width="thinmathspace"></mspace>
<mn>007</mn>
</mrow>
```

and the shorthands `thinmathspace` etc. are not supported in MathML Core. `thinmathspace`, for instance, should be written as `0.1667em` instead, and this renders correctly in both Chrome and Firefox.

The instance of Mathoid on https://dmoj.ca has been untouched since 2017, so I figured I could just update it and things would Just Work. After all, the MathJax codebase has no more references to `thinmathspace`. How wrong I was.

First off, Mathoid has slightly changed its config format, in a backwards-incompatible way. Old configs will fail to load with `Cannot read property 'displayMessages' of undefined` errors on every request.

Next, Mathoid has removed the ability to generate PNGs recently, in https://github.com/wikimedia/mathoid/commit/b152b3c45bd96f5a08598d8232493f9a9b646032. This makes DMOJ's renderer complain with messages like

```
ERROR 2023-01-16 04:29:32,343 mathoid Mathoid did not return required information (mml, png, svg, mathoidStyle needed):
```

[...]

PR #2130, opened 2023-01-16
Remove png support in mathoid

int-y1:
Part of #2129. I deleted `output_png` because there's no `png` math engine. (I think `raw` is dead too, but it's probably easier to keep it)

It works in shell.

```py
>>> from judge.utils.mathoid import *
>>> print( MathoidMathParser.output_jax(None, {'svg': 'meme.svg', 'css': 'display:none;', 'tex':'1 \\le 2', 'display': 0}) )
<span class="inline-math"><img class="tex-image" src="meme.svg" style="display:none;" alt="1 \le 2"><span class="tex-text" style="display:none">~1 \le 2~</span></span>
>>> print( MathoidMathParser.output_svg(None, {'svg': 'meme.svg', 'css': 'display:none;', 'tex':'1 \\le 2', 'display': 0}) )
<img class="inline-math" src="meme.svg" style="display:none;" alt="1 \le 2">
```

Todo:

* test in prod
* check if texoid is affected by the png removal
* check all math engines, check pdf math

### DIFF
--- a/judge/models/choices.py
+++ b/judge/models/choices.py
@@ -57,9 +57,9 @@ ACE_THEMES = (
 
 MATH_ENGINES_CHOICES = (
     ('tex', _('Leave as LaTeX')),
-    ('svg', _('SVG with PNG fallback')),
+    ('svg', _('SVG only')),
     ('mml', _('MathML only')),
-    ('jax', _('MathJax with SVG/PNG fallback')),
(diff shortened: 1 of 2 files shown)


quantum5 (approved):

Xyene (approved):
LGTM, thanks!

PR #2131, opened 2023-01-16
Add OpenGraph to blog posts

int-y1:
this should improve the discord preview of `https://dmoj.ca/post/<id>-<slug>`

* For `meta_description`, use `summary` and fall back to `content`.
* Fix a bug with `generate_opengraph`. If `data` is one line, `tree` would be a `<p>` and `.//p` can't find anything. Use `..//p`.

### DIFF
--- a/judge/views/blog.py
+++ b/judge/views/blog.py
@@ -11,6 +11,7 @@ from judge.models import BlogPost, Comment, Contest, Language, Problem, ProblemC
     Ticket
 from judge.utils.cachedict import CacheDict
 from judge.utils.diggpaginator import DiggPaginator
+from judge.utils.opengraph import generate_opengraph
 from judge.utils.tickets import filter_visible_tickets
 from judge.utils.views import TitleMixin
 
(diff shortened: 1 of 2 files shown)


judge/views/blog.py
```diff
@@ -101,7 +102,12 @@ def get_comment_page(self):
 
     def get_context_data(self, **kwargs):
         context = super(PostView, self).get_context_data(**kwargs)
-        context['og_image'] = self.object.og_image
+
+        if not self.object.og_image or not self.object.summary:
+            metadata = generate_opengraph('generated-meta-blog:%d' % self.object.id, self.object.content, 'blog')
+        context['meta_description'] = self.object.summary or metadata[0]
```
quantum5, line 108: Not sure if `self.object.summary` is truly appropriate and we shouldn't just display `metadata[0]` always. `summary` is what's displayed on the home page, which can be very long.

int-y1: Updated to `self.object.summary or self.object.content`.

quantum5 (approved):

Xyene:
Thanks!

PR #2133, opened 2023-01-16
Introduce navbar arrow

int-y1:
These arrows never show up because the code did `node.is_leaf_node` instead of `node.is_leaf_node()`.

Let's re-introduce the navbar arrow (and clean some of its code). It works well enough locally.

Capture

### DIFF
--- a/resources/common.js
+++ b/resources/common.js
@@ -125,7 +125,6 @@ $(function () {
             $(this).blur().removeClass('hover');
         else {
             $(this).addClass('hover');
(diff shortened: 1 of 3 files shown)


quantum5:
Wait these arrows were lost in the Jinja2 transformation, and should probably exist so that you can open menus on mobile in a non-crazy way.

Ninjaclasher (approved):
LGTM

Xyene (approved):
LGTM, thanks!

PR #2136, opened 2023-01-21
Return message instead of 404 for muted users on edit profile page

Ninjaclasher:


int-y1:
Is 400 or 403 more appropriate here? Otherwise, lgtm.

int-y1:
Change `Bad request` to `Can't edit profile` to make it clear that the user is trying to edit profile.

Ninjaclasher:
>  Is 400 or 403 more appropriate here?
Changed to `403`.
> Change Bad request to Can't edit profile to make it clear that the user is trying to edit profile.
Done.

kiritofeng (approved):
LGTM

Xyene (approved):
LGTM, thanks!

PR #2139, opened 2023-01-23
Delete old, deprecated API v1

### LINKED ISSUE #1879: Deprecate API v1
We should consider deprecating API v1. There are some endpoints that dump megabytes of data and were not designed for the current size of DMOJ. Notably:
 1. https://dmoj.ca/api/user/list dumps all users (9MB at the time of writing)
 2. https://dmoj.ca/api/user/submissions/wleung_bvg (4.1MB at the time of writing)
 3. https://dmoj.ca/api/user/submissions/xyene (4.8MB at the time of writing)

API v2 contains all the data in v1 and has proper support for pagination, so we should be directing users to upgrade to v2.

quantum5:
Fixes #1879


Xyene (approved):
L
G
T
M

PR #2140, opened 2023-01-23
Use a better default for timezone map

int-y1:
Replace `http://momentjs.com/static/img/world.png` with `https://static.dmoj.ca/assets/earth.jpg`. This image is poor quality, but it's good enough for now.

In the future, maybe move the map to `static('libs/timezone-map/foo.png')`. It's possible to use a cleaned-up version of https://en.wikipedia.org/wiki/File:Blue_Marble_2002.png


kiritofeng:
:+1: for downloading it (also, is the license on the image ok with us using it like this?)

Xyene (approved):
LGTM, thanks!

PR #2146, opened 2023-01-30
Add site theme profile option to toggle dark/light themes

Ninjaclasher:
Part of #2035.


templates/user/edit-profile.html
```diff
@@ -294,6 +294,10 @@
                             <td><label class="inline-header grayed">{{ _('Editor theme:') }}</label></td>
                             <td><span class="fullwidth">{{ form.ace_theme }}</span></td>
                         </tr>
+                        <tr>
```
Xyene, line 297: We should gate this field behind the user having experimental features enabled. Also, we should only emit the light color scheme `<link.` even if `site_theme` is set to default, if experimental features are not enabled.

(Don't want to immediately start receiving many duplicate bug reports until dark mode has fewer outstanding issues.)

Ninjaclasher: Fixed.

Xyene (approved):
LGTM, thanks!

PR #2152, opened 2023-02-05
Standardize site language's names

int-y1:
The site footer's language dropdown is unaffected.

The admin site's language dropdown is affected:
Capture


quantum5:
Is the language selection dropdown in admin the only place that's affected? I am not convinced that should be changed to the native name for all languages. Presumably, as an admin who doesn't understand all languages DMOJ supports, you'd prefer seeing the language names in your preferred language?

int-y1:
The site footer's language dropdown used the native name for each language, and I thought it's a good idea to copy it over to admin. This will make the dropdowns more consistent.

An admin can read the language codes to determine which languages are supported. In the old version, admin site doesn't show language codes.

quantum5:
I don't think this is a question of consistency. The language dropdown in the site footer is meant for the user to select their native language when they may not necessarily understand the current UI language. If the footer is in Chinese, you would never recognize the word for "English" and switch back, for example.

In the admin, it is assumed that the language you have chosen is a language you are comfortable with. In such a case, it is more ergonomic to refer to the language as you would refer to it, rather than a native speaker. You would never say "the Ελληνικά version of this problem", only "the Greek version of this problem" if you are an English speaker, for example. Therefore, `Greek` is a lot more helpful than `Ελληνικά (el)` in this context. Most people wouldn't even know `el` is Greek...

PR #2153, opened 2023-02-05
Fix grading spinner colour on dark mode

Ninjaclasher:
Part of #2035.

### DIFF
--- a/resources/submission.scss
+++ b/resources/submission.scss
@@ -54,6 +54,10 @@
         .score {
             font-size: 1.3em;
             color: #000;
+
+            .grading-spinner {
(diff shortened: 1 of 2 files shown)


resources/submission.scss
```diff
@@ -54,6 +54,10 @@
         .score {
             font-size: 1.3em;
             color: #000;
+
+            .grading {
```
Xyene, line 58: Can we call this something more descriptive, like `grading-spinner`?

Ninjaclasher: Done

Xyene (approved):
LGTM, thanks!

PR #2154, opened 2023-02-05
Add ability to specify table ids for user tables

Ninjaclasher:
This is necessary to be able to move contest ranking CSS to SCSS files without affecting other user tables.

With this, we can add dark mode colours for the contest ranking table, which will be done in a future PR.


Xyene (approved):
LGTM, thanks!

PR #2156, opened 2023-02-05
Add submission activity dark mode

Ninjaclasher:
Part of #2035. Dark mode colours shamelessly taken from Github contribution activity.

After:
image

### DIFF
--- a/resources/vars-default.scss
+++ b/resources/vars-default.scss
@@ -18,6 +18,7 @@ $color_link200: #faa700;    // active
 
 $color_pageBg: #fff;
 
+// custom one-off colours
 $color_rating_none: #999;
 $color_rating_newbie: #999;
 $color_rating_amateur: #00a900;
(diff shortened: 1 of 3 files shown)


resources/vars-default.scss
```diff
@@ -30,3 +30,10 @@ $color_rating_target: #700;
 $path_to_root: '.';     // relative path from style.css to STATIC_ROOT
 
 @mixin vars-img {}
+
+// custom colours
+$color_user_submission_activity0: #ddd;
```
Xyene, line 35: Is there a reason this is after the mixin, or can we keep things ordered a little bit better?

Ninjaclasher: Was thinking to separate one-off colours but I can move it with the rest.

Ninjaclasher: Done

Xyene (approved):
LGTM, thanks!

PR #2157, opened 2023-02-05
Fix dark mode colour for alert close button

Ninjaclasher:
Part of #2035.

`text-shadow` does nothing on light mode, and messes up the rendering on dark mode, so just remove it.


Xyene (approved):
LGTM, thanks!

PR #2161, opened 2023-02-05
Standardize gravatar images

Ninjaclasher:
Consists of 4 changes:
 - Use gravatar size parameter in URL instead of always grabbing `size=135` and then resizing ourselves.
 - Always use `user-gravatar` class. Side effect is comment/ticket gravatars now have rounded borders.
      - Add overrides for comment gravatars on small-width devices so the gravatar is not a circle.
 - Add `background-color: white` to all gravatars (part of #2035).

Before:
image
image

After:
image
image

### DIFF
--- a/resources/users.scss
+++ b/resources/users.scss
@@ -117,6 +117,7 @@ tr:target {
 img.user-gravatar {
     display: block;
     border-radius: 6px;
+    background-color: white;
 }
 
 .user-content {
--- a/resources/comments.scss
+++ b/resources/comments.scss
@@ -144,11 +144,6 @@ a {
         display: flex;
     }
 
-    .gravatar {
-        width: 75px;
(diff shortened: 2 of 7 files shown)


Xyene (approved):
LGTM, thanks!

resources/users.scss
```diff
@@ -117,6 +117,7 @@ tr:target {
 img.user-gravatar {
     display: block;
     border-radius: 6px;
```
Xyene, line 119: We should also standardize this. It looks like we're using 2px in comments but 6px in the users page.

Ninjaclasher: We're only using 2px in comments for small-width screens since we use a smaller gravatar size (otherwise the gravatar is a circle).

PR #2162, opened 2023-02-05
Use a darker (light theme) / lighter (dark theme) color for tab text

Ninjaclasher:
This matches the tab text color before 1b68ed36b. Currently, the text looks off since it has the same colour as unclickable text.


Ninjaclasher:
Before this change:
image
image

After this change:
image
image

Xyene (approved):
LGTM, thanks!

PR #2164, opened 2023-02-05
Use better favicon to fix #2163

### LINKED ISSUE #2163: Favicon is not stable between page loads
On some page loads, we get:

image

While on others (of the same page!), we get:

image

I've observed this before between the admin site and the regular site on Linux Firefox and Chrome OS Chrome, but this seems to happen for every page on Firefox for macOS.

quantum5:
New favicon correctly uses borderless version for 32x32 and 16x16.

Not sure why we have two versions, but updating them to be consistent.


Xyene (approved):
LGTM, yolo

PR #2165, opened 2023-02-06
Use custom dmoj select2 theme for dark mode support

Ninjaclasher:
Part of #2035. This PR will make the problem submit language select2 look bad due to a lot of custom CSS, but it will be fixed in a followup PR as this one is getting too large.

After:
image
image

image
image

The admin site only uses the light mode version:
image

### DIFF
--- a/judge/template_context.py
+++ b/judge/template_context.py
@@ -30,6 +30,7 @@ def get_resource(request):
         'FONTAWESOME_CSS': settings.FONTAWESOME_CSS,
         'DMOJ_SCHEME': scheme,
         'DMOJ_CANONICAL': settings.DMOJ_CANONICAL,
+        'DMOJ_SELECT2_THEME': settings.DMOJ_SELECT2_THEME,
     }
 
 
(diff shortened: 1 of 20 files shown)


Ninjaclasher:
Changed so the search dropdown on light theme has a white background, so now it looks like this:
image

judge/template_context.py
```diff
@@ -30,6 +30,7 @@ def get_resource(request):
         'FONTAWESOME_CSS': settings.FONTAWESOME_CSS,
         'DMOJ_SCHEME': scheme,
         'DMOJ_CANONICAL': settings.DMOJ_CANONICAL,
+        'SELECT2_THEME': settings.DMOJ_SELECT2_THEME,
```
Xyene, line 33: For consistency with the other settings, this should be

```suggestion
        'DMOJ_SELECT2_THEME': settings.DMOJ_SELECT2_THEME,
```

Ninjaclasher: Done

Xyene (approved):
LGTM, thanks!

PR #2167, opened 2023-02-06
Clean up problem submit select2 CSS

Ninjaclasher:
Part of #2035. Depends on #2165.

Before:
image

After:
image
image


Xyene (approved):
LGTM, thanks!

PR #2175, opened 2023-02-06
Migrate comment's associated page to use id

### LINKED ISSUE #2145: Comment page is not changed when object code is changed
For example, a problem with code `example` will have its comments stored in page `p:example`. When the problem code is changed to `example2`, the new page would be `p:example2`. However, the existing comments still have page `p:example`. As a result, all comments are lost (until the problem is renamed back to `example`).

int-y1:
Fixes #2145. List of changes:

* Create `page_old` that holds the old values. Admins can see `page_old`, and this field won't be used anywhere.
* Change links from string to id. `c:contestkey` `p:problemcode` `s:problemcode` -> `c:123` `p:123` `s:123`. If id doesn't exist, use 0.
* Update `Comment.most_recent` and `Comment.link` to use id. This affects the home page's "Comment stream". (Also update `get_page_title` and `is_accessible_by`)

I did some simple testing and nothing died. This pr needs more testing.

(Tickets aren't affected because they use `content_type` and `object_id`)

Future tasks:

* Update comments / comment locks that refer to `c:0` `p:0` `s:0`.
* Delete `page_old`.


int-y1:
It's bad design to put id into a `CharField`. If the id approach is used, it's better to do what tickets do.

Issue #2171, opened 2023-02-06
Clean up submit form alignment


Xyene:
### Current

<img width="994" alt="image" src="https://user-images.githubusercontent.com/1403503/216886894-d314a73b-6af1-4cf4-9b55-2a73c8038574.png">

### Proposed

<img width="1113" alt="image" src="https://user-images.githubusercontent.com/1403503/216886850-72d4b8ba-e16f-402b-bbbd-a20017c7b80a.png">

### Or even...

<img width="593" alt="image" src="https://user-images.githubusercontent.com/1403503/216887246-9d650cf9-4106-4fe3-8aa0-47399d3672f2.png">

PR #2176, opened 2023-02-06
Submit Form Alignment Fix

nachiketkanore:
Changed the CSS properties for the alignment issue mentioned in #2171

### DIFF
--- a/templates/problem/list.html
+++ b/templates/problem/list.html
@@ -207,7 +207,7 @@
                             <a href="{{ sort_links.group }}">{{ _('Category') }}{{ sort_order.group }}</a>
                         </th>
                         {% if show_types %}
-                            <th>
+                            <th class="types">
                                 <a href="{{ sort_links.type }}">{{ _('Types') }}{{ sort_order.type }}</a>
                             </th>
                         {% endif %}


kiritofeng:
Can you please attach a screenshot of the before/after?

nachiketkanore:
image
image

Ninjaclasher:
What does the submit bar and language select look like?

kiritofeng:
It seems the grey border doesn't cover the corner:
image
Can you see if this can be fixed?

nachiketkanore:
New screenshots
image
zoomed in:
image

resources/select2/_multiple.scss
```diff
@@ -81,6 +81,10 @@
 }
 
 &.select2-container--focus {
+    width: 100%;
```
Ninjaclasher, line 84: Let's ... not change values here, as they affect _all_ select2 widgets across the entire site. If they are necessary, add them in `problem.scss` inside i.e. `#problem-submit`.

resources/problem.scss
```diff
@@ -237,13 +237,14 @@ ul.problem-list {
...
         float: right;
+        padding-right: 6px;
+        padding-bottom: 6px;
     }
 
     #submit-wrapper {
-        margin-top: 0.7em;
+        margin-top: .2em;
 
         #editor, #language {
```
Ninjaclasher, line 247: This block is empty, just delete.

resources/base.scss
```diff
@@ -86,7 +86,6 @@ th {
 .form-area {
     display: inline-block;
     background: $color_primary5;
-    padding: 5px 10px 10px;
```
Ninjaclasher, line 89: Again, let's not change it globally unless you're sure this won't adversely affect any other `form-area` across the entire site.

PR #2178, opened 2023-02-08
Use different default ace theme depending on site theme

Ninjaclasher:
Part of #2035. No visual changes apart from a different default theme is used if dark mode is activated.


judge/models/choices.py
```diff
@@ -21,6 +21,7 @@ def make_timezones():
 del make_timezones
 
 ACE_THEMES = (
+    ('auto', _('Follow theme default')),
```
Ninjaclasher, line 24: Note: this isn't sorted so that it's shown first in the dropdown in edit profile.

judge/models/profile.py
```diff
@@ -226,6 +226,15 @@ def display_name(self):
...
         return self.submission_set.filter(result='AC', case_points__gte=F('case_total')).exists()
 
+    @cached_property
+    def resolved_ace_theme(self):
+        if self.ace_theme != 'auto':
+            return self.ace_theme
+        if self.site_theme != 'auto':
+            return settings.DMOJ_THEME_DEFAULT_ACE_THEME.get(self.site_theme)
+        # This must be resolved client-side using prefers-color-scheme.
+        return None
```
Xyene, line 236: Maybe we should call this `"auto"` or something. Passing `None` around doesn't carry the semantic meaning of "this needs to be resolved client-side".

Ninjaclasher: If we use something like `auto`, then we'll also need to add a check specifically for that value in the ace widget (https://github.com/DMOJ/online-judge/blob/master/django_ace/widgets.py#L43), while using `None` sort of means "we don't have a preferred theme, use whatever is the default determined by `django_ace`"

Xyene: Well, we already have a check in the ace widget around whether the theme is null or not. But I'm happy enough with your parenthesized explanation.

Xyene (approved):

PR #2181, opened 2023-02-08
Make dark mode normal text slightly darker

Ninjaclasher:
Part of #2035.

Before:
image

After:
image


resources/vars-dark.scss
```diff
@@ -11,7 +11,7 @@ $color_primary50: #808080;
 $color_primary66: #aaa;     // tabs
 $color_primary75: #ccc;     // widget
 $color_primary90: #eee;
-$color_primary100: #f8f8f8;
+$color_primary100: #e0e0e0;
```
int-y1, line 14: this is darker than primary90

PR #2191, opened 2023-02-12
Use multi-column index for User model

quantum5:
Haven't tested the performance, but index is no brainer given the queries we run.


Xyene (approved):
LGTM, thanks!

PR #2192, opened 2023-02-12
Use multi-column indices for Submission model

quantum5:
Haven't tested the performance, but index is no brainer given the queries we run.


Xyene (commented):
LGTM, thanks!

Xyene (approved):

PR #2197, opened 2023-02-17
Validate site data file names and contents

### LINKED ISSUE #2196: Validate zip for ProblemData uploads
The upload for "Data zip file" currently does not check if the uploaded file is actually a `zip`. This can potentially cause an issue if the user tries to upload an `init.yml` instead of a zip file.

Why people cannot just follow instructions, I do not know.

Riolku:
Fixes #2196. The old check for zip file format correctness was broken. We also need a check for generator names.

### DIFF
--- a/judge/views/problem_data.py
+++ b/judge/views/problem_data.py
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
     def clean_zipfile(self):
         if hasattr(self, 'zip_valid') and not self.zip_valid:
             raise ValidationError(_('Your zip file is invalid!'))
-        return self.cleaned_data['zipfile']
+
+        zipfile = self.cleaned_data['zipfile']
+        if zipfile and not zipfile.name.endswith('.zip'):
+            raise ValidationError(_("Zip files must end in '.zip'"))
+
+        return zipfile
+
+    def clean_generator(self):
+        generator = self.cleaned_data['generator']
+        if generator and generator.name == 'init.yml':
+            raise ValidationError(_('Generators must not be named init.yml.'))
+
+        return generator
 
     clean_checker_args = checker_args_cleaner
 
@@ -166,7 +178,7 @@ class ProblemDataView(TitleMixin, ProblemManagerMixin):
             elif data.zipfile:
                 return ZipFile(data.zipfile.path).namelist()
         except BadZipfile:
-            return []
+            return None
         return []
 
     def get_context_data(self, **kwargs):
@@ -174,10 +186,10 @@ class ProblemDataView(TitleMixin, ProblemManagerMixin):
         if 'data_form' not in context:
             context['data_form'] = self.get_data_form()
             valid_files = context['valid_files'] = self.get_valid_files(context['data_form'].instance)
-            context['data_form'].zip_valid = valid_files is not False
             context['cases_formset'] = self.get_case_formset(valid_files)
-        context['valid_files_json'] = mark_safe(json.dumps(context['valid_files']))
-        context['valid_files'] = set(context['valid_files'])
+        if context['valid_files']:
+            context['valid_files_json'] = mark_safe(json.dumps(context['valid_files']))
+            context['valid_files'] = set(context['valid_files'])
         context['all_case_forms'] = chain(context['cases_formset'], [context['cases_formset'].empty_form])
         return context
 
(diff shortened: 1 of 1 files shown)


judge/views/problem_data.py
```diff
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
...
             raise ValidationError(_('Your zip file is invalid!'))
-        return self.cleaned_data['zipfile']
+
+        zipfile = self.cleaned_data['zipfile']
+        if zipfile and zipfile.name == 'init.yml':
+            raise ValidationError(_('Zip files must not be named init.yml'))
+
+        return zipfile
+
+    def clean_generator(self):
```
kiritofeng, line 54: Dumb question: are we sure this is called?

kiritofeng (commented):
LGTM, but I also want @quantum5 and/or @Xyene to look this over before we merge.

judge/views/problem_data.py
Riolku: I did test it, so... I guess, yes?

judge/views/problem_data.py
```diff
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
     def clean_zipfile(self):
         if hasattr(self, 'zip_valid') and not self.zip_valid:
             raise ValidationError(_('Your zip file is invalid!'))
-        return self.cleaned_data['zipfile']
+
+        zipfile = self.cleaned_data['zipfile']
+        if zipfile and zipfile.name == 'init.yml':
+            raise ValidationError(_('Zip files must not be named init.yml'))
```
Ninjaclasher, line 50:
```suggestion
            raise ValidationError(_('Zip files must not be named init.yml.'))
```
For consistency with the other error messages.

judge/views/problem_data.py
```diff
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
...
+
+        return zipfile
+
+    def clean_generator(self):
+        generator = self.cleaned_data['generator']
+        if generator and generator.name == 'init.yml':
+            raise ValidationError(_('Generators must not be named init.yml'))
```
Ninjaclasher, line 57:
```suggestion
            raise ValidationError(_('Generators must not be named init.yml.'))
```
For consistency with the other error messages.

judge/views/problem_data.py
```diff
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
     def clean_zipfile(self):
         if hasattr(self, 'zip_valid') and not self.zip_valid:
             raise ValidationError(_('Your zip file is invalid!'))
-        return self.cleaned_data['zipfile']
+
+        zipfile = self.cleaned_data['zipfile']
+        if zipfile and zipfile.name == 'init.yml':
```
Xyene, line 49: Rather than this very specific check, I feel like we should enforce that zip files... end with `.zip`.

Riolku: Fixed. The check for the generator has been left as is.

judge/views/problem_data.py
```diff
@@ -44,7 +44,19 @@ class ProblemDataForm(ModelForm):
     def clean_zipfile(self):
         if hasattr(self, 'zip_valid') and not self.zip_valid:
             raise ValidationError(_('Your zip file is invalid!'))
-        return self.cleaned_data['zipfile']
+
+        zipfile = self.cleaned_data['zipfile']
+        if zipfile and not zipfile.endswith('.zip'):
+            raise ValidationError(_('Zip files must not be named init.yml.'))
```
kiritofeng, line 50:
```suggestion
            raise ValidationError(_("Zip files should end in '.zip'."))
```

kiritofeng (commented):
Don't forget to update the error message

judge/views/problem_data.py
Riolku: I forgot :(

kiritofeng (approved):
LGTM

judge/views/problem_data.py
```diff
@@ -166,15 +178,15 @@ def get_valid_files(self, data, post=False):
...
-            return []
+            return None
         return []
 
     def get_context_data(self, **kwargs):
         context = super(ProblemDataView, self).get_context_data(**kwargs)
         if 'data_form' not in context:
             context['data_form'] = self.get_data_form()
             valid_files = context['valid_files'] = self.get_valid_files(context['data_form'].instance)
-            context['data_form'].zip_valid = valid_files is not False
```
quantum5, line 177: When could the old code return `False` for this? Is this just... completely borked?

quantum5 (approved):
Doesn't look broken.

judge/views/problem_data.py
Riolku: Yeah, it sort of looks like it currently always returns "yep, valid zip".

Riolku: Ok, updated and tested. Now we correctly display errors for invalid zip files, instead of prod which does:
image

PR #2198, opened 2023-02-17
Prevent data leak after login but before 2fa completed

jdabtieu:
When a user logs in, but before they complete 2fa, the `window.user` object is populated with the user's data (including email) and the user-script is injected.

This is not intended behavior because the user should not be considered logged in yet, so they should not get access to those resources.

### DIFF
--- a/judge/views/two_factor.py
+++ b/judge/views/two_factor.py
@@ -14,6 +14,7 @@ from django.urls import reverse
 from django.utils.http import is_safe_url
 from django.utils.translation import gettext as _, gettext_lazy
 from django.views.generic import FormView, View
+from django.views.generic.base import ContextMixin
 from django.views.generic.detail import SingleObjectMixin
 
 from judge.forms import TOTPEnableForm, TOTPForm, TwoFactorLoginForm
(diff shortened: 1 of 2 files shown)


judge/views/two_factor.py
```diff
@@ -225,10 +226,11 @@ def post(self, request, *args, **kwargs):
         return HttpResponse()
 
 
-class TwoFactorLoginView(SuccessURLAllowedHostsMixin, TOTPView):
+class TwoFactorLoginView(SuccessURLAllowedHostsMixin, TOTPView, ContextMixin):
     form_class = TwoFactorLoginForm
     title = gettext_lazy('Perform Two-factor Authentication')
     template_name = 'registration/two_factor_auth.html'
+    extra_context = {'twofa_active': True}
```
quantum5, line 233: Please don't call it `twofa_active`. `twofa` hurts my eyes. Call it `_2fa` or `tfa` or `mfa` or anything else. Also it's not `active`, it's really "in progress".

jdabtieu: Renamed

quantum5 (approved):

PR #2199, opened 2023-02-17
Create sitemap generation script

quantum5:
This PR creates a new management command that renders multiple paginated sitemap files and a sitemap index file. This can be used by larger installs to render a series of static sitemap files inside a cronjob.


kiritofeng (approved):
Seems legit

PR #2206, opened 2023-02-18
Optimize contest ranked submissions

Ninjaclasher:
Remove filtering non-zero point submissions since they're already filtered. Remove joinined contest participation/contest submission since we only need the contest id, which we can get from `contest_object_id`.

### DIFF
--- a/judge/views/ranked_submission.py
+++ b/judge/views/ranked_submission.py
@@ -18,10 +18,9 @@ class RankedSubmissions(ProblemSubmissions):
     def get_queryset(self):
         params = [self.problem.id]
         if self.in_contest:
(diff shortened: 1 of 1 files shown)


judge/views/ranked_submission.py
```diff
@@ -21,7 +21,7 @@ def get_queryset(self):
             contest_join = """INNER JOIN judge_contestsubmission AS cs ON (sub.id = cs.submission_id)
                               INNER JOIN judge_contestparticipation AS cp ON (cs.participation_id = cp.id)"""
```
quantum5, line 22: Do we still need this join?

Ninjaclasher: Fixed.

Xyene (approved):
LGTM, thanks!

PR #2208, opened 2023-02-18
Avoid expensive problem queries on submissions

quantum5:
This PR avoids querying for `completed_problem_ids`, `editable_problem_ids`, and `tester_problem_ids` on submission lists unless absolutely necessary. This eliminates the massively expensive query for `editable_problem_ids` for users who can edit almost all problems.


Xyene (approved):
LGTM, thanks!

PR #2216, opened 2023-02-20
Store revision count on comment model

quantum5:
Did some basic testing (including the migration) on a test site, seems to work.


Xyene (approved):
LGTM, thanks!

PR #2218, opened 2023-02-20
Drop submission indexes that are prefixes of other indexes

quantum5:
The four dropped `Submission` indexes will save 78 MB each in production, and the `SubmissionTestCase` index saves 2.6 GB.


Xyene (approved):
LGTM, thanks!

PR #2221, opened 2023-02-20
Use faster author/curator/tester check for get_visible_problems

quantum5:


judge/views/problem.py
```diff
@@ -449,9 +448,7 @@ def get_normal_queryset(self):
         if self.profile is not None:
             # This is way faster than the obvious |= Q(authors=self.profile) et al. because we are not doing
```
Xyene, line 449: Outdated comment

Xyene (approved):
LGTM, thanks!

PR #2222, opened 2023-02-20
Use faster organization check for get_visible_problems

quantum5:
This means that we can remove the expensive distinct call from `get_visible_problems` callers.


Xyene (approved):
LGTM, thanks!

PR #2229, opened 2023-02-23
Rework misc config to run exactly one query

quantum5:
There should never be that much stuff in MiscConfig, so let's just read it all and deal with it in python instead of trying to run one query for each item of interest and poorly cache it.

This PR also moved MiscConfigDict into a middleware so it can be used outside of templates.


Xyene (approved):
LGTM, thanks!

PR #2232, opened 2023-02-26
Remove old PDF renderers

Xyene:
The SlimerJS and PhantomJS backends have not been used in many years, are undocumented, and should not be used.


Ninjaclasher:
https://github.com/DMOJ/online-judge/blob/master/dmoj/settings.py#L142

The settings should probably also be cleaned up.

quantum5 (approved):

PR #2236, opened 2023-03-02
Fix bug with non-existent organizations

kiritofeng:


judge/views/organization.py
```diff
@@ -131,8 +132,11 @@ class OrganizationUsers(QueryStringSortMixin, DiggPaginatorMixin, BaseOrganizati
     context_object_name = 'users'
 
     def get_queryset(self):
-        return self.object.members.filter(is_unlisted=False).order_by(self.order).select_related('user') \
-            .defer('about', 'user_script', 'notes')
+        try:
+            return self.object.members.filter(is_unlisted=False).order_by(self.order).select_related('user') \
+                .defer('about', 'user_script', 'notes')
+        except Http404:
+            raise
```
Xyene, line 139: What is this doing?

quantum5: This is literally a no-op, please delete @kiritofeng.

judge/views/organization.py
```diff
@@ -47,7 +48,7 @@ def dispatch(self, request, *args, **kwargs):
         try:
             return super(OrganizationMixin, self).dispatch(request, *args, **kwargs)
         except Http404:
-            key = kwargs.get(self.slug_url_kwarg, None)
+            key = kwargs.get(self.slug_url_kwarg, None) if hasattr(self, 'slug_url_kwarg') else None
```
quantum5, line 51: I don't know if we want this `hasattr` check. For `AttributeError` to occur, the site would be beyond broken.

quantum5 (approved):

judge/views/organization.py
kiritofeng: Fixed by #2240.

PR #2243, opened 2023-03-05
Record request language in uWSGI logs

Xyene:


quantum5 (approved):
This is safe because `DMOJLoginMiddleware` runs after `LocaleMiddleware`.

PR #2245, opened 2023-03-06
Fix PDF rendering for relative STATIC_URL

Ninjaclasher:
If `STATIC_URL` is relative, then we get a relative path in the HTML, which fails to render correctly. Use `urljoin` with the full URL to convert to an absolute URL. Should still work if `STATIC_URL` is already absolute.

```py
>>> from urllib.parse import urljoin
>>> urljoin('https://dmoj.ca/', '/static/a.js')
'https://dmoj.ca/static/a.js'
>>> urljoin('https://dmoj.ca/', 'https://static.dmoj.ca/a.js')
'https://static.dmoj.ca/a.js'
```

### DIFF
--- a/templates/problem/raw.html
+++ b/templates/problem/raw.html
@@ -1,7 +1,7 @@
 <!DOCTYPE html>
 <html>
 <head>
-    <link rel="stylesheet" href="{{ static('style.css') }}">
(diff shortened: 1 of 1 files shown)


kiritofeng:
Can you add this behaviour as a comment to the code? Otherwise LGTM

Xyene (approved):
LGTM, thanks!

PR #2248, opened 2023-03-19
Pin social-auth version to avoid CI breakage

quantum5:


kiritofeng (approved):
LGTM. As a reminder to future us, pinning is because of https://github.com/python-social-auth/social-core/pull/590#issuecomment-1471631500.

PR #2251, opened 2023-03-30
Autofocus TOTP input field when logging in

Ninjaclasher:
This is so you avoid having to click into the field before being able to enter code.


Xyene (approved):
LGTM, thanks!

Issue #2255, opened 2023-05-10
Create problem from command lines


DapperX:
For some automation reasons, I was trying to create problems from command lines.
What I did was to invoke `python manage.py create_problem ...` since I accidently found this command though it seems to not appear in the docs. 
Then I got an error saying **`TypeError: Direct assignment to the forward side of a many-to-many set is prohibited. Use types.set() instead`**, at the line `problem.types = [ProblemType.objects.get(name=options['type'])]`, in `judge/management/commands/create_problem.py`
After doing some searches online, I had a workaround that changes the code as
```
    def handle(self, *args, **options):
        problem = Problem()
        problem.code = options['code']
        problem.name = options['name']
        problem.description = options['body']
        problem.group = ProblemGroup.objects.get(name=options['group'])
        #problem.types = [ProblemType.objects.get(name=options['type'])]  # <---changes begin here
        types = ProblemType.objects.get(name=options['type'])
        problem.save()
        problem.types.set([types])  # <---changes end here
        problem.save()
```
Then the error became **`django.db.utils.IntegrityError: (1048, "Column 'time_limit' cannot be null")`** which was solved by adding a line `problem.time_limit = ...` After repeating this for the fields `.memory_limit` and `.points` and adding corresponding arguments to the parser, it eventually works, and the newly added problem seems good on the webpage.

While a few questions come to me:
1. I am not familiar with Django so I wonder if my code above is correct/safe to operate on the models. Is there a better way to do the same?
2. What is the standard way to add problems from command lines? `python manage.py create_problem` is not documented and it needs quite a few changes to be functional.

Xyene:
Thanks for the report!

> 1. I am not familiar with Django so I wonder if my code above is correct/safe to operate on the models. Is there a better way to do the same?

Your code looks fine to me. There was a change back some versions of Django that made direct assignment illegal, and it seems we forgot to update that command.

Seems like Django also got stricter around enforcing integrity and not generating defaults for `time_limit` and `memory_limit.

> 2. What is the standard way to add problems from command lines?

`python manage.py create_problem` is the way to go, we just don't use it on https://dmoj.ca so it has bitrotted over time.

I think your changes seem reasonable, would you be open to creating a pull request with them?

DapperX:
Thanks for the quick response! 
It is good to know my changes look fine. I am happy to submit a PR with them.

PR #2256, opened 2023-05-12
Make 'create_problem' command work and add options

DapperX:
Corresponding to the issue #2255 

Commit details:
- properly assign `types` to match Django's new rules
- initialize fields of `time_limit`, `memory_limit`, and `points` to be able to save the model
- add options `--partial-points` and `--all-languages` to make added problems immediately usable in automation

### DIFF
--- a/judge/management/commands/create_problem.py
+++ b/judge/management/commands/create_problem.py
@@ -1,7 +1,6 @@
 from django.core.management.base import BaseCommand
 
-from judge.models import Problem, ProblemGroup, ProblemType
-from judge.models.runtime import Language
+from judge.models import Problem, ProblemGroup, ProblemType, Language
 
 
 class Command(BaseCommand):
@@ -27,7 +26,7 @@ class Command(BaseCommand):
         problem.name = options['name']
         problem.description = options['body']
         problem.group = ProblemGroup.objects.get(name=options['group'])
-        problem.time_limit = options["time_limit"]
+        problem.time_limit = options['time_limit']
         problem.memory_limit = options["memory_limit"]
         problem.points = options["points"]
         if options['pr_pts']:
@@ -36,5 +35,5 @@ class Command(BaseCommand):
 
         problem.types.set([ProblemType.objects.get(name=options['type'])])
         if options['all_lang']:
-            problem.allowed_languages.set([l for l in Language.objects.all()])
+            problem.allowed_languages.set(Language.objects.all())
         problem.save()
--- a/judge/management/commands/create_problem.py
+++ b/judge/management/commands/create_problem.py
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 
 from judge.models import Problem, ProblemGroup, ProblemType
+from judge.models.runtime import Language
 
 
 class Command(BaseCommand):
@@ -12,6 +13,13 @@ class Command(BaseCommand):
         parser.add_argument('body', help='problem description')
         parser.add_argument('type', help='problem type')
         parser.add_argument('group', help='problem group')
+        parser.add_argument('time_limit', help='time limit in (fractional) seconds')
+        parser.add_argument('memory_limit', help='memory limit in kilobytes')
+        parser.add_argument('points', help='problem points')
+        parser.add_argument('--partial-points', help='allow partial points',
+            action='store_true', default=False, dest='pr_pts')
+        parser.add_argument('--all-languages', help='allow all languages',
+            action='store_true', default=False, dest='all_lang')
 
     def handle(self, *args, **options):
         problem = Problem()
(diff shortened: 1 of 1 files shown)


judge/management/commands/create_problem.py
```diff
@@ -1,6 +1,7 @@
 from django.core.management.base import BaseCommand
 
 from judge.models import Problem, ProblemGroup, ProblemType
+from judge.models.runtime import Language
```
Ninjaclasher, line 4: I think we should be able to import `Language` directly from `judge.models` (i.e. merge the import on this line with the previous line)

judge/management/commands/create_problem.py
```diff
@@ -12,12 +13,28 @@ def add_arguments(self, parser):
...
+            action='store_true', default=False, dest='all_lang')
 
     def handle(self, *args, **options):
         problem = Problem()
         problem.code = options['code']
         problem.name = options['name']
         problem.description = options['body']
         problem.group = ProblemGroup.objects.get(name=options['group'])
-        problem.types = [ProblemType.objects.get(name=options['type'])]
+        problem.time_limit = options["time_limit"]
```
Ninjaclasher, line 30: For consistency, let's use single quotes here (and the lines below)
```suggestion
        problem.time_limit = options['time_limit']
```

judge/management/commands/create_problem.py
```diff
@@ -12,12 +13,28 @@ def add_arguments(self, parser):
...
+        problem.memory_limit = options["memory_limit"]
+        problem.points = options["points"]
+        if options['pr_pts']:
```
Ninjaclasher, line 33: Is this `if` necessary? It seems we can just do
```py
problem.partial = options['pr_pts']
```

judge/management/commands/create_problem.py
```diff
@@ -12,12 +13,28 @@ def add_arguments(self, parser):
...
+            problem.partial = 1
+        problem.save()
+
+        problem.types.set([ProblemType.objects.get(name=options['type'])])
+        if options['all_lang']:
+            problem.allowed_languages.set([l for l in Language.objects.all()])
```
Ninjaclasher, line 39:
```suggestion
            problem.allowed_languages.set(Language.objects.all())
```
This should work and removes the unnecessary list comprehension.

judge/management/commands/create_problem.py
DapperX, line 4:
```suggestion
from judge.models import Problem, ProblemGroup, ProblemType, Language
```

DapperX: Thanks for pointing it out. It is more concise.

judge/management/commands/create_problem.py
DapperX: Tested it works fine. Thanks.

judge/management/commands/create_problem.py
DapperX: I wrote this for not assuming the default value of this field, i.e., if `Allows partial points` is enabled by default at some point, the code should still work, though I don't have strong opinions about this.

judge/management/commands/create_problem.py
```diff
@@ -12,12 +12,28 @@ def add_arguments(self, parser):
         parser.add_argument('body', help='problem description')
         parser.add_argument('type', help='problem type')
         parser.add_argument('group', help='problem group')
+        parser.add_argument('time_limit', help='time limit in (fractional) seconds')
+        parser.add_argument('memory_limit', help='memory limit in kilobytes')
+        parser.add_argument('points', help='problem points')
+        parser.add_argument('--partial-points', help='allow partial points',
+            action='store_true', default=False, dest='pr_pts')
```
DapperX, line 19:
```suggestion
                            action='store_true', default=False, dest='pr_pts')
```

judge/management/commands/create_problem.py
```diff
@@ -12,12 +12,28 @@ def add_arguments(self, parser):
...
+        parser.add_argument('--all-languages', help='allow all languages',
+            action='store_true', default=False, dest='all_lang')
```
DapperX, line 21:
```suggestion
                            action='store_true', default=False, dest='all_lang')
```

judge/management/commands/create_problem.py
```diff
@@ -1,6 +1,6 @@
 from django.core.management.base import BaseCommand
 
-from judge.models import Problem, ProblemGroup, ProblemType
+from judge.models import Problem, ProblemGroup, ProblemType, Language
```
DapperX, line 3:
```suggestion
from judge.models import Language, Problem, ProblemGroup, ProblemType
```

DapperX (commented):
Fixed for the build #5492

Ninjaclasher (approved):

PR #2257, opened 2023-05-22
UserList: sort users by id

hieplpvip:
Otherwise, `user_ranking_redirect` returns the wrong page for users at the end of the ranking table because it assumes the list is sorted by `performance_points` and `id`, and those users have `performance_points = 0`.

I've only tested this on my fork (VNOJ, too lazy to setup DMOJ), though. Please test this on DMOJ.


Xyene (approved):
LGTM, thanks!

PR #2258, opened 2023-05-23
Make "Contest Ends..." link to timeanddate.com of contest end time instead of start time

### LINKED ISSUE #2027: "Contest ends" links to timeanddate.com of contest start time
For example, on https://dmoj.ca/contest/help2022

image

the link is to https://www.timeanddate.com/worldclock/fixedtime.html?msg=Halloween%20Evaluation%20of%20Logic%20and%20Programming%202022&iso=2022-10-31T04:00:00

It should instead link to the contest end time once the contest has started.

tandr3w:
Fixes #2027

the link now goes to the contest end time if it has already started
might be a bad solution, this is my first pr ever idk

### DIFF
--- a/templates/contest/contest.html
+++ b/templates/contest/contest.html
@@ -27,7 +27,8 @@
 {% block body %}
     <div id="banner">
         <a href="https://www.timeanddate.com/worldclock/fixedtime.html?msg={{ contest.name|urlquote('') }}&amp;iso=
-                {{- contest.start_time|utc|date('Y-m-d\TH:i:s') }}" class="date">
+                {% if contest.start_time > now %} {{- contest.start_time|utc|date('Y-m-d\TH:i:s') }}
+                {% else %} {{- contest.end_time|utc|date('Y-m-d\TH:i:s') }} {% endif %}" class="date">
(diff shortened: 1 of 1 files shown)


templates/contest/contest.html
```diff
@@ -27,7 +27,8 @@
 {% block body %}
     <div id="banner">
         <a href="https://www.timeanddate.com/worldclock/fixedtime.html?msg={{ contest.name|urlquote('') }}&amp;iso=
-                {{- contest.start_time|utc|date('Y-m-d\TH:i:s') }}" class="date">
+                {% if contest.start_time > now %} {{- contest.start_time|utc|date('Y-m-d\TH:i:s') }}
```
Xyene, line 30: I think we can write this using a ternary, with

```
{{ (... if ... else ...)|utc|... }}
```

so that we don't have to write the `|utc|date...` filter twice.

Xyene: Also a small nit: commit names should be descriptive about what you are fixing, e.g. "Make contest timeanddate link to end time after contest has started".

Xyene:
Thanks!

PR #2259, opened 2023-06-12
Relabel the version matrix page to just "Versions"

Xyene:
We had an inconsistency in "Version Matrix" versus "Version matrix", but in fixing it I figured just calling this tab "Versions" makes sense.

### DIFF
--- a/judge/views/status.py
+++ b/judge/views/status.py
@@ -105,7 +105,7 @@ def version_matrix(request):
 
     languages = sorted(languages, key=lambda lang: version.parse(lang.name))
     return render(request, 'status/versions.html', {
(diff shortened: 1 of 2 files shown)


judge/views/status.py
```diff
@@ -105,7 +105,7 @@ def version_matrix(request):
 
     languages = sorted(languages, key=lambda lang: version.parse(lang.name))
     return render(request, 'status/versions.html', {
-        'title': _('Version matrix'),
+        'title': _('Version'),
```
hieplpvip, line 108: Shouldn't this be `_('Versions')` ?

Issue #1052, opened 2019-09-17
Mismatching points and non-consecutive ranks


TheAvidDev:
So, this could be two (three?) issues in one:

1. Users listed on https://dmoj.ca/users/?page=387 (any any low-rank page) all have the same rank. The same goes for rank by points on user's pages.

2. Users on the previous page have 100 less points, instead of 1 less or the same: https://dmoj.ca/users/?page=386 (non-consecutive rankings)

3. A user on the last page is shown as having rank `#38601` but when going to their user page, the rank by points is `12068`. (This is probably not a bug)

quantum5:
The only bug is the pagination offset being used as the starting rank on the page, instead of number of users with more points.

PR #2264, opened 2023-06-23
Fixed Organization Pagination Rank Reset On Each Page

### LINKED ISSUE #2224: Organization user list pagination rank is reset on each page
https://dmoj.ca/organization/96-dmpg-18/users?page=2

Rank numbering should continue from the previous page, not reset back to 1 on each page.

jazyz:
Closes #2224


Ninjaclasher:
This solution will make us have the same issue as #1052 on the organization users list. We should probably just do it correctly instead of copying what is done for the users list.

PR #2267, opened 2023-07-06
Fix the favicons URL to respect STATIC_URL setting

munhyunsu:
In `template/base.html`, the URL of favicons was hardcoded to `/`
The administrator can change `STATIC_URL` in `dmoj/local_settings.py`, so I think the favicons url should follow these settings.

### DIFF
--- a/templates/base.html
+++ b/templates/base.html
@@ -11,22 +11,22 @@
     {% endif %}
     <meta id="viewport" name="viewport" content="width=device-width, initial-scale=1">
     <!-- Favicons-->
-    <link rel="apple-touch-icon" sizes="57x57" href="/apple-touch-icon-57x57.png">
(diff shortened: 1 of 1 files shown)


Xyene:
Thanks for the PR. However, the hardcoding is intentional; `manifest.json` itself is not a dynamically generated file, and references icons from the root of the website.

https://github.com/DMOJ/online-judge/blob/64a1eca408c770c7ad26b604aa9cc265a1ddd6f3/resources/icons/manifest.json#L5

So, a DMOJ install that isn't serving these icons from `/` is already broken, even with this PR applied. I don't think it's worth doing anything more complicated here (like generating the `manifest.json` dynamically). The sample nginx configs provided in the docs already map `/icons` to the root of the domain: https://github.com/DMOJ/docs/blob/15ce3a0d6397e81c3a903ecca7184bce2441d17e/sample_files/nginx.conf#L24-L27

PR #2272, opened 2023-07-18
Make user profile picture look better in dark mode

RunTheBot:
# Problem:
Background not inherited from body.

# Solution

Remove the pointless background tag, as in light mode, the tag is also inherited from body

## Dark
Before:
image

ewwwww

After:
image

Yay my pfp works

## Light
Before:
image

After:
image

no change :)


int-y1:
This wouldn't work for https://dmoj.ca/user/Pleedoh

PR #2275, opened 2023-08-15
Add unordered checker as an option

MysticalX21:
This adds the unordered checker to the dropdown containing the list of available checkers to use for a problem.


Ninjaclasher:
The `unordered` checker is deprecated. Instead, you should be using the `sorted` checker with the `split_on` parameter set to `whitespace`.

I would be open to a PR that implements adding a `split_on` parameter for the `sorted` checker into the web UI.

PR #2277, opened 2023-08-23
Fix #2276

### LINKED ISSUE #2276: Incorrect time displayed for submissions and comments
When I check the time of a submission or comment, it displays it in UTC instead of EDT, even thought I selected Toronto as my time zone.

61izzy:
Probably needs review

### DIFF
--- a/judge/jinja2/datetime.py
+++ b/judge/jinja2/datetime.py
@@ -4,6 +4,7 @@ from django.template.defaultfilters import date, time
 from django.templatetags.tz import localtime
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
+from django.utils import timezone
 from django.utils.timezone import utc
 from django.utils.translation import gettext as _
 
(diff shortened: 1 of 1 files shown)


judge/jinja2/datetime.py
```diff
@@ -26,7 +27,7 @@ def wrapper(datetime, *args, **kwargs):
 
 @registry.function
 def relative_time(time, **kwargs):
-    abs_time = date(time, kwargs.get('format', _('N j, Y, g:i a')))
+    abs_time = date(time.astimezone(timezone.get_current_timezone()), kwargs.get('format', _('N j, Y, g:i a')))
```
Ninjaclasher, line 30: Does this work when the user is logged out? Ideally, we should use the browser's timezone when they are logged out.

61izzy: I think that might have to be fixed separately in timezone.py

PR #2278, opened 2023-08-24
Fix AttributeError when rendering translated problem PDFs

Ninjaclasher:
Seems like a typo from 3e597c03b.


Xyene (approved):
LGTM, thanks!

PR #2280, opened 2023-08-26
Improve import locations

int-y1:
this fixes 1% of mypy errors


Riolku:
Why does this fix mypy errors? Because of the locations of type stubs?

int-y1:
> Why does this fix mypy errors? Because of the locations of type stubs?

mypy couldn't find the item in the module

```
judge/highlight_code.py:1: error: Module "django.utils.html" has no attribute "mark_safe"  [attr-defined]
judge/admin/interface.py:3: error: Module "django.contrib.flatpages.admin" has no attribute "FlatpageForm"  [attr-defined]
judge/views/contests.py:27: error: Module "django.views.generic.detail" does not explicitly export attribute "View"  [attr-defined]
```

kiritofeng (approved):
LGTM, thansk!

PR #2281, opened 2023-08-26
Use decorator for admin display functions

int-y1:
This moves every `short_description` and `admin_order_field` into `@admin.display`


quantum5 (approved):
Looks much better now.

PR #2282, opened 2023-08-26
Use consistent capitals for admin column descriptions

int-y1:


Ninjaclasher (approved):
LGTM

quantum5 (commented):
Pretty sure Django automatically capitalizes these fields. See https://github.com/django/django/blob/11920e77959deaa65eb86ccc5d39da903fd3dd41/django/contrib/admin/templates/admin/change_list_results.html#L21. The actual database fields are not capitalized either.

int-y1:
> Pretty sure Django automatically capitalizes these fields. See https://github.com/django/django/blob/11920e77959deaa65eb86ccc5d39da903fd3dd41/django/contrib/admin/templates/admin/change_list_results.html#L21. The actual database fields are not capitalized either.

This convention also works. I didn't see anything wrong on the site.

Issue #1493, opened 2020-08-02
Add mypy type annotations


quantum5:
We should start adding mypy annotations to the codebase, preferably starting with the stuff we already have unit tests for.

PR #2283, opened 2023-08-27
Add mypy annotations

Ninjaclasher:
Supercedes #1495. Refs #1493.

Very much a WIP.


dmoj/settings.py
```diff
@@ -11,152 +11,159 @@
...
 import datetime
 import os
+from typing import Any, Dict, List, Literal, Optional, Sequence, Set, TYPE_CHECKING, Tuple, Union
 
 from django.utils.translation import gettext_lazy as _
 from django_jinja.builtins import DEFAULT_EXTENSIONS
 from jinja2 import select_autoescape
 
+import django_stubs_ext  # noqa: I100, I102, I202
+django_stubs_ext.monkeypatch()
```
int-y1, line 21: is it necessary to keep `monkeypatch()` out of `TYPE_CHECKING`?

judge/management/commands/generate_sitemap.py
```diff
@@ -10,7 +10,7 @@
 
 
 class Command(BaseCommand):
-    requires_system_checks = False
+    requires_system_checks = ()
```
int-y1, line 13: what does this change do?

judge/views/submission.py
```diff
@@ -576,11 +577,11 @@ def _get_result_data(self, queryset=None):
 
 
 class ForceContestMixin(object):
-    @property
+    @cached_property
     def in_contest(self):
         return True
 
-    @property
+    @cached_property
```
int-y1, line 584: what does this change do?
does mypy complain less with `@cached_property`?

.github/workflows/build.yml
```diff
@@ -5,24 +5,47 @@ jobs:
...
+      with:
+        python-version: 3.8
+    - name: Cache pip
+      uses: actions/cache@v2
+      with:
+        path: ~/.cache/pip
+        key: pip-${{ runner.os }}-${{ secrets.CACHE_VERSION }}-${{ hashFiles('requirements.txt') }}-${{ hashFiles('requirements-mypy.txt') }}
+    - name: Install mypy dependencies
+      run: |
+        pip install mypy
```
int-y1, line 33: should you move `mypy` into a requirements file?

judge/views/submission.py
Ninjaclasher: These two properties override the ones in `SubmissionsListBase`, which are `cached_property`s.

judge/management/commands/generate_sitemap.py
Ninjaclasher: According to [docs](https://docs.djangoproject.com/en/3.2/howto/custom-management-commands/#django.core.management.BaseCommand.requires_system_checks), there was a change to make this a list of tags instead of a boolean in Django 3.2. We missed it when upgrading.

dmoj/settings.py
Ninjaclasher: Yeah, we need it when running in production as well. Otherwise, we still get the same `TypeError` (ref: https://github.com/typeddjango/django-stubs#i-cannot-use-queryset-or-manager-with-type-annotations)

.github/workflows/build.yml
Ninjaclasher: Sure, I'll move it to `requirements-mypy.txt`.

dmoj/settings.py
```diff
@@ -11,152 +11,159 @@
...
+DMOJ_REQUIRE_STAFF_2FA: bool = True
 # Display warnings that admins will not perform 2FA recovery.
-DMOJ_2FA_HARDCORE = False
+DMOJ_2FA_HARDCORE: bool = False
 
 # Set to 1 to use HTTPS if request was made to https://
 # Set to 2 to always use HTTPS for links
 # Set to 0 to always use HTTP for links
-DMOJ_SSL = 0
+DMOJ_SSL: int = 0
```
kiritofeng, line 51: Should we use `Literal[0,1,2]` instead of `int?

dmoj/settings.py
```diff
@@ -11,152 +11,159 @@
...
+DMOJ_SUBMISSION_LIMIT: int = 2
+DMOJ_SUBMISSIONS_REJUDGE_LIMIT: int = 10
 
 # Whether to allow users to view source code: 'all' | 'all-solved' | 'only-own'
-DMOJ_SUBMISSION_SOURCE_VISIBILITY = 'all-solved'
-DMOJ_BLOG_NEW_PROBLEM_COUNT = 7
-DMOJ_TOTP_TOLERANCE_HALF_MINUTES = 1
-DMOJ_SCRATCH_CODES_COUNT = 5
-DMOJ_USER_MAX_ORGANIZATION_COUNT = 3
+DMOJ_SUBMISSION_SOURCE_VISIBILITY: Union[Literal['all'], Literal['all-solved'], Literal['only-own']] = 'all-solved'
```
kiritofeng, line 87: Does `Literal['all', 'all-solved', 'only-own']` not work?

judge/utils/problem_data.py
```diff
@@ -10,7 +10,9 @@
 from django.utils.translation import gettext as _
 
 if os.altsep:
-    def split_path_first(path, repath=re.compile('[%s]' % re.escape(os.sep + os.altsep))):
+    repath = re.compile('[%s]' % re.escape(os.sep + os.altsep))
+
+    def split_path_first(path):
```
kiritofeng, line 15: This change seems semantically different to me; are you sure we don't (and won't need to) pass `repath` as a argument to this function?

judge/views/comment.py
```diff
@@ -87,9 +89,9 @@ def downvote_comment(request):
 
 
 class CommentMixin(object):
-    model = Comment
+    model: type = Comment
```
kiritofeng, line 92: Out of curiosity should this really be `type` and not `Model` (or a subclass)?

.github/workflows/build.yml
int-y1, line 1: needs an update due to #2362

PR #2290, opened 2023-10-11
Implement user email change functionality

### LINKED ISSUE #1996: Allow Changing Emails
Currently, we have users email an admin email to change their email. We could, and really, should, automate this process, by sending a verification to both emails... Or maybe just the new one?

We should consider a case where a user has lost access to their old email but still has access to their account through, say, their password. Do we want to support that as a common case or should we fall back to emailing?

Ninjaclasher:
Fixes #1996.

Request email change process:
1. User goes to email change endpoint with a form that consists of a password and email. This endpoint is accessible via edit profile.
2. Verify password and verify email is not used or blacklisted.
3. Create a base64-encoded key that is {user id, new email} signed by SECRET_KEY and create an endpoint that accepts this key. This key will contains a timestamp, which will be used to check expiry.
4. Send this URL to the new email, send a security notice to the old email (email change requested).

Confirm email change process:
1. User clicks the URL in the email.
2. Verify signing, verify expiry timestamp, verify user id matches currently logged in user.
3. Verify again that new email is not used and update email in the database (atomically).

### DIFF
--- a/dmoj/settings.py
+++ b/dmoj/settings.py
@@ -107,8 +107,15 @@ DMOJ_STATS_SUBMISSION_RESULT_COLORS = {
 }
 DMOJ_API_PAGE_SIZE = 1000
 
-DMOJ_PASSWORD_RESET_LIMIT_WINDOW = 3600
+# Number of password resets per window (in minutes)
+DMOJ_PASSWORD_RESET_LIMIT_WINDOW = 60
 DMOJ_PASSWORD_RESET_LIMIT_COUNT = 10
+
+# Number of email change requests per window (in minutes)
+DMOJ_EMAIL_CHANGE_LIMIT_WINDOW = 60
+DMOJ_EMAIL_CHANGE_LIMIT_COUNT = 10
+# Number of minutes before an email change request activation key expires
+DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES = 60
 
 # At the bare minimum, dark and light theme CSS file locations must be declared
 DMOJ_THEME_CSS = {
@@ -600,6 +607,9 @@ except IOError:
 
 # Check settings are consistent
 assert DMOJ_PROBLEM_MIN_USER_POINTS_VOTE >= DMOJ_PROBLEM_MIN_PROBLEM_POINTS
+
+# <= 1 minute expiry is unusable UX
+assert DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES > 1
 
 if DMOJ_PDF_PDFOID_URL:
     # If a cache is configured, it must already exist and be a directory
--- a/judge/forms.py
+++ b/judge/forms.py
@@ -6,6 +6,7 @@ import webauthn
 from django import forms
 from django.conf import settings
 from django.contrib.auth.forms import AuthenticationForm
+from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator
 from django.db.models import Q
@@ -17,6 +18,7 @@ from django.utils.translation import gettext_lazy as _, ngettext_lazy
 from django_ace import AceWidget
 from judge.models import Contest, Language, Organization, Problem, ProblemPointsVote, Profile, Submission, \
     WebAuthnCredential
+from judge.utils.mail import validate_email_domain
 from judge.utils.subscription import newsletter_id
 from judge.widgets import HeavyPreviewPageDownWidget, Select2MultipleWidget, Select2Widget
 
@@ -93,6 +95,26 @@ class ProfileForm(ModelForm):
         if not self.fields['organizations'].queryset:
             self.fields.pop('organizations')
         self.fields['user_script'].widget = AceWidget(mode='javascript', theme=user.profile.resolved_ace_theme)
+
+
+class EmailChangeForm(Form):
+    password = CharField(widget=forms.PasswordInput())
+    email = forms.EmailField()
+
+    def __init__(self, *args, user, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.user = user
+
+    def clean_email(self):
+        if User.objects.filter(email=self.cleaned_data['email']).exists():
+            raise ValidationError(_('This email address is already taken.'))
+        validate_email_domain(self.cleaned_data['email'])
+        return self.cleaned_data['email']
+
+    def clean_password(self):
+        if not self.user.check_password(self.cleaned_data['password']):
+            raise ValidationError(_('Invalid password'))
+        return self.cleaned_data['password']
 
 
 class DownloadDataForm(Form):
--- a/judge/utils/mail.py
+++ b/judge/utils/mail.py
@@ -1,0 +1,42 @@
+import re
+from typing import Any, Dict, List, Optional, Pattern
+
+from django.conf import settings
+from django.core.mail import EmailMultiAlternatives
+from django.forms import ValidationError
+from django.template import loader
+from django.utils.translation import gettext
+
+
+bad_mail_regex: List[Pattern[str]] = list(map(re.compile, settings.BAD_MAIL_PROVIDER_REGEX))
+
+
+def validate_email_domain(email: str) -> None:
+    if '@' in email:
+        domain = email.split('@')[-1].lower()
+        if domain in settings.BAD_MAIL_PROVIDERS or any(regex.match(domain) for regex in bad_mail_regex):
+            raise ValidationError(gettext('Your email provider is not allowed due to history of abuse. '
+                                          'Please use a reputable email provider.'))
+
+
+# Inspired by django.contrib.auth.forms.PasswordResetForm.send_mail
+def send_mail(
+    context: Dict[str, Any],
+    *,
+    from_email: Optional[str] = None,
+    to_email: str,
+    subject_template_name: str,
+    email_template_name: str,
+    html_email_template_name: Optional[str] = None,
+) -> None:
+    subject = loader.render_to_string(subject_template_name, context)
+    # Email subject *must not* contain newlines
+    subject = ''.join(subject.splitlines())
+    body = loader.render_to_string(email_template_name, context)
+
+    email_message = EmailMultiAlternatives(subject, body, from_email, [to_email])
+    if html_email_template_name is not None:
+        html_email = loader.render_to_string(html_email_template_name, context)
+        email_message.attach_alternative(html_email, 'text/html')
+
+    email_message.send()
--- a/judge/views/register.py
+++ b/judge/views/register.py
@@ -1,6 +1,4 @@
 # coding=utf-8
-import re
-
 from django import forms
 from django.conf import settings
 from django.contrib.auth.models import User
@@ -14,11 +12,10 @@ from registration.forms import RegistrationForm
 from sortedm2m.forms import SortedMultipleChoiceField
 
 from judge.models import Language, Organization, Profile, TIMEZONE
+from judge.utils.mail import validate_email_domain
 from judge.utils.recaptcha import ReCaptchaField, ReCaptchaWidget
 from judge.utils.subscription import Subscription, newsletter_id
 from judge.widgets import Select2MultipleWidget, Select2Widget
-
-bad_mail_regex = list(map(re.compile, settings.BAD_MAIL_PROVIDER_REGEX))
 
 
 class CustomRegistrationForm(RegistrationForm):
@@ -43,12 +40,7 @@ class CustomRegistrationForm(RegistrationForm):
         if User.objects.filter(email=self.cleaned_data['email']).exists():
             raise forms.ValidationError(gettext('The email address "%s" is already taken. Only one registration '
                                                 'is allowed per address.') % self.cleaned_data['email'])
-        if '@' in self.cleaned_data['email']:
-            domain = self.cleaned_data['email'].split('@')[-1].lower()
-            if (domain in settings.BAD_MAIL_PROVIDERS or
-                    any(regex.match(domain) for regex in bad_mail_regex)):
-                raise forms.ValidationError(gettext('Your email provider is not allowed due to history of abuse. '
-                                                    'Please use a reputable email provider.'))
+        validate_email_domain(self.cleaned_data['email'])
         return self.cleaned_data['email']
 
     def clean_organizations(self):
--- a/judge/views/user.py
+++ b/judge/views/user.py
@@ -1,3 +1,5 @@
+import base64
+import binascii
 import itertools
 import json
 import os
@@ -8,9 +10,11 @@ from django.conf import settings
 from django.contrib.auth import logout as auth_logout
 from django.contrib.auth.decorators import login_required
 from django.contrib.auth.mixins import LoginRequiredMixin
-from django.contrib.auth.models import Permission
+from django.contrib.auth.models import Permission, User
 from django.contrib.auth.views import LoginView, PasswordChangeView, PasswordResetView, redirect_to_login
 from django.contrib.contenttypes.models import ContentType
+from django.contrib.sites.shortcuts import get_current_site
+from django.core import signing
 from django.core.cache import cache
 from django.core.exceptions import PermissionDenied, ValidationError
 from django.db.models import Count, Max, Min
@@ -27,13 +31,14 @@ from django.views.decorators.http import require_POST
 from django.views.generic import DetailView, FormView, ListView, TemplateView, View
 from reversion import revisions
 
-from judge.forms import CustomAuthenticationForm, DownloadDataForm, ProfileForm, newsletter_id
+from judge.forms import CustomAuthenticationForm, DownloadDataForm, EmailChangeForm, ProfileForm, newsletter_id
 from judge.models import Profile, Submission
 from judge.performance_points import get_pp_breakdown
 from judge.ratings import rating_class, rating_progress
 from judge.tasks import prepare_user_data
 from judge.utils.celery import task_status_by_id, task_status_url_by_id
 from judge.utils.infinite_paginator import InfinitePaginationMixin
+from judge.utils.mail import send_mail
 from judge.utils.problems import contest_completed_ids, user_completed_ids
 from judge.utils.pwned import PwnedPasswordsValidator
 from judge.utils.ranker import ranker
@@ -494,6 +499,9 @@ class UserLogoutView(TitleMixin, TemplateView):
         return HttpResponseRedirect(request.get_full_path())
 
 
+MINUTES_TO_SECONDS = 60
+
+
 class CustomPasswordResetView(PasswordResetView):
     template_name = 'registration/password_reset.html'
     html_email_template_name = 'registration/password_reset_email.html'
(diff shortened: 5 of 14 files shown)


kiritofeng:
Can we also add a 2FA check before changing the email? (I believe the password change form doesn't require 2FA verification but I think we should also add 2FA to that).

Ninjaclasher:
After some more thought, the issue with requiring 2fa is that we need to support all the different types of 2fa (backup code, totp, hardware key), which requires quite a bit of refactoring. Also, at the moment, adding 2fa won't increase security, since a user can regenerate their totp token/get new backup codes without any re-authentication.

I think adding 2fa to this should be part of a future change where we harden everything together, and is out of the scope of the current PR.

judge/views/user.py
```diff
@@ -507,3 +513,121 @@ def post(self, request, *args, **kwargs):
...
+            except (binascii.Error, signing.BadSignature):
+                raise ValueError(_('Invalid activation key. Please try again.'))
+            except signing.SignatureExpired:
+                raise ValueError(_('This request is expired. Please try again.'))
+            if data['id'] != request.user.id:
+                raise ValueError(
+                    _('Please try again from the account this email change was originally requested from.'),
+                )
+            with revisions.create_revision(atomic=True):
+                if User.objects.filter(email=data['email']).exists():
```
kiritofeng, line 616: Should this go outside the revision?

judge/forms.py
```diff
@@ -95,6 +97,26 @@ def __init__(self, *args, **kwargs):
...
+class EmailChangeForm(Form):
+    password = CharField(widget=forms.PasswordInput())
+    email = forms.EmailField()
+
+    def __init__(self, *args, user, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.user = user
+
+    def clean_email(self):
+        if User.objects.filter(email=self.cleaned_data['email']).exists():
```
kiritofeng, line 109: What about deactivated accounts?

templates/registration/email_change.html
```diff
@@ -0,0 +1,23 @@
+{% extends "base.html" %}
```
kiritofeng, line 1: Would it be possible to see a screenshot of these emails (mostly to check they render correctly and don't have funny messed up links.)

dmoj/settings.py
```diff
@@ -110,6 +110,10 @@
 DMOJ_PASSWORD_RESET_LIMIT_WINDOW = 3600
 DMOJ_PASSWORD_RESET_LIMIT_COUNT = 10
 
+DMOJ_EMAIL_CHANGE_LIMIT_WINDOW = 3600
```
kiritofeng, line 113: Might want to add comments for units / explaining purpose (my first thought was `DMOJ_EMAIL_CHANGE_LIMIT_COUNT` was a hard cap on the number of times you could change your email in total.

judge/views/user.py
```diff
@@ -507,3 +513,121 @@ def post(self, request, *args, **kwargs):
...
+    form_class = EmailChangeForm
+
+    activate_html_email_template_name = 'registration/email_change_activate_email.html'
+    activate_email_template_name = 'registration/email_change_activate_email.txt'
+    activate_subject_template_name = 'registration/email_change_activate_subject.txt'
+    notify_html_email_template_name = 'registration/email_change_notify_email.html'
+    notify_email_template_name = 'registration/email_change_notify_email.txt'
+    notify_subject_template_name = 'registration/email_change_notify_subject.txt'
+
+    def send_mail(
```
kiritofeng, line 530: Should this method be here, or should we refactor it out into some util file?

kiritofeng (commented):
Would also like @quantum5 and/or @Xyene to take a look.

judge/forms.py
Ninjaclasher: I believe the current check (during registration) includes all users, including deactivated ones. Should we allow users to reuse emails from deactivated emails?

kiritofeng: I think a better solution might be to set the emails of deactivated accounts to a dummy email and leave the current checks intact, cc @quantum5 @Xyene for thoughts.

judge/views/user.py
Ninjaclasher: The idea was to make this atomic (since the revision block is atomic), so we avoid any races between when we check and when we update the email address. Email uniqueness is not guaranteed on the database level, so we have to do this. If you have any suggestions for how to make it better, please let me know.

Ninjaclasher:
Screenshots
image
image
image
image

templates/registration/email_change.html
Ninjaclasher: I've attached some screenshots of both the HTML email and the plaintext email.

judge/views/user.py
kiritofeng: That is a good point; race conditions hard :(

judge/views/user.py
Ninjaclasher: Good point, I've refactored it.

dmoj/settings.py
```diff
@@ -107,9 +107,16 @@
...
 
+# Number of password reset per window (in seconds)
 DMOJ_PASSWORD_RESET_LIMIT_WINDOW = 3600
 DMOJ_PASSWORD_RESET_LIMIT_COUNT = 10
 
+# Number of email change requests per window (in seconds)
+DMOJ_EMAIL_CHANGE_LIMIT_WINDOW = 3600
+DMOJ_EMAIL_CHANGE_LIMIT_COUNT = 10
+# Number of minutes before an email change request activation key expires
+DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES = 10
```
Riolku, line 118: I'm mildly more in favour of a default around 1 hour.

judge/forms.py
```diff
@@ -95,6 +97,28 @@ def __init__(self, *args, **kwargs):
         self.fields['user_script'].widget = AceWidget(mode='javascript', theme=user.profile.resolved_ace_theme)
 
 
+class EmailChangeForm(Form):
+    password = CharField(widget=forms.PasswordInput())
```
Riolku, line 101: Obviously, all this explodes for social auth. Do we have a notice about this? What if a user has social auth but also a password?

judge/utils/mail.py
```diff
@@ -0,0 +1,36 @@
...
+from django.core.mail import EmailMultiAlternatives
+from django.template import loader
+
+
+bad_mail_regex = list(map(re.compile, settings.BAD_MAIL_PROVIDER_REGEX))
+
+
+def is_email_address_bad(email):
+    if '@' in email:
+        domain = email.split('@')[-1].lower()
```
Riolku, line 13: What does the email RFC say about multiple @ in email addresses?

judge/forms.py
```diff
@@ -95,6 +97,28 @@ def __init__(self, *args, **kwargs):
...
+    def __init__(self, *args, user, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.user = user
+
+    def clean_email(self):
+        if User.objects.filter(email=self.cleaned_data['email']).exists():
+            raise ValidationError(_('This email address is already taken.'))
+        if is_email_address_bad(self.cleaned_data['email']):
+            raise ValidationError(_('Your email provider is not allowed due to history of abuse. '
+                                    'Please use a reputable email provider.'))
```
Riolku, line 113: The function name doesnt match the error message. Rename the function to something like `is_email_address_bad_domain` maybe.

judge/views/register.py
```diff
@@ -43,12 +40,9 @@ def clean_email(self):
...
                                                 'is allowed per address.') % self.cleaned_data['email'])
-        if '@' in self.cleaned_data['email']:
-            domain = self.cleaned_data['email'].split('@')[-1].lower()
-            if (domain in settings.BAD_MAIL_PROVIDERS or
-                    any(regex.match(domain) for regex in bad_mail_regex)):
-                raise forms.ValidationError(gettext('Your email provider is not allowed due to history of abuse. '
-                                                    'Please use a reputable email provider.'))
+        if is_email_address_bad(self.cleaned_data['email']):
+            raise forms.ValidationError(gettext('Your email provider is not allowed due to history of abuse. '
+                                                'Please use a reputable email provider.'))
```
Riolku, line 45: Let's refactor this common code. Potentially make a function called `validate_not_bad_email_domain` or something.

judge/views/user.py
```diff
@@ -507,3 +512,98 @@ def post(self, request, *args, **kwargs):
...
+            'site_name': current_site.name,
+            'protocol': 'https' if self.request.is_secure() else 'http',
+            'site_admin_email': settings.SITE_ADMIN_EMAIL,
+            'expiry_minutes': settings.DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES,
+            'user': self.request.user,
+            'activation_key': activation_key,
+            'new_email': new_email,
+        }
+        send_mail(
+            self.notify_subject_template_name, self.notify_email_template_name, context, None, self.request.user.email,
```
Riolku, line 549: Why is from_email None? Also, maybe we should mark the parameter so it's clear what is being passed as None.

judge/views/user.py
```diff
@@ -507,3 +512,98 @@ def post(self, request, *args, **kwargs):
...
+        try:
+            try:
+                data = signer.unsign_object(
+                    base64.urlsafe_b64decode(activation_key.encode()).decode(),
+                    max_age=settings.DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES * 60,
+                )
+            except (binascii.Error, signing.BadSignature):
+                raise ValueError(_('Invalid activation key. Please try again.'))
+            except signing.SignatureExpired:
+                raise ValueError(_('This request is expired. Please try again.'))
```
Riolku, line 590: This request has expired

judge/views/user.py
```diff
@@ -507,3 +512,98 @@ def post(self, request, *args, **kwargs):
...
+            if data['id'] != request.user.id:
+                raise ValueError(
+                    _('Please try again from the account this email change was originally requested from.'),
```
Riolku, line 593: I feel like an email change confirmation shouldn't require a login. I often click on confirmation links on my phone that i requested on my computer.

judge/views/user.py
```diff
@@ -507,3 +512,98 @@ def post(self, request, *args, **kwargs):
...
+                )
+            with revisions.create_revision(atomic=True):
+                if User.objects.filter(email=data['email']).exists():
+                    raise ValueError(_('The email originally requested is in use. Please try again with a new email.'))
+                request.user.email = data['email']
+                request.user.save()
+                revisions.set_user(request.user)
+                revisions.set_comment(_('Changed email address'))
+        except ValueError as ve:
+            return generic_message(request, _('Email change failed'), str(ve), status=403)
```
Riolku, line 603: When does this happen?

Riolku (commented):
Will have to review HTML another time, or leave to someone else.

templates/registration/email_change_notify_email.html
```diff
@@ -0,0 +1,24 @@
...
+{{ _('If this was you, no further action is required.') }}
+
+<br>
+<b>
+{% if site_admin_email %}
+{% with link='<a href="mailto:%(email)s">%(email)s</a>'|safe|format(email=site_admin_email) %}
+{{ _('If this was not you, please email us immediately at %(email)s.', email=link) }}
+{% endwith %}
+{% else %}
+{{ _('If this was not you, please reply to this email immediately.') }}
```
Riolku, line 20: Shouldnt we suggest changing pwd as well?

templates/registration/email_change_activate_email.html
```diff
@@ -0,0 +1,24 @@
+<div style="border:2px solid #fd0;margin:4px 0;"><div style="background:#000;border:6px solid #000;">
```
Riolku, line 1: Is this HTML copied from somewhere?

Ninjaclasher: Yes, it's copied from the password reset/account activation emails.

judge/views/user.py
Ninjaclasher: Any of the above errors that get raised are caught here and returned as a generic message.

judge/views/register.py
Ninjaclasher: This is what I had originally, but I dislike having a function `validate_...` that doesn't return anything and can only raise. I'm open to other suggestions though.

judge/utils/mail.py
Ninjaclasher: It looks like you can only have multiple @ in the username part of the email address, while the domain part cannot contain a @. I think this should be fine since we can assume anything after the last part is the domain.

judge/views/register.py
Ninjaclasher: On further thought, I'll change back to `validate_not_bad_email_domain`, since it looks like Django itself uses a similar construct.

templates/registration/email_change_notify_email.html
Ninjaclasher: Good idea, updated.

judge/views/user.py
Ninjaclasher: Done, also added a comment to explain why. When `from_email` is none, then the `DEFAULT_FROM_EMAIL` setting is used.

judge/forms.py
Ninjaclasher: We don't right now, but we also don't have anything set up for password change. In a future PR, we should just hide these endpoints if the user has no password set and say something about using social auth.

Right now, I'm of the opinion that if a user is using social auth but has a password set, then they should be treated as a normal user and be allowed to change their emails.

judge/views/user.py
Ninjaclasher: Hmm, I'm open to discussion on this. I do think there's some merit to ensuring the user is logged in before switching emails. This prevents a user mistyping an email that is actually in use, and then the user of the mistyped email accidentally clicking on the link causing the email to be switched.

judge/forms.py
Riolku: What happens when a user with social auth changes their email? Does their social auth still work? If so I think it's probably fine.

judge/views/register.py
Riolku: I think `throw_if_bad_email_domain` could work too, but don't feel like bikeshedding much more.

judge/forms.py
Ninjaclasher: Yes it should be fine. I don't believe social auth uses this email field after initially populating it.

judge/views/user.py
Riolku: Like you mean user A requests a change to email B, and the owner of B can now take over the account of A...

Ok, granted, that seems bad. I think this is fine then.

judge/views/user.py
Riolku: Do we have this pattern elsewhere?

I'm tempted to say we should move the logic of changing to its own function, and make the displaying be in `get()`.

Also, is there a method that makes more sense than `str(ve)`

Riolku: Finally can we use a dedicated exception instead of `ValueError`? I thought we failed some conversion or smth.

Ninjaclasher: Changed to a `EmailChangedError`. I think `str(e)` is the standard way to retrieve the message from an error, unless I am mistaken.

Riolku: Are you against moving the logic into its own function to avoid the double try?

Ninjaclasher: I don't really see the point, since it'll only ever be used once, but I'll move it to a separate function if it makes it easier to read the code.

Riolku: Other people can comment on this.. I could live with the current code.

Riolku (approved):
LGTM aside from the minor bikeshedding on code readability.

EDIT: Oops seems I missed the recent update.

Either way, thanks for working on this!

judge/views/user.py
Ninjaclasher: I've moved it to a separate function.

judge/forms.py
Xyene: Agree, I think this mirroring the registration logic is good.

judge/utils/mail.py
```diff
@@ -0,0 +1,39 @@
+import re
```
Xyene, line 1: Since this is brand new code, let's add type annotations.

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+            'protocol': 'https' if self.request.is_secure() else 'http',
+            'site_admin_email': settings.SITE_ADMIN_EMAIL,
+            'expiry_minutes': settings.DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES,
+            'user': self.request.user,
+            'activation_key': activation_key,
+            'new_email': new_email,
+        }
+        # When from_email is none, we use the DEFAULT_FROM_EMAIL setting
+        send_mail(
+            self.notify_subject_template_name, self.notify_email_template_name, context, from_email=None,
```
Xyene, line 550: If you are going to pass these as kwargs, then you should make it explicitly required to do so by adding `*` to the parameter list of `send_mail`.

judge/utils/mail.py
```diff
@@ -0,0 +1,39 @@
...
+    if '@' in email:
+        domain = email.split('@')[-1].lower()
+        if domain in settings.BAD_MAIL_PROVIDERS or any(regex.match(domain) for regex in bad_mail_regex):
+            raise ValidationError(gettext('Your email provider is not allowed due to history of abuse. '
+                                          'Please use a reputable email provider.'))
+
+
+def send_mail(
+    subject_template_name,
+    email_template_name,
```
Xyene, line 23: The ordering of these parameters is a little strange, it'd be good to keep everything template-related together, not separated by email parameters.

dmoj/settings.py
```diff
@@ -107,9 +107,16 @@
 }
 DMOJ_API_PAGE_SIZE = 1000
 
+# Number of password resets per window (in seconds)
 DMOJ_PASSWORD_RESET_LIMIT_WINDOW = 3600
 DMOJ_PASSWORD_RESET_LIMIT_COUNT = 10
 
+# Number of email change requests per window (in seconds)
+DMOJ_EMAIL_CHANGE_LIMIT_WINDOW = 3600
```
Xyene, line 115: Probably makes sense for these fields to all be in minutes.

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+                base64.urlsafe_b64decode(activation_key.encode()).decode(),
+                max_age=settings.DMOJ_EMAIL_CHANGE_EXPIRY_MINUTES * 60,
+            )
+        except (binascii.Error, signing.BadSignature):
+            raise self.EmailChangeFailedError(_('Invalid activation key. Please try again.'))
+        except signing.SignatureExpired:
+            raise self.EmailChangeFailedError(_('This request has expired. Please try again.'))
+        if data['id'] != request.user.id:
+            raise self.EmailChangeFailedError(
+                _('Please try again from the account this email change was originally requested from.'),
```
Xyene, line 595:
```suggestion
                _('Please try again while logged in to the account this email change was originally requested from.'),
```

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+            )
+        with revisions.create_revision(atomic=True):
+            if User.objects.filter(email=data['email']).exists():
+                raise self.EmailChangeFailedError(
+                    _('The email originally requested is in use. Please try again with a new email.'),
+                )
+            request.user.email = data['email']
+            request.user.save()
+            revisions.set_user(request.user)
+            revisions.set_comment(_('Changed email address'))
```
Xyene, line 605: We could save people a click by embedding the to/from emails here.

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+        except signing.SignatureExpired:
+            raise self.EmailChangeFailedError(_('This request has expired. Please try again.'))
+        if data['id'] != request.user.id:
+            raise self.EmailChangeFailedError(
+                _('Please try again from the account this email change was originally requested from.'),
+            )
+        with revisions.create_revision(atomic=True):
+            if User.objects.filter(email=data['email']).exists():
+                raise self.EmailChangeFailedError(
+                    _('The email originally requested is in use. Please try again with a new email.'),
```
Xyene, line 600:
```suggestion
                    _('The email you originally requested has since been registered by another user. Please try again with a new email.'),
```

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+                )
+            request.user.email = data['email']
+            request.user.save()
+            revisions.set_user(request.user)
+            revisions.set_comment(_('Changed email address'))
+
+    def get(self, request, *args, **kwargs):
+        try:
+            self.update_user_email(request, kwargs['activation_key'])
+        except self.EmailChangeFailedError as e:
```
Xyene, line 610: Is the `self.` here actually necessary?

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+            return generic_message(request, _('Email change failed'), str(e), status=403)
+        else:
+            return generic_message(
+                request,
+                _('Email change succeeded'),
```
Xyene, line 615:
```suggestion
                _('Email successfully changed'),
```

judge/views/user.py
```diff
@@ -507,3 +512,106 @@ def post(self, request, *args, **kwargs):
...
+                _('The email attached to your account has been changed.'),
```
Xyene, line 616: If it is easy (and I think we have the data here?) we should say to what in this message.

templates/registration/email_change_activate_email.html
```diff
@@ -0,0 +1,24 @@
...
+</div></div>
+
+<div style="border:2px solid #337ab7;margin:4px 0;"><div style="background:#fafafa;border:12px solid #fafafa;font-family:segoe ui,lucida grande,Arial,sans-serif;font-size:14px;">
+<br><br>
+{{ user.get_username() }},
+<br>
+{% trans %}You have requested to change your email address to this email for your user account at {{ site_name }}.{% endtrans %}
+<br><br>
+{% trans trimmed count=expiry_minutes %}
+Please click the link to confirm this email change. The link will expire in {{ count }} minute.
```
Xyene, line 12: Let's just assert that the expiry is > 1 second as part of config validation, and avoid needing to pluralize this. 1 minute expiry is unusable UX anyway.

templates/registration/email_change_activate_email.html
```diff
@@ -0,0 +1,24 @@
+<div style="border:2px solid #fd0;margin:4px 0;"><div style="background:#000;border:6px solid #000;">
+<a href="{{ protocol }}://{{ domain }}"><img src="{{ protocol }}://{{ domain }}/static/icons/logo.svg" alt="{{ site_name }}" width="160" height="44"></a>
+</div></div>
+
+<div style="border:2px solid #337ab7;margin:4px 0;"><div style="background:#fafafa;border:12px solid #fafafa;font-family:segoe ui,lucida grande,Arial,sans-serif;font-size:14px;">
+<br><br>
+{{ user.get_username() }},
+<br>
+{% trans %}You have requested to change your email address to this email for your user account at {{ site_name }}.{% endtrans %}
```
Xyene, line 9:
```suggestion
{% trans %}Someone (hopefully you!) has requested to change the email address for their user account at {{ site_name }} to this email.{% endtrans %}
```

templates/registration/email_change_notify_email.html
```diff
@@ -0,0 +1,24 @@
+<div style="border:2px solid #fd0;margin:4px 0;"><div style="background:#000;border:6px solid #000;">
+<a href="{{ protocol }}://{{ domain }}"><img src="{{ protocol }}://{{ domain }}/static/icons/logo.svg" alt="{{ site_name }}" width="160" height="44"></a>
+</div></div>
+
+<div style="border:2px solid #337ab7;margin:4px 0;"><div style="background:#fafafa;border:12px solid #fafafa;font-family:segoe ui,lucida grande,Arial,sans-serif;font-size:14px;">
+<br><br>
+{{ user.get_username() }},
+<br>
+{% trans %}You have requested to change your email address to {{ new_email }} for your user account at {{ site_name }}.{% endtrans %}
```
Xyene, line 9:
```suggestion
{% trans %}Someone (hopefully you!) has requested to change the email address associated with your user account at {{ site_name }} to {{ new_email }}. {% endtrans %}
```

templates/registration/email_change_notify_email.html
```diff
@@ -0,0 +1,24 @@
...
+{% if site_admin_email %}
+{% with link='<a href="mailto:%(email)s">%(email)s</a>'|safe|format(email=site_admin_email) %}
+{{ _('If this was not you, please change your password and email us immediately at %(email)s.', email=link) }}
+{% endwith %}
+{% else %}
+{{ _('If this was not you, please change your password and reply to this email immediately.') }}
+{% endif %}
+</b>
+</div></div>
+
```
Xyene, line 24:
```suggestion
```

templates/registration/email_change_notify_email.txt
```diff
@@ -0,0 +1,9 @@
+{{ user.get_username() }},
+{% trans %}You have requested to change your email address to {{ new_email }} for your user account at {{ site_name }}.{% endtrans %}
```
Xyene, line 2:
```suggestion
{% trans %}Someone (hopefully you!) has requested to change the email address associated with your user account at {{ site_name }} to {{ new_email }}. {% endtrans %}
```

judge/views/user.py
Ninjaclasher: Yes, otherwise we get a `NameError` (flake8 also complains)

judge/utils/mail.py
Ninjaclasher: This was originally copied from https://github.com/django/django/blob/main/django/contrib/auth/forms.py#L287. I think their parameters were ordered that way since the HTML template was optional. Since we want likely want our emails to always contain an HTML version, I will reorder the parameters.

templates/registration/email_change_activate_email.html
Ninjaclasher: I've added an assertion for > 1 minute expiry, but going to wait on @quantum5's reply before changing the template to verify this will work for other languages.

Ninjaclasher: Seems like this is necessary for other languages, according to @int-y1.

Xyene (approved):
LGTM, could you please rebase and squash?

@quantum5, could you also take a quick look over this before I merge it?

templates/registration/email_change.html
```diff
@@ -0,0 +1,23 @@
...
+
+{% block media %}
+<style>
+.errorlist {
+    list-style-type: none;
+    margin-block-start: 0;
+    margin-block-end: 0.5em;
+    padding: 0;
+}
```
quantum5, line 10: Maybe move this into a CSS file eventually, but who cares right now.

templates/registration/email_change_activate_email.html
quantum5: Yes, lack of ngettext is going to be so much fun in a bunch of languages...

quantum5 (approved):
LGTM, get this merged soon so we don't have to deal with all those manual requests.

PR #2293, opened 2023-11-18
Migrate all remaining Pagedown widgets to Martor

Ninjaclasher:
Migrate all remaining Pagedown widgets to Martor. The first commit migrates all widgets to Martor. The second commit strips Pagedown from the codebase.

This PR is a prerequisite for username changes, so that we do not need to implement converting between username <-> id for both Pagedown and Martor.

All functionality and UI has been pretty thoroughly tested, including on mobile devices. Note that any UI that previously didn't display well on mobile still won't display well after these changes (i.e. comment editting popup / tickets), as these require more substantial UI changes that are out of the scope of this PR.

Screenshots:
image
image

image
image
image
image
image
image


Xyene:
We should make the editor text area flush with the edges of its component, as for problem editing:
image

Xyene:
One issue I remember having with the admin widget is a bunch of styles colliding. Can you test that if you take the markdown of a complicated problem text and preview it as a comment, it looks the same as the problem text itself? That should sus out any unexpected issues with things like list padding, etc.

Ninjaclasher:
Tested the markdown preview for `problemtemplate`, and the generated preview looks exactly the same as the problem except the parts which are disallowed (i.e. raw HTML, real latex).

Removed the padding on the editor to make it flush with the edges of its component:
image

Xyene:
Thanks for testing. Code-wise this looks fine to me. Can we also go ahead and disable image uploads?

Ninjaclasher:
I've disabled image uploads for all non-admin widgets. Relevant PR in the martor repo: DMOJ/martor#9.

Xyene (approved):
LGTM, thanks!

PR #2296, opened 2023-12-05
Minor improvements to API v2

Ninjaclasher:
- Sort all list endpoints by ID, which reduces the likelihood of objects moving between pages during requests. Objects will only shift if an object is deleted, which usually does not happen.
- Add a list filter for ID (and other identifier fields), so users can request a list of only the objects they care about (rather than requesting all and then filtering on their side)
- Use the rating from `Profile.rating` instead of re-computing it each time. This attribute should already be updated during each contest rate.

### DIFF
--- a/judge/views/api/api_v2.py
+++ b/judge/views/api/api_v2.py
@@ -199,6 +199,7 @@ class APIContestList(APIListView):
         ('is_rated', 'is_rated'),
     )
     list_filters = (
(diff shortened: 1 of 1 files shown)


Ninjaclasher:
Docs PR: DMOJ/docs#154

Xyene (approved):
LGTM, thanks!

PR #2297, opened 2023-12-21
Add Node.js to list of languages on README and demo languages

pxpeterxu:
### Motivation

Related to https://github.com/DMOJ/judge-server/pull/1143:

After enabling support in judge-server for a Node.js-based executor, we need add the language to the demo instance and the documentation

### Changes

- Add JavaScript (Node.js) to the demo instance's languages list (`language_all.json`)
- Add Node.js to the list of supported languages in the README

### Testing

Ran `python3 manage.py loaddata language_all` and verified the languages were loaded properly in the Admin interface

image


README.md
```diff
@@ -121,6 +121,7 @@ The judge can also grade in the languages listed below:
 * GAS x86/x64/ARM
 * Haskell
 * INTERCAL
+* JavaScript (Node.js, V8)
```
Xyene, line 124: I suppose for consistency with the C++ variants this should say "(Node.js and V8)".

pxpeterxu: Changed to say `Node.js and V8`!

judge/fixtures/language_all.json
```diff
@@ -942,5 +942,21 @@
...
         "extension": "cpp"
     }
+},
+{
+    "model": "judge.language",
+    "pk": 77,
+    "fields": {
+        "key": "NODEJS",
+        "name": "JavaScript (Node.js)",
+        "short_name": "Node.js",
```
Xyene, line 952: The short name is used on the submission listing. For consistency, it should be all uppercase. If you leave it as `null`, then `key` will be used instead (which seems fine in this case).

judge/fixtures/language_all.json
```diff
@@ -942,5 +942,21 @@
         "description": "",
         "extension": "cpp"
     }
+},
+{
+    "model": "judge.language",
+    "pk": 77,
+    "fields": {
+        "key": "NODEJS",
+        "name": "JavaScript (Node.js)",
```
Xyene, line 951: This naming sounds good to me -- could we also rename V8 to "JavaScript (V8)" for consistency?

judge/fixtures/language_all.json
pxpeterxu: Good call: I changed this to `null`, and the `key` should work well here

judge/fixtures/language_all.json
pxpeterxu: Changed to JavaScript V8!

pxpeterxu:
Thanks for the review @Xyene and sorry about the extreme lateness of this follow-up. I just made the requested changes, so this should probably be good to go.

I'm not sure who's the best to merge this in, but @kiritofeng and @quantum5, might either of you be open to taking a quick look?

PR #2298, opened 2023-12-22
Test Cases Improvements

FherStk:
Dear DMOJ team,

I developed the following (in my opinion) improvements:
## Test cases (edit form):
- When entering on 'edit test cases', if no data has been entered through the site but the test-cases have been created manually into the file system, the file-system data will be loaded automatically to maintain the website/file-system consistency. This loading process is performed by parsing the YML file, useful when the problems are created using a script (for example, when migrating from other platforms into DMOJ).
- When selecting 'clean the zip file', all the test cases will be also removed to keep the consistency between the displayed data, the database and the YML file.
- A new dropdown option called 'Overwrite test cases from zip file' has been added (disabled by default). If enabled, the test-cases will be automatically loaded and created using all the zip file entries (using default data, zero points also). Useful to quickly load all the test-cases from a zip: 
image

## Problem (edit form and statement):
- A new option called 'Include test cases' has been added to the 'edit problem' form (disabled by default, no pre-tests will be shown):  
image  
When enabled, the current test-cases will be displayed after the problem statement (the content of the in/out files within the zip file will be used). This data is being stored into the database, in the problem_data model: 
image
Including the test-cases input and output manually in the problem statement is still possible, just non-enabling this new option, but this option is useful to avoid inconsistencies between the stored test files and the displayed data. 
- A new option has been included for attribution purposes, in order to give the authorship to external DMOJ users (like websites, books, etc):
image
This data will be displayed at the end of the problem statement, just before the license:
image

## General:
- Added new django settings to disable social and comments sections (false by default):
image

Social and comments enabled (by default):
image

Social and comments disabled:
image

PR #2304, opened 2023-12-27
New settings for disable comments and social

FherStk:
Related with: https://github.com/DMOJ/online-judge/pull/2298

Added new settings to disable social and comments sections (false by default):
image

Blog post example with social and comments enabled (by default):
image

Blog post example with social and comments disabled:
image


dmoj/settings.py
```diff
@@ -161,6 +161,9 @@
 FONTAWESOME_CSS = '//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css'
 DMOJ_CANONICAL = ''
 
+DMOJ_DISABLE_COMMENTS = False
```
Xyene, line 164: I think it makes the code a little clearer (and removes the double-negatives) if we had these be `DMOJ_ENABLE_{COMMENTS,SOCIAL} = True`.

templates/problem/editorial.html
```diff
@@ -37,9 +37,12 @@
             {% endif %}
         {% endwith %}
         {{ solution.content|markdown('solution', MATH_ENGINE)|reference|str|safe }}
-    </div>
-    <hr>
-    {% include "comments/list.html" %}
+    </div> 
+       
+    {% if not disable_comments %}
```
Xyene, line 42: The comments used to be inside the `<div/>`, but are now outside. Was this change intentional?

dmoj/settings.py
FherStk: Done :)

templates/problem/editorial.html
FherStk: If I'm not wrong, is outside the "div" in the current main branch from where I branched:
https://github.com/DMOJ/online-judge/blob/master/templates/problem/editorial.html

dmoj/settings.py
```diff
@@ -161,6 +161,9 @@
 FONTAWESOME_CSS = '//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css'
 DMOJ_CANONICAL = ''
 
+DMOJ_ENABLE_COMMENTS = True
```
Xyene, line 164: Small nit, let's also move these up to below the rest of the `DMOJ_` constants/

Xyene (commented):
Otherwise LGTM -- could you please rebase and squash your commits? Then we can go ahead and merge this.

dmoj/settings.py
FherStk: Done :)

FherStk:
> Otherwise LGTM -- could you please rebase and squash your commits? Then we can go ahead and merge this.

I'm not used to squash and rebase, but I hope I squashed it correctly :)

FherStk:
> Not quite :)
> 
> You want to run `git rebase -i HEAD~10`, which will bring up interactive rebase for the last 10 commits, and change the last few lines to
> 
> ```
> ...
> pick 47863906 new settings options for disabling comments and social
> s eec0b6a9 new settings options for disabling comments and social
> s a528dd54 disable-{social | comments} --> enable-{social | comments}
> s 9f2f5fa5 Fixes for 2304#d
> ```
> 
> `s` short for 'squash', see the big comment in interactive rebase for details. Then you'll be prompted to fix up the commit text for the commit things are being squashed into (`47863906`), after which you'll be able to `--force` push the result onto your branch.

Hi,

Sorry for the delay, I've been a little overwhelmed with my work, but I hope I squashed correctly now. I'm very grateful for all your help and support. 

Thanks again!
Fher

Xyene (approved):
Thanks for the contribution!

resources/libs
int-y1, line 1: why was this submodule updated?

FherStk: I guess this is related with my tries of smashing commits, because the links displays no changes:
imagen
imagen

Issue #2031, opened 2022-11-01
Display pretests


Xyene:
Split out from https://github.com/DMOJ/judge-server/issues/1068:

* [ ] Write site patch to display pretest cases as pretests even when not in a pretested contest.

We probably want to add an indicator in the judge-site protocol for whether a case is a pretest or not. We can _infer_ today from `total = 0`, but we want to move towards a world where pretests can have nonzero points but not count towards a submission's total. This requires separate tagging.

PR #2305, opened 2023-12-27
Improvements for test cases

FherStk:
Related with: https://github.com/DMOJ/online-judge/pull/2298

When entering on 'edit test cases', if no data has been entered through the site but the test-cases have been created manually into the file system, a warning will be displayed:
image

- Choosing the 'infer' option, the zip file will be used to create the test-cases (one entry per file, zero points per entry, default values). 
- Choosing the 'rebuilld' option, the yml file will be used to create the test-cases, but only works with simple problems. 

Also, a new option has been added to the form:
image
- If the zip file is requested to be deleted, all the test-cases will be also removed from the database.
- If a zip file is uploaded, existing test-cases will be reused if possible (points, pretest and other configurations will be preserved), outdated cases will be removed, and the new ones will be added. 

Finally, problems can display their test-cases after the problem statement if needed:
image

All the test-cases (pretests will be ignored, so 'secret' test cases can be used as usual) will append after the problem statement automatically, avoiding inconsistencies between the files and the displayed data:
image

All these changes are an addition, completely optional, and does not alter the default behavior of the base DMOJ. 

I hope you like it, have nice holidays!
Fher


FherStk:
Sorry, but I've found a little bug:

When a test is marked as "pretest" it won't be displayed at the problem statement (if the option is selected within the problem settings), but "pretest" are executed prior to regular tests, so the input / output numeration between the statement and the judge results won't match. 

I think that the best option to fix this is to implement this issue: https://github.com/DMOJ/online-judge/issues/2031

If you agree, I will create a new PR to implement https://github.com/DMOJ/online-judge/issues/2031 and, once integrated, this one will be ready to be integrated too.

Thanks and kind regards,
Fher

FherStk:
> Sorry, but I've found a little bug:
> 
> When a test is marked as "pretest" it won't be displayed at the problem statement (if the option is selected within the problem settings), but "pretest" are executed prior to regular tests, so the input / output numeration between the statement and the judge results won't match.
> 
> I think that the best option to fix this is to implement this issue: #2031
> 
> If you agree, I will create a new PR to implement #2031 and, once integrated, this one will be ready to be integrated too.
> 
> Thanks and kind regards, Fher

Well... I see that the #2031 issue is something bigger than I imagined, and I don't feel ready to address it right now. Maybe I should reimagine how the pretest behavior should be treated because:
1. Pretests are executed before the regular tests.
2. Pretests exists in order to execute them exclusively during a contest, so the regular tests won't be executed till the end of the contest. Does this mean that regular test's inputs and outputs should be hidden during a contest? 
3. If the answer to the prior question is "yes", this means that only pretest's inputs and outputs should be displayed within the problem statement, but it has no sense to mark all tests as pretests in order to display them all.
4. So, arrived to this point, maybe should be better to add a new column to mark a test (or pretest) as hidden, but this will not solve the reordering problem when a pretest is selected... I wouldn't like the idea of disabling pretest selection when a problem enabled the "include test cases" option...

Sorry, I have so many doubts  ^_^U

FherStk:
Hi there! 

Sorry, I've been a bit out but I'm now ready to finish this implementation. I decided to proceed adding a new column to tests in order to mark them as private, so those private tests's inputs and outputs wont be displayed at the problem statement. Also, pretests will be displayed prior to regular tests in order to avoid correlation issues between the judge results and the statement order.

I also made changes in my devel environment so it will be easier for me to squash the commits and send to you the PR with no conflicts.

Please, feel free to contact me if you need further information. Cheers!
Fher

FherStk:
Hi again,

PR ready to review and merge (if you like). 
I also updated the [first post](https://github.com/DMOJ/online-judge/pull/2305#issue-2057607022), so all the changes are described there.

Cheers!
Fher

PR #2308, opened 2024-01-02
Fix user table CSS

int-y1:
Fix CSS highlighting bug in the user table. Compare https://dmoj.ca/users/#!BalintR with https://dmoj.ca/users/#!zhouzixiang2004

Decrease the CSS specificity of the row stripes. This allows the yellow highlight to take priority.


kiritofeng (approved):
LGTM

Xyene (approved):
Thanks!

PR #2309, opened 2024-01-02
Sort contest users with 0 points by submission count

### LINKED ISSUE #650: Contest leaderboard does not sort users with 0 points corectly
since pretty much all rated contests are only rated if you submit at least once, it's probably more helpful to contestants to display the leaderboard such that users who submit and have a score of 0 are ranked above those who don't submit at all.

(i don't think this change affects rating calculations)

int-y1:
Fixes #650. When 2 users score 0 points, the user who submitted will appear first.

Screenshot 2024-01-02 003918

Also, using `'submission_cnt'` will sort the users the other way.

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -677,7 +677,8 @@ def base_contest_ranking_list(contest, problems, queryset):
 def contest_ranking_list(contest, problems):
     return base_contest_ranking_list(contest, problems, contest.users.filter(virtual=0)
                                      .prefetch_related('user__organizations')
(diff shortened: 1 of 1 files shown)


kiritofeng (approved):
LGTM, thanks!

quantum5 (commented):
Before we merge this, can we check to see how much worse the query is?

kiritofeng:
Tested against `dmopc23c1`, running in the prod database the difference is < 0.1s with 10 and 50 iterations as measured by `timeit`

PR #2310, opened 2024-01-09
views/data: fix get_valid_files

Riolku:
Before #2197, `get_valid_files` returned a `List[str]`. Then it returned an `Option[List[str]]`, because we wanted to flag invalid zips. This caused issues with iterating, and is a pretty terrible design.

Instead, we just re-raise the BadZipfile, and handle it in the two places this function is used.

### DIFF
--- a/judge/views/problem_data.py
+++ b/judge/views/problem_data.py
@@ -2,6 +2,7 @@ import json
 import mimetypes
 import os
 from itertools import chain
+from typing import List
 from zipfile import BadZipfile, ZipFile
 
 from django.conf import settings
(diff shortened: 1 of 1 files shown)


kiritofeng (approved):
LGTM, we can add mypy afterwards

judge/views/problem_data.py
```diff
@@ -178,26 +179,35 @@ def get_valid_files(self, data, post=False):
             elif data.zipfile:
                 return ZipFile(data.zipfile.path).namelist()
         except BadZipfile:
-            return None
+            raise
```
quantum5, line 182: Why are you catching this one exception only and reraising it? That's equivalent to not catching it since you don't have any subsequent `except` clauses.

Riolku: I know, my point was to be explicit about letting the exception escape the function.

PR #2312, opened 2024-01-20
Add hCaptcha backend as an alternative to reCAPTCHA

Jazzzny:
Screenshot 2024-01-20 at 4 57 08 PM

This PR adds support for an hCaptcha backend powered by `django-hcaptcha-field`. Most additions are duplicated from the existing reCAPTCHA backend as the hCaptcha package is very similar.

See https://blog.cloudflare.com/moving-from-recaptcha-to-hcaptcha for possible benefits over reCAPTCHA.

### DIFF
--- a/judge/views/register.py
+++ b/judge/views/register.py
@@ -14,6 +14,7 @@ from sortedm2m.forms import SortedMultipleChoiceField
 from judge.models import Language, Organization, Profile, TIMEZONE
 from judge.utils.mail import validate_email_domain
 from judge.utils.recaptcha import ReCaptchaField, ReCaptchaWidget
+from judge.utils.hcaptcha import hCaptchaField, hCaptchaWidget
 from judge.utils.subscription import Subscription, newsletter_id
 from judge.widgets import Select2MultipleWidget, Select2Widget
 
@@ -34,7 +35,10 @@ class CustomRegistrationForm(RegistrationForm):
         newsletter = forms.BooleanField(label=_('Subscribe to newsletter?'), initial=True, required=False)
 
     if ReCaptchaField is not None:
-        captcha = ReCaptchaField(widget=ReCaptchaWidget())
+        recaptcha = ReCaptchaField(widget=ReCaptchaWidget())
+    
+    if hCaptchaField is not None:
+        hcaptcha = hCaptchaField()
 
     def clean_email(self):
         if User.objects.filter(email=self.cleaned_data['email']).exists():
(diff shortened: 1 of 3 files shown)


judge/views/register.py
```diff
@@ -34,7 +35,10 @@ class CustomRegistrationForm(RegistrationForm):
         newsletter = forms.BooleanField(label=_('Subscribe to newsletter?'), initial=True, required=False)
 
     if ReCaptchaField is not None:
-        captcha = ReCaptchaField(widget=ReCaptchaWidget())
```
Xyene, line 37: Rather than having this and then needing to have duplicated blocks in the template, could we keep calling this `captcha`, and only do the hcaptcha thing in an else branch?

Xyene (commented):
Thanks for the contribution. The idea of using hCaptcha as an alternate backend seems good to me, especially given how easy this looks to support.

judge/views/register.py
Jazzzny: Sorry for letting this go mouldy, I'll get this done soon.

Jazzzny:
Ready for review. Only one backend can ever be active at a given time, however that seems logical.

judge/views/register.py
```diff
@@ -33,8 +33,12 @@ class CustomRegistrationForm(RegistrationForm):
     if newsletter_id is not None:
         newsletter = forms.BooleanField(label=_('Subscribe to newsletter?'), initial=True, required=False)
 
-    if ReCaptchaField is not None:
-        captcha = ReCaptchaField(widget=ReCaptchaWidget())
+    if CaptchaField is not None:
+        if ReCaptchaWidget is not None:
+            captcha = CaptchaField(widget=ReCaptchaWidget())
```
quantum5, line 38: Do we even need this special construction? https://github.com/DMOJ/django-recaptcha2/blob/master/snowpenguin/django/recaptcha2/fields.py#L22 seems to imply that `widget=ReCaptchaWidget()` is pointless.

Honestly, at this rate, I am tempted to drop ReCAPTCHA soon, given all the problems I ran into with them and also the fact I had to fork the library...

judge/utils/captcha.py
```diff
@@ -0,0 +1,20 @@
+try:
+    from snowpenguin.django.recaptcha2.fields import ReCaptchaField
+    from snowpenguin.django.recaptcha2.widgets import ReCaptchaWidget
+except ImportError:
+    ReCaptchaField = None
+    ReCaptchaWidget = None
+try:
```
quantum5, line 7: nit:
```suggestion

try:
```

PR #2316, opened 2024-01-24
Add Kazakh language option in UI

int-y1:
* Add `kk` to language selector dropdown.
* Fix a bug with `{{ _('**a** b')|markdown('default', strip_paragraphs=True) }}`.
  * The bug is that `p.text` could be `None`, and Python can't add a string to `None`.


quantum5 (approved):
Looks good, why the (seemingly) unrelated change though?

int-y1:
the translation for `'proudly powered by **DMOJ**'` runs into the bug i mentioned

PR #2327, opened 2024-04-02
Show Problem Types In Contest When Enabled

### LINKED ISSUE #1872: Show Types in Contests with "Show Types"
Related: #1859. The types column is forcibly muted in contest. Instead, we should check if types are enabled in the contest, and display the column accordingly.

tandr3w:
Fixes #1872 (or attempts to)

### DIFF
--- a/judge/views/problem.py
+++ b/judge/views/problem.py
@@ -492,8 +492,6 @@ class ProblemList(QueryStringSortMixin, TitleMixin, SolvedProblemMixin, ListView
     def get_context_data(self, **kwargs):
         context = super(ProblemList, self).get_context_data(**kwargs)
         context['hide_solved'] = 0 if self.in_contest else int(self.hide_solved)
-        if self.in_contest:
-            self.show_types = int(not self.contest.hide_problem_tags)
         context['show_types'] = int(self.show_types)
         context['has_public_editorial'] = 0 if self.in_contest else int(self.has_public_editorial)
         context['full_text'] = 0 if self.in_contest else int(self.full_text)
@@ -549,7 +547,11 @@ class ProblemList(QueryStringSortMixin, TitleMixin, SolvedProblemMixin, ListView
 
     def setup_problem_list(self, request):
         self.hide_solved = self.GET_with_session(request, 'hide_solved')
-        self.show_types = self.GET_with_session(request, 'show_types')
+        if self.in_contest:
+            self.show_types = int(not self.contest.hide_problem_tags)
+        else:
+            self.show_types = self.GET_with_session(request, 'show_types')
+
         self.full_text = self.GET_with_session(request, 'full_text')
         self.has_public_editorial = self.GET_with_session(request, 'has_public_editorial')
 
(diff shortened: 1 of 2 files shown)


judge/views/problem.py
```diff
@@ -490,7 +492,9 @@ def get_queryset(self):
     def get_context_data(self, **kwargs):
         context = super(ProblemList, self).get_context_data(**kwargs)
         context['hide_solved'] = 0 if self.in_contest else int(self.hide_solved)
-        context['show_types'] = 0 if self.in_contest else int(self.show_types)
```
int-y1, line 493: why did you avoid
```py
context['show_types'] = int(not self.contest.hide_problem_tags) if self.in_contest else int(self.show_types)
```

judge/views/problem.py
```diff
@@ -425,8 +426,9 @@ def get_contest_queryset(self):
             'points': p['points'],
             'partial': p['partial'],
             'user_count': p['user_count'],
-        } for p in queryset.values('problem_id', 'problem__code', 'problem__name', 'i18n_name',
-                                   'problem__group__full_name', 'points', 'partial', 'user_count')]
+            'types_list': q.problem.types_list,
+        } for p, q in zip(queryset.values('problem_id', 'problem__code', 'problem__name', 'i18n_name',
+                                          'problem__group__full_name', 'points', 'partial', 'user_count'), queryset)]
```
int-y1, line 431: `zip(queryset.values(...), queryset)` is pretty weird, is there a better way to do this?

int-y1: `zip(queryset.values(...), queryset)` defeats the point of having a `queryset.values(...)`

int-y1:
can you post screenshots of the problem list?

judge/views/problem.py
tandr3w: > why did you avoid
> 
> ```python
> context['show_types'] = int(not self.contest.hide_problem_tags) if self.in_contest else int(self.show_types)
> ```

I wasn't sure whether or not the self.show_types attribute was used elsewhere and had to be changed as well

judge/views/problem.py
tandr3w: > `zip(queryset.values(...), queryset)` is pretty weird, is there a better way to do this?

Couldn't think of a solution using queryset.values(...) since I believe it can't access the types_list property. Maybe getting rid of the queryset.values(...) would be better?

tandr3w:
Before:
image

After: 
image

judge/views/problem.py
int-y1: `get_context_data` doesn't look like the right place to modify `self.show_types`. what happens if you put your changes in `setup_problem_list`?

judge/views/problem.py
quantum5: The `zip` is fundamentally broken.

quantum5 (changes requested):
PR is completely borked, not to mention inefficient.

tandr3w:
- Moved show_types change into the setup_problem_list function
- Replaced zip with an annotation
- Fixed inconsistent padding due to the in-contest types table header not having the "types" class

tandr3w:
New Screenshots:

Before changes:
image

With types disabled, after changes
image

With types enabled, after changes
image

tandr3w:
Now works for translated types, ready for another review
image

tandr3w:
Ready for re-review
image

PR #2330, opened 2024-04-12
Fix comment revision count

### LINKED ISSUE #2295: Fix off-by-1 error in comment revisions
I think https://github.com/DMOJ/online-judge/blob/292ba1a600091dd30c41f34fead38deadaf76536/templates/comments/list.html#L21-L22 has an off-by-1 error which causes comment history to be displayed incorrectly.

As a corollary, I think the code in https://github.com/DMOJ/online-judge/blob/292ba1a600091dd30c41f34fead38deadaf76536/templates/comments/list.html#L70 may also have an off-by-1 error but I am less sure of this.

## TODO
- [ ] investigate off-by-1 when populating `data-*-revision` fields
- [ ] investigate off-by-1 when printing number of edits

int-y1:
Fixes #2295. This fixes all comments created since #2216.

When a comment is created, the comment is on the first revision. So the default value of `revisions` should be 1.

Migration 0146 is a copy of 0142, except for the `AlterField` operation.

## Screenshots

Stay on the old code. Make a comment with 3 edits: Screenshot 2024-04-11 205130

Go to the new code and run migration. Make a comment with 5 edits: Screenshot 2024-04-11 205412

### DIFF
--- a/judge/migrations/0146_comment_revision_count_v2.py
+++ b/judge/migrations/0146_comment_revision_count_v2.py
@@ -1,0 +1,39 @@
+from django.core.exceptions import ObjectDoesNotExist
+from django.db import migrations, models
+
+
+def populate_revisions(apps, schema_editor):
+    db_alias = schema_editor.connection.alias
+    ContentType = apps.get_model('contenttypes', 'ContentType')
+    try:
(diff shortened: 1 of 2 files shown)


Xyene (approved):
LGTM, thanks for the fix!

PR #2343, opened 2024-05-28
Add warning when posting code to comments

int-y1:
cleaned-up version of https://github.com/VNOI-Admin/OJ/pull/201

Screenshot 2024-05-27 231527

### DIFF
--- a/templates/comments/media-js.html
+++ b/templates/comments/media-js.html
@@ -217,6 +217,16 @@
                 $comment.find('.comment-body').show();
                 $comment.find('.bad-comment-body').hide();
             };
+
+            var code_regex = [/input\(/, /#include/, /void\s+main/, /fn\s+main/, /func\s+main/];
+            $(document).on('click', 'form.comment-warn-code .button[type=submit]', function (e) {
+                var text = $(this).parents('form').find('#id_body').val();
+                if (code_regex.some(function (regex) {return regex.test(text);})) {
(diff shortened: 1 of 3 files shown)


templates/comments/media-js.html
```diff
@@ -217,6 +217,16 @@
                 $comment.find('.comment-body').show();
                 $comment.find('.bad-comment-body').hide();
             };
+
+            var code_regex = [/input\(/, /#include/, /void\s+main/, /fn\s+main/, /func\s+main/];
+            $(document).on('click', 'form.comment-submit-form .button[type=submit]', function (e) {
+                var text = $(this).parents('form').find('#id_body').val();
+                if (code_regex.some(function (regex) {return regex.test(text);})) {
+                    if (!confirm({{ _('Looks like you\'re trying to post a source code!\n\nThe comment section is not for posting source code.\nIf you want to submit your solution, please use the "Submit solution" button.\n\nAre you sure you want to post this?')|htmltojs }})) {
```
Ninjaclasher, line 225: The English here feels a little off. Could we rewrite it? Specifically "post a source code".

Ninjaclasher:
I think this code is also used for contest / blog post comments. Maybe we want a different message there since redirecting to the "submit solution" button is no longer meaningful.

templates/comments/media-js.html
int-y1: done. the message is changed to:

```
Looks like you're trying to post some source code!

The comment section is not for posting source code.
If you want to submit your solution, please use the "Submit solution" button.

Are you sure you want to post this?
```

int-y1:
list of changes:

* popup message changed. "post a source code" -> "post some source code"
* popup only appears on `/problem/<problem code>`. all other comment sections still work, and aren't affected.

Ninjaclasher (approved):

PR #2346, opened 2024-08-21
Add judge tiering for automatic failover

### LINKED ISSUE #2099: Primary / secondary judges as first-class scheduler concepts
@quantum5 and I both have machines capable of running backup judges for dmoj.ca at ~all times, but they are not calibrated to be quite the same as what we run in production.

If we had a Primary | Secondary choice in the judge model, we could mark them as Secondary and leave them connected to the prod bridge. The scheduler would then be modified to only schedule to secondary judges if there are no online, non-disabled (after #2005) primary judges.

While we haven't had frequent outages of the current judges, we have had a nonzero number and this would increase the reliability of the platform.

quantum5:
This PR adds judge tiering. Only the current lowest tier judge is used for grading while all higher tier judges are held in reserve for high availability.

Fixes #2099.

### DIFF
--- a/judge/admin/runtime.py
+++ b/judge/admin/runtime.py
@@ -65,12 +65,12 @@ class JudgeAdmin(VersionAdmin):
     readonly_fields = ('created', 'online', 'start_time', 'ping', 'load', 'last_ip', 'runtimes', 'problems',
                        'is_disabled')
     fieldsets = (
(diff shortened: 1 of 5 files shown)


Ninjaclasher:
Fixes #2099.

Ninjaclasher (approved):
Looks reasonable enough.

PR #2347, opened 2024-08-29
models/contest: clarify rate_all behaviour

Riolku:
Rate all is confusingly named, since it really means "rate even users who don't submit". In my opinion, this should have been the default, but instead we can simply fix the description.


kiritofeng (approved):
LGTM, thanks!

PR #2348, opened 2024-09-01
Fix ticket text wrapping

kiritofeng:


templates/ticket/message.html
```diff
@@ -20,7 +20,9 @@
             </span>
         </div>
         <div class="content content-description">
-            {{ message.body|markdown('ticket', MATH_ENGINE)|reference|str|safe }}
+            <div class="ticket-body">
```
quantum5, line 23: Why a new div and not just put this as an additional class on the outer div?

kiritofeng: We use nested `div`s for comments (but that has a functional purpose); should still add as an additional class on the outer `div`?

quantum5: I don't see the point of the extra layer of div. Just put it on the parent div.

int-y1: I also don't see the point of this new div. What is the functional purpose of nested divs in comments, and is that needed here?

quantum5 (approved):

PR #2355, opened 2024-10-26
Minor changes to fix compatibility with older browsers

Jazzzny:
This PR provides 2 changes to fix compatibility with older browsers.

1. Older browsers that do not support the `prefers-color-scheme` media query will fail to load any CSS when the dark/light mode code is enabled. 

Additionally, the W3 Specification states that [the light theme should be loaded as a fallback if no preference is given](https://www.w3.org/TR/mediaqueries-5/#prefers-color-scheme:~:text=or%20has%20not%20expressed%20an%20active%20preference%20(and%20thus%20should%20receive%20the%20%22web%20default%22%20of%20a%20light%20theme).).

This can trivially be resolved by removing the `prefers-color-scheme: light` media query from the light theme, to set it as the default if the browser does not handle the query.

https://github.com/user-attachments/assets/01b6cf52-811f-437a-aa9b-a01782173f5d

This does not impact browsers that support the query, as `prefers-color-scheme` takes priority over the "base" stylesheet:
Screenshot 2024-10-26 at 2 04 33 PM

2. Older browsers do not support addEventListener on matchMedia, instead add a fallback using addListener.

### DIFF
--- a/django_ace/static/django_ace/widget.js
+++ b/django_ace/static/django_ace/widget.js
@@ -111,9 +111,15 @@
                 }
 
                 setEditorTheme(window.matchMedia('(prefers-color-scheme: dark)').matches);
-                window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', function(ev) {
-                    setEditorTheme(ev.matches);
-                })
+                try {
+                    window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', function(ev) {
+                        setEditorTheme(ev.matches);
+                    })
+                } catch (err) {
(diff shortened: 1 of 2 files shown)


Jazzzny:
Appended with a fallback for addEventListener. This makes the editor fully functional on every browser I have tested, back to  Safari 5.1.7.

Ninjaclasher (approved):
LGTM, thanks!

Ninjaclasher:
@Jazzzny could you squash your commits into one?

Jazzzny:
Yep, squashed 👍

PR #2360, opened 2024-12-06
admin: Add `create_private_problem` perm

kiritofeng:
Analogue of the contest perm `create_private_contest` for problems.


Ninjaclasher:
1. Add missing migration
2. Needs something similar to https://github.com/DMOJ/online-judge/blob/af8dbfd934fcf9fb348644d9977a710067db9ad9/judge/admin/contest.py#L229-L230 from contests for both `make_public` and `make_private`, otherwise problem setters can bypass this permission by using the batch public action.

quantum5 (approved):

PR #2361, opened 2024-12-06
Fix ContestAdmin make_hidden to actually hide contests

Ninjaclasher:
This seems like a regression from 25213ed68a0395cbabdb4899af99e58241ecbb67, which seems to have accidentally used the wrong value (https://github.com/DMOJ/online-judge/commit/25213ed68a0395cbabdb4899af99e58241ecbb67#diff-3925f94ab82b2f59d01c363a29dc2c95b6465b3084feddc465d30643260b9b4dR195)


kiritofeng (approved):
I can't believe no one noticed this for over 4 years...

PR #2363, opened 2024-12-18
Remove `contest_object` from submissions whose problem is removed from contest; fixes #2349

### LINKED ISSUE #2349: Removing problem from contest does not remove `contest_object` from affected subs
Which in turn, corrupts the DB state.

kiritofeng:


kiritofeng:
Update: added comments explaining what `contest` and `contest_object` are because it's not clear what their purpose is just from their names.

quantum5 (approved):

PR #2377, opened 2024-12-25
Merge Martor repo into main repo

Ninjaclasher:
Decided to do a complete clone of the martor repo, at https://github.com/DMOJ/martor, instead of only picking the parts that are used. We can reduce the footprint later.


quantum5 (approved):
Complete import looks fine. We can definitely trim later.

PR #2378, opened 2024-12-26
Trim in-repo copy of Martor

Ninjaclasher:
Also stop checking if a request is_ajax since this is now deprecated and was supposedly never a reliable indicator whether a request was ajax or not.


martor/settings.py
```diff
@@ -4,7 +4,6 @@
 # Input: string boolean, `true/false`
 MARTOR_ENABLE_CONFIGS = getattr(
     settings, 'MARTOR_ENABLE_CONFIGS', {
-        'emoji': 'true',        # to enable/disable emoji icons.
         'imgur': 'true',        # to enable/disable imgur/custom uploader.
```
kiritofeng, line 7: Should we also need to remove `imgur` and `mention`?

Ninjaclasher: They're actually being used - imgur is actually generic custom image uploader and mention is for allowing search when you have `[user:<cursor>]`.

quantum5: If they are being used, is it safe to delete the extensions?

martor/static/martor/js/martor.js
```diff
@@ -145,9 +127,9 @@
             }
             // Set autocomplete for ace editor
             if (editorConfig.mention === 'true') {
-                editor.completers = [emojiWordCompleter, mentionWordCompleter]
+                editor.completers = [mentionWordCompleter]
             }else {
```
quantum5, line 131: Let's autoformat this js at some point... Disgusting...

quantum5 (commented):
LGTM, one final question.

martor/settings.py
Ninjaclasher: Unless I am mistaken, I believe extensions are only used if we use their markdownify function (https://github.com/DMOJ/online-judge/blob/master/martor/utils.py#L29), which I don't forsee any plans to. Other than that, they are unused.

quantum5 (approved):

martor/static/martor/js/martor.js
int-y1: #2388

PR #2389, opened 2024-12-27
Prettify Ace editor scrollbar

Ninjaclasher:
We copy the scrollbar from the Martor widget. Part of https://github.com/DMOJ/online-judge/issues/2035.

image
image

### DIFF
--- a/resources/ace-dmoj.scss
+++ b/resources/ace-dmoj.scss
@@ -14,6 +14,25 @@
     &.loading {
         display: none;
     }
+
+    /* Scroll bar */
+    ::-webkit-scrollbar-track {
+        -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,0.3);
+        border-radius: 10px;
+        background-color: $color_primary5;
+    }
+
+    ::-webkit-scrollbar {
+        height: 6px;
+        width: 6px;
+        background-color: $color_primary5;
+    }
+
+    ::-webkit-scrollbar-thumb {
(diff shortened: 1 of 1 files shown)


resources/ace-dmoj.scss
```diff
@@ -14,6 +14,25 @@
...
+        height: 6px;
+        width: 6px;
+        background-color: $color_primary5;
+    }
+
+    ::-webkit-scrollbar-thumb {
+        border-radius: 10px;
+        -webkit-box-shadow: inset 0 0 6px rgba(0,0,0,.3);
+        background-color: $color_primary66;
+    }
```
quantum5, line 35: Can we use the new standard way of doing this and let autoprefixer take care of this fallback? Right now this doesn't work on Firefox at all.

quantum5 (changes requested):

resources/ace-dmoj.scss
Ninjaclasher: The firefox scrollbar already looks pretty good, though I'm not sure what it would look like in other non-WebKit  browsers.

It also doesn't seem Firefox supports scrollbar styling (other than the standardized properties that only allow color and width), so I think these properties will need to be webkit-specific. It also doesn't seem autoprefixer supports falling back to these properties yet (https://github.com/postcss/autoprefixer/issues/821).

Firefox scrollbar:
image

quantum5 (approved):
Alright then.

PR #2391, opened 2024-12-28
Migrate martor guide to featherlight

int-y1:
part of #2035 - this makes the martor guide dark mode friendly

* apply dmoj theme to martor guide
* for admin martor, give access to `featherlight.css` and `featherlight.min.js`
* set `max-height: 85%` to avoid issues with the admin site's z-index

## screenshots

Screenshot 2024-12-27 223616
Screenshot 2024-12-27 223647

### DIFF
--- a/martor/templates/martor/guide.html
+++ b/martor/templates/martor/guide.html
@@ -1,10 +1,9 @@
 {% load i18n static %}
-<div class="ui medium modal scrolling transition modal-help-guide">
-  <i class="close icon"></i>
-  <div class="header"><i class="help circle icon"></i> {% trans "Markdown Guide" %}</div>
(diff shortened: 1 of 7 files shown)


martor/templates/martor/guide.html
```diff
@@ -25,10 +24,6 @@
           </tr>
         {% endif %}
 
-        {% if mentions_enabled %}
```
Ninjaclasher, line 28: What did this used to do / why was it deleted?

int-y1: this inserts an empty row into the table, and i think it's necessary

quantum5 (approved):

PR #2392, opened 2024-12-28
Fix dark mode for reply comment

int-y1:
part of #2035

make sure this code isn't run again when replying to a comment

### DIFF
--- a/martor/static/martor/js/martor.js
+++ b/martor/static/martor/js/martor.js
@@ -54,7 +54,7 @@
             var editorConfig = JSON.parse(textareaId.data('enable-configs').replace(/'/g, '"'));
 
             if (darkMode) {
-                mainMartor.find('.ui').toggleClass('inverted');
+                mainMartor.find('.ui').addClass('inverted');
             }
 
             editor.setTheme('ace/theme/' + ace_theme);


martor/static/martor/js/martor.js
```diff
@@ -53,7 +53,7 @@
             var editor       = ace.edit(editorId);
             var editorConfig = JSON.parse(textareaId.data('enable-configs').replace(/'/g, '"'));
 
-            if (darkMode) {
+            if (darkMode && !field_name.match(/^reply-/)) {
```
quantum5, line 56: Not convinced this is sufficiently generic. I think you should check for `inverted` being on a known class or not.

quantum5 (changes requested):

martor/static/martor/js/martor.js
int-y1: `addClass` should work here - https://stackoverflow.com/a/7403519

`inverted` isn't used for light mode, so there is no need for `toggleClass`

quantum5: `inverted` was used in light mode in the emoji picker, but that has since been deleted, so `addClass` is now sufficient.

quantum5 (approved):

PR #2401, opened 2025-01-05
Improve i18n for formatting memory size

int-y1:
this will use the correct decimal separator for certain languages. example:

Screenshot 2025-01-05 151942


kiritofeng (approved):
LGTM, Thanks!

PR #2396, opened 2024-12-30
Dark mode support for alert boxes

int-y1:
part of #2035

* for light mode, shorten `#123456` to `#123` without modifying the color too much
* add support for dark mode

Screenshot 2024-12-29 211122

PR #2402, opened 2025-01-06
Dark mode support for more alert boxes

int-y1:
part of #2035.

i missed some alert boxes from #2396. `messages.success` produced alert boxes that are not dark-mode friendly.

## screenshots

left is old. middle and right are new.

Screenshot 2025-01-05 214752


Ninjaclasher (approved):
LGTM

PR #2403, opened 2025-01-14
Standardize i18n for martor

int-y1:
most templates use `{{ _('foo') }}` for i18n. apply this to martor.

this shouldn't change the editor's behaviour


quantum5:
> most templates use `{{ _('foo') }}` for i18n. apply this to martor.

"Most" templates are Jinja2. I can't tell if these martor templates are being loaded as Jinja2 or plain Django templates.

> this shouldn't change the editor's behaviour

If it's loaded as Django, very bad things will happen.

PR #2404, opened 2025-01-19
New "exception" checker, for expected exceptions.

FherStk:
Added a new "exception" checker, so now it's possible to check for expected exceptions (blank argument means any type of exception). 

This PR is related with the one with the same name within the "judge" repo.


Ninjaclasher:
See the comment on the judge-side PR. However, I do like some of the other changes in this PR! If you have the time, we would gladly accept a PR to generalize `checker_precision` into `checker_args` (albeit there'll only be precision for now).

PR #2406, opened 2025-02-11
Force users to click "Preview" before posting

int-y1:
force users to click on "Preview" before posting a comment or creating a ticket

Screenshot 2025-02-10 231310

this should improve the quality of comments and tickets

places that use a MartorWidget:

| form | affected url's | notes |
| -- | -- | -- |
| `CommentForm` | `/post/1-first-post`, `/contest/<key>`, `/problem/aplusb`, `/problem/aplusb/editorial` | the 2 messages are `Please click on "Preview" before posting your comment.` and `Post!` |
| `ProfileForm` | `/edit/profile/` and a social auth thing? | no change |
| `EditOrganizationForm` | `/organization/1-dmoj/edit` | no change |
| `CommentEditForm` | editing a comment | no change, button is still `Post!` |
| `TicketForm` | `/problem/aplusb/tickets/new` | the 2 messages are `Please click on "Preview" before creating your ticket.` and `Create` |
| `TicketCommentForm` | `/ticket/1` | no change, button is still `Post!` |

### DIFF
--- a/judge/comments.py
+++ b/judge/comments.py
@@ -31,12 +31,15 @@ class CommentForm(ModelForm):
             'parent': forms.HiddenInput(),
         }
 
-        widgets['body'] = MartorWidget(attrs={'data-markdownfy-url': reverse_lazy('comment_preview')})
+        widgets['body'] = MartorWidget(
+            editor_msg=_('Please click on "Preview" before posting your comment.'),
+            button_text=_('Post!'),
+            attrs={'data-markdownfy-url': reverse_lazy('comment_preview')},
+        )
 
     def __init__(self, request, *args, **kwargs):
         self.request = request
         super(CommentForm, self).__init__(*args, **kwargs)
-        self.fields['body'].widget.attrs.update({'placeholder': _('Comment body')})
 
     def clean(self):
         if self.request is not None and self.request.user.is_authenticated:
--- a/martor/templates/martor/editor.html
+++ b/martor/templates/martor/editor.html
@@ -17,6 +17,18 @@
     <div class="ui bottom attached tab segment martor-preview" data-tab="preview-tab-{{ field_name }}">
       <p>{% trans "Nothing to preview" %}</p>
     </div>
+    {% if editor_msg and button_text %}
+      <hr>
+      <div class="ui tab show-on-editor active">
+        <input style="float: right" type="submit" value="{{ button_text }}" class="button disabled" disabled>
+      </div>
+      <div class="ui tab show-on-preview">
+        <input style="float: right" type="submit" value="{{ button_text }}" class="button">
+      </div>
+      <div class="ui tab show-on-editor active">
+        <p>{{ editor_msg }}</p>
+      </div>
+    {% endif %}
   </div><!-- end  /.section-martor -->
 
   {% include 'martor/guide.html' %}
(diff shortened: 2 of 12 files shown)


judge/comments.py
```diff
@@ -31,12 +31,15 @@ class Meta:
...
+        widgets['body'] = MartorWidget(
+            editor_msg=_('Please click on "Preview" before posting your comment.'),
+            button_text=_('Post!'),
+            attrs={'data-markdownfy-url': reverse_lazy('comment_preview')},
+        )
 
     def __init__(self, request, *args, **kwargs):
         self.request = request
         super(CommentForm, self).__init__(*args, **kwargs)
-        self.fields['body'].widget.attrs.update({'placeholder': _('Comment body')})
```
Ninjaclasher, line 39: Just to confirm, is this dead code? I see it's being removed without anything similar being added.

martor/templates/martor/editor.html
```diff
@@ -17,6 +17,22 @@
     <div class="ui bottom attached tab segment martor-preview" data-tab="preview-tab-{{ field_name }}">
       <p>{% trans "Nothing to preview" %}</p>
     </div>
+    {% if editor_msg or button_text %}
```
Ninjaclasher, line 20: I think every time we want this behaviour, we should have some explanation for why the button is disabled. Could we make the `editor_msg` mandatory if we use this behaviour? Similarly, I don't see a clear use case for this message if we don't require preview, so maybe make it so either both `button_text` and `editor_msg` are specified or none of them. (would also make the comment below make more sense)

martor/templates/martor/editor.html
```diff
@@ -17,6 +17,22 @@
...
+    {% if button_text %}
+      <div class="ui tab show-on-editor active">
+        <input style="float: right" type="submit" value="{{ button_text }}" class="button disabled" disabled>
+      </div>
+      <div class="ui tab show-on-preview">
+        <input style="float: right" type="submit" value="{{ button_text }}" class="button">
+      </div>
+    {% endif %}
+    {% if editor_msg %}
+      <div class="ui tab show-on-editor active">
```
Ninjaclasher, line 32: Would it be difficult to make this message a tooltip when hovering the button on desktop/browsers with tooltip hover support? Would need a fallback for mobile. Ignore if this is too annoying to implement.

Ninjaclasher (commented):
Overall looks pretty good, thanks for the cleanup!

I'm just slightly worried about the intuitiveness when someone creates a new `MartorWidget` and wants this behaviour. In particular, it seems to get the behaviour, you need specify `button_text`. I don't have any better solutions off the top of my head, so maybe just add a comment about it for now?

judge/comments.py
int-y1: yes, this became dead when dmoj migrated from pagedown to martor

martor/templates/martor/editor.html
int-y1: tooltips are annoying to implement. it is possible to do `<input ... disabled title="foo">` but this is too hidden to be useful.

martor/templates/martor/editor.html
int-y1: done, i changed it so that you need to specify both or neither. i used `ImproperlyConfigured` for the wrong use.

Ninjaclasher (approved):

PR #2407, opened 2025-02-23
Hide sign up button when registration is disabled

mircearoata:
`REGISTRATION_OPEN = False`
image

`REGISTRATION_OPEN = True`
image


judge/template_context.py
```diff
@@ -61,6 +61,7 @@ def general_info(request):
         'nav_tab': FixedSimpleLazyObject(partial(__nav_tab, request.path)),
         'nav_bar': NavigationBar.objects.all(),
         'LOGIN_RETURN_PATH': '' if path.startswith('/accounts/') else path,
+        'REGISTRATION_OPEN': getattr(settings, 'REGISTRATION_OPEN', True),
```
Ninjaclasher, line 64: Instead of using `getattr`, add this setting into `settings.py` with the default of `True. See #993 for motivation.

kiritofeng (approved):
LGTM, thanks!

PR #2410, opened 2025-02-28
add perf_ceiling in rating computation

### LINKED ISSUE #2409: Add perf_ceiling to ratings
Contests with a rated ceiling are still scaled on the normal performance curve, which has some unintuitive performance for rated contests.

We should cap the maximum performance to be rating_ceiling + EPSILON for such contests.

xiaowuc1:
This fixes #2409.


int-y1:
the code looks reasonable. still need to check if the code runs properly.

kiritofeng:
Superseded by #2417 .

PR #2411, opened 2025-03-02
Add field to quarantine a submission

int-y1:
there are 2 ways to quarantine a submission:

1. go to `/problem/<code>/manage/submission` and click "Quarantine all locked submissions"
2. go to `/admin/judge/submission/<code>/change/` and toggle "Is quarantined"

user-facing changes when a submission is quarantined:

* during point recalculation (e.g. `/admin/judge/profile/ > Action > Recalculate scores`): submission is not considered when recalculating points, problems solved, and pp
* `/user/<name>/solved`: submission no longer appears in "Points breakdown" and in the problems solved table
* `/problem/<code>/rank/`: submission no longer appears as a best solution

admin-facing changes when a submission is quarantined:

* `/problem/<code>/manage/submission`: submission can still be rescored, but can't be rejudged or quarantined again
* `/submissions/` and `/problem/<code>/submissions/`: "rejudge" link changes to "quarantined"
* `/submission/<id>`: "Rejudge" link no longer appears
* `/src/<id>`: "Rejudge" link no longer appears
* `/admin/judge/submission/`: "Rejudge" button changes to "Quarantined"
* `/admin/judge/submission/<id>/change/`: "Rejudge" button in footer toolbar no longer appears


judge/tasks/submission.py
```diff
@@ -18,11 +18,40 @@ def apply_submission_filter(queryset, id_range, languages, results):
...
+        Submission.objects.filter(problem_id=problem_id, is_archived=False, locked_after__lt=timezone.now())
+
+    with Progress(self, submissions.count(), stage=_('Modifying submissions')) as p:
+        archived = 0
+        for submission in submissions.iterator():
+            submission.is_archived = True
+            submission.save(update_fields=['is_archived'])
+            archived += 1
+            if archived % 10 == 0:
+                p.done = archived
```
Riolku, line 39: Does this loop result in N database calls? Can't we do this with one database call?

Riolku:
Overall ty for working on this, it looks great :)

judge/tasks/submission.py
kiritofeng: I don't think this results in DB calls, rather it's just used for the progress bar (someone please check me on that).

judge/admin/submission.py
```diff
@@ -244,7 +246,9 @@ def language_column(self, obj):
 
     @admin.display(description='')
     def judge_column(self, obj):
-        if obj.is_locked:
+        if obj.is_archived:
+            return format_html('<input type="button" disabled value="{0}"/>', _('Archived'))
+        elif obj.is_locked:
```
Riolku, line 251: As discussed in discord briefly, a submission being archived but not locked should mean "this submission should be invalidated for points, but is otherwise valid". So maybe archived submissions should still be rejudgeable?

judge/migrations/0150_submission_is_archived.py
```diff
@@ -0,0 +1,35 @@
...
+                    ('spam_submission', 'Submit without limit'),
+                    ('view_all_submission', 'View all submission'),
+                    ('resubmit_other', "Resubmit others' submission"),
+                    ('lock_submission', 'Change lock status of submission'),
+                    ('archive_submission', 'Archive any submission'),
+                ),
+                'verbose_name': 'submission',
+                'verbose_name_plural': 'submissions',
+            },
+        ),
```
Riolku, line 29: Err, what does this part of the migration do?

judge/models/submission.py
```diff
@@ -121,7 +122,7 @@ def is_locked(self):
         return self.locked_after is not None and self.locked_after < timezone.now()
 
     def judge(self, *args, rejudge=False, force_judge=False, rejudge_user=None, **kwargs):
-        if force_judge or not self.is_locked:
+        if force_judge or not (self.is_archived or self.is_locked):
```
Riolku, line 125: If archived submissions are deemed rejudgeable, this should be changed.

judge/tasks/submission.py
```diff
@@ -18,11 +18,40 @@ def apply_submission_filter(queryset, id_range, languages, results):
         queryset = queryset.filter(language_id__in=languages)
     if results:
         queryset = queryset.filter(result__in=results)
-    queryset = queryset.exclude(locked_after__lt=timezone.now()) \
+    queryset = queryset.exclude(is_archived=True) \
+                       .exclude(locked_after__lt=timezone.now()) \
```
Riolku, line 22: Why is this added? I dont see it immediately, what is this function for?

templates/submission/row.html
```diff
@@ -54,16 +57,22 @@
...
-                        <i class="fa fa-refresh fa-fw"></i><span class="label">{{ _('rejudge') }}</span>
-                    </a>
-                {% else %}
+                {% if submission.is_archived %}
+                    <i class="fa fa-refresh fa-fw grey-icon"></i>
+                    <span class="label grey-label"
+                          title="{{ _('This submission has been archived, and cannot be rejudged.') }}">
+                        {{ _('archived') }}
+                    </span>
+                {% elif submission.is_locked %}
```
Riolku, line 66: This will need to be updated potentially.

templates/submission/status.html
```diff
@@ -74,7 +74,7 @@
     {% if request.user == submission.user.user or perms.judge.resubmit_other %}
         <div><a href="{{ url('problem_submit', submission.problem.code, submission.id) }}">{{ _('Resubmit') }}</a></div>
     {% endif %}
-    {% if perms.judge.rejudge_submission and not submission.is_locked %}
+    {% if perms.judge.rejudge_submission and not (submission.is_archived or submission.is_locked) %}
```
Riolku, line 77: This and source.html might need an update.

judge/migrations/0150_submission_is_archived.py
int-y1: this is needed so that the permission `archive_submission` appears under `/admin/auth/user/<id>/change/ > Available user permissions`

judge/tasks/submission.py
int-y1: this queryset is used by `/problem/aplusb/manage/submission > Rejudge selected submissions`

`.exclude(is_archived=True)` is not needed anymore, and i'll remove it

judge/tasks/submission.py
int-y1: i copy pasted from `rescore_problem`. i guess `rescore_problem` is poorly designed too.

int-y1: this code no longer exists
i went with a different approach (see new toggle in `templates/problem/manage_submission.html`)

judge/views/problem_manage.py
```diff
@@ -82,15 +82,18 @@ def perform_action(self):
         except ValueError:
             return HttpResponseBadRequest()
 
-        return self.generate_response(id_range, languages, self.request.POST.getlist('result'))
+        archive_locked = self.request.user.has_perm('judge.archive_submission') and \
```
Ninjaclasher, line 85: I think having the archiving by default when locked would be better when the user doesn't have this permission. For example, some admin locks some submissions for a problem. Then, a problem setter wants to rejudge all submissions on their problem. Locked submissions should be automatically archived in my opinion.

This use case is probably more common than a problem setter _not_ wanting to archive locked submissions.

judge/models/profile.py
```diff
@@ -243,15 +243,16 @@ def calculate_points(self, table=_pp_table):
         from judge.models import Problem
         public_problems = Problem.get_public_problems()
         data = (
-            public_problems.filter(submission__user=self, submission__points__isnull=False)
+            public_problems.filter(submission__user=self, submission__is_archived=False,
```
Ninjaclasher, line 246: We should also remove archived submissions from problem stats I think.

Ninjaclasher (commented):
It would be nice if, in the places where we currently show `locked` instead of `Rejudge`, we give the option of `Archive`. Off the top of my head, this is currently in `judge/admin/submission.py`'s `judge_column`, the submission row (`templates/submission/row.html`), and submission status/source (`templates/submission/(status|source).html`).

This would match the behaviour in other places (e.g. contest problem rejudge) where there is no check for `is_locked` and thus we go to `Submission.judge`  and automatically archive.

Otherwise LGTM.

int-y1:
* allow access to "Archive locked submissions". don't put this behind the "judge.archive_submission" permission.
* add `is_archived=False` to more queries

judge/models/profile.py
int-y1: done. i also added `is_archived=False` to all of the queries that i found.

judge/models/submission.py
```diff
@@ -129,6 +130,9 @@ def judge(self, *args, rejudge=False, force_judge=False, rejudge_user=None, **kw
                     revisions.set_comment('Rejudged')
                     revisions.add_to_revision(self)
             judge_submission(self, *args, rejudge=rejudge, **kwargs)
+        elif not self.is_archived:
+            self.is_archived = True
+            self.save(update_fields=['is_archived'])
```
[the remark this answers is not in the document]
int-y1: the signal `submission_update` should handle this update

Ninjaclasher (approved):
LGTM, but given the complexity of this PR, will wait for another reviewer to approve before merging.

quantum5 (commented):
No issue with the code, but I don't like the overloading rejudge to also mark as archived. Why not make it a separate option?

kiritofeng:
I remember the reason we wanted "rejudge" to be "rejudge and archive" was because in most cases, you'd want to do both simultaneously.
I think a compromise solution might be to have a button for "rejudge and archive", and possibly separate buttons for "rejudge" and "mark archived"?

Thoughts? @int-y1 @quantum5 @Ninjaclasher

judge/migrations/0150_submission_is_archived.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 4.2.17 on 2025-03-02 10:16
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('judge', '0149_add_organization_private_problems_permission'),
```
kiritofeng, line 9: You'll need to rebase the migration.

int-y1:
changes:
* rebased the migration to 0151
* moved the archiving code from `Submission.judge` to `Submission.archive`
* `rejudge_problem_filter` will determine if the submission is rejudged or archived

i didn't add a button for only "mark archived" because it's not needed yet

kiritofeng (approved):
LGTM.

PR #2416, opened 2025-03-31
Add permission-based navbar items

mircearoata:
image
image
image


judge/template_context.py
```diff
@@ -59,7 +59,7 @@ def general_info(request):
     path = request.get_full_path()
     return {
         'nav_tab': FixedSimpleLazyObject(partial(__nav_tab, request.path)),
-        'nav_bar': NavigationBar.objects.all(),
+        'nav_bar': NavigationBar.for_user(request.user),
```
quantum5, line 62: What about `nav_tab`? This is going to break it.

judge/models/interface.py
```diff
@@ -61,6 +64,11 @@ def pattern(self, cache={}):
             pattern = cache[self.regex] = re.compile(self.regex, re.VERBOSE)
             return pattern
 
+    @staticmethod
+    def for_user(user):
+        all_navbar_items = NavigationBar.objects.all()
+        return list(item for item in all_navbar_items if item.permission is None or user.has_perm(item.permission))
```
quantum5, line 70: Why are you using `list(...)` to emulate list comprehensions when you can use the real thing?

judge/models/interface.py
```diff
@@ -61,6 +64,11 @@ def pattern(self, cache={}):
             pattern = cache[self.regex] = re.compile(self.regex, re.VERBOSE)
             return pattern
 
+    @staticmethod
```
quantum5, line 67: Can we use `classmethod`? This is unpythonic, especially when you refer to the class in the body.

judge/models/interface.py
```diff
@@ -61,6 +64,11 @@ def pattern(self, cache={}):
...
+    def for_user(user):
+        all_navbar_items = NavigationBar.objects.all()
+        return list(item for item in all_navbar_items if item.permission is None or user.has_perm(item.permission))
+
```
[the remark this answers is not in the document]
quantum5: Did you test this code? Is Copilot hallucinating?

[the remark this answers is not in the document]
kiritofeng: Copilot....might actually not be hallucinating....

judge/template_context.py
mircearoata: Right, this would break if multiple navbar items match the current path, but the first result is the one that gets hidden, then nothing would be highlighted.

I can see a few ways to fix this:
* passing the path to `for_user` and moving the regex query there, but if anything else would use the navbar in the future it would run into the same issue again
* making `for_user` take a queryset and the function would only apply the permission filter, but this would then result in `NavigationBar.for_user(NavigationBar.objects.all())` and `NavigationBar.for_user(NavigationBar.objects.extra(where=['%s REGEXP BINARY regex'], params=[path]))`
* adding a custom queryset class for NavigationBar that applies the filter in the query (the one from https://github.com/DMOJ/online-judge/pull/2416#discussion_r2022801531)

quantum5: Given that we are already fetching the entire navbar, computing `nav_tab` on the db layer is unnecessary. You can probably just change `nav_tab` to be computed in Python based on `NavigationBar.for_user(request.user)`.

PR #2417, opened 2025-04-01
Add `performance_ceiling` in rating computations; superceeds #2410, fixes #2409

### LINKED ISSUE #2409: Add perf_ceiling to ratings

kiritofeng:
This addresses the issue with the current rating system where a new account can join a contest with a rating ceiling (say of 1800) and get a massive amount of rating (i.e. go straight to red).

This PR addresses this in two ways:
1. it introduces a parameter `performance_ceiling`, which defaults to `rating_ceiling + 400` so that the maximum rating you can get from doing a contest with a rating ceiling is hard capped (this was implemented by @xiaowuc1 in #2410).
2. it allows for `performance_ceiling` to be manually set (`performance_ceiling_override`); this is useful should we have to further update the rating system (because if we do so, everyone's ratings can change, and thus potentially invalidate `rating_ceiling`)

Under this new regime, for example, we can retroactively cap the rating gain on those old "junior" contests, without having to set a rating ceiling (since that could render some users unrated, as these contests were made prior to the `rating_ceiling` being implemented).

Note for anyone running a fork of DMOJ: merging this PR will require re-computing ratings for all users.
As doing so can potentially invalidate previously set `rating_ceiling`s, it is recommended to update the lists of users excluded from being rated on contests, based on the previous ratings, before merging this PR.


kiritofeng:
The permissions seem to work correctly:
image
image

judge/user_translations.py
```diff
@@ -7,7 +7,6 @@
     _translations = {}
 
     def translation(language):
-        global _translations
```
int-y1, line 10: is this random deletion safe?

kiritofeng: Lint told me to delete it (and because we don't re-assign `_translations` I think it is).

int-y1:
seems ok, no major bugs.
optionally, you could add a check that `rating_ceiling <= performance_ceiling_override`.

quantum5 (approved):
Nothing looks egregiously wrong. I assume you tested this.

kiritofeng:
Yeah, it's currently running in the test site we call prod. :rofl:

PR #2423, opened 2025-04-06
chore: replace one last django.utils.timezone.utc with datetime.timezone.utc

pxpeterxu:
### Motivation

I was trying to deploy DMOJ to a Ubuntu 24.04 LTS server, and for it to work, I had to upgrade to the latest version of Django. I realized that https://github.com/DMOJ/online-judge/commit/6285f972d5e4849c65ad7f21d6830aa469744743 already fixed most cases of `django.utils.timezone.utc` no longer existing (thanks @kiritofeng!), but there was one instance that was still causing errors.

### Changes

- Replace `contests.py` use of timezone.utc with datetime.timezone.utc too

### Testing

Ran:

```sh
python3 -Wa manage.py test
```

on a server with MySQL, etc. installed as recommended in https://docs.dmoj.ca/#/site/installation

### DIFF
--- a/judge/views/contests.py
+++ b/judge/views/contests.py
@@ -1,4 +1,5 @@
 import json
+import datetime
 from calendar import Calendar, SUNDAY
 from collections import defaultdict, namedtuple
 from datetime import date, datetime, time, timedelta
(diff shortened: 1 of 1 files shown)


kiritofeng:
Thank you for bringing this to our attention!

PR #2428, opened 2025-08-06
Django 5.2 upgrade

alaireselene:
Changes:
- Remove `USE_TZ = True`, this option is [default from Django 5.0](https://docs.djangoproject.com/en/5.2/releases/5.0/#features-removed-in-5-0) 
- Replace `get_joining_columns()` (which [deprecated from Django 5.0](https://docs.djangoproject.com/en/5.2/releases/5.0/#features-deprecated-in-5-0))
- Update `lookup_allowed()` with `request` signature ([deprecated from Django 5.0](https://docs.djangoproject.com/en/5.2/releases/5.0/#features-deprecated-in-5-0))
- Calling hooks instead of calling db directly in `judge/app.py` to prevent failing if DB is not ready
- Manual handle constraints as DMOJ use MariaDB. Old way use syntax that for Portgres only.


alaireselene:
Close as needed to fix more stuff.

PR #2445, opened 2026-03-12
Fix #2444 | .table width 100% -> auto in resources/table.scss

### LINKED ISSUE #2444: Automatically set `table.table` width to `auto` on all problem statements
This is for the markdown tables in problem statements; for some reason we have a rule overriding them to make them the full width at all times which looks kinda weird.

DanPlus6:
I really hope this is the right css file
kirito pls dont annihilate me i have a family and two kids

### DIFF
--- a/resources/table.scss
+++ b/resources/table.scss
@@ -6,7 +6,7 @@
 
 .table {
     border-spacing: 0;
-    width: 100%;
+    width: auto;
     margin-left: auto;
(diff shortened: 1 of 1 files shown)


kiritofeng (changes requested):
Are you sure this only affects the tables in **problem statements** and not **everywhere**.

PR #2449, opened 2026-03-20
Fixed issue 2211

### LINKED ISSUE #2211: User search select2 doesn't round gravatars or color users
image

This doesn't match all other places on the site.

We should also consider coloring the usernames as per rating.

weewoo14:
Added a 4px border-radius to the gravatar images in the search results (keeping it consistent with the rest of the site).
Also (hopefully) added the rating ranks to the user search bar. 

Closes #2211


DanPlus6:
.DS_Store 😭

PR #2452, opened 2026-03-28
Add configurable max output limit for site-managed problems

### LINKED ISSUE #2447: Add limit on output limit for site-managed problems
There should be a limit set on the `output length limit` for site-managed problems, and verified whenever a user tries to save the configuration.

This should be configurable in `settings.py`, and default to 100MB. This is to prevent users from shooting themselves in the foot when they set a **2GB** output length limit and crash a judge.

AdhamAbouaita:
**Feature:** Limits output sizes to prevent judge server crashes.

**Implementation:** Added DMOJ_PROBLEM_MAX_OUTPUT_LIMIT (default 100MB) and applied MaxValueValidator to ProblemData and ProblemTestCase fields.

**Migration:** Added 0153_problem_data_output_limit_validator.py to sync Django’s state. While it generates no SQL, it prevents makemigrations from flagging missing validators in future runs.

Closes #2447


judge/migrations/0153_problem_data_output_limit_validator.py
```diff
@@ -0,0 +1,24 @@
...
+
+    dependencies = [
+        ('judge', '0152_deactivate_user_permission'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='problemdata',
+            name='output_limit',
+            field=models.IntegerField(blank=True, null=True, validators=[django.core.validators.MaxValueValidator(104857600)], verbose_name='output limit length'),
```
quantum5: I love the pretense that this is actually configurable.

quantum5 (changes requested):

judge/migrations/0153_problem_data_output_limit_validator.py
AdhamAbouaita: I removed the migration file entirely since validators are enforced in Python at the form level, not at the database level, so I felt it was a no-op that just hardcoded the limit value without actually doing anything useful.

PR #2453, opened 2026-03-28
Add ability to pin comments for staff announcements

### LINKED ISSUE #2415: Pinned comments for administrator
As comments are sometimes used by staff for important announcements (e.g. procedures for contesting being disqualified on a contest / procedures for requests an unrated due to issues), we should allow for the ability to pin comments.

A pinned comment should:
- Appear before all of its sibling comments on the site
- Will never have its contents hidden (even if it is downvoted)
- Optionally, appear in a separate comment stream from the usual comments.

This would likely require a new flag `is_pinned` on comments, as well as a new permission `judge_can_pin_comment`.

AdhamAbouaita:
Pinned comments appear before all sibling comments and are never hidden by the downvote threshold. Adds a new `judge.pin_comment` permission and a pin/unpin toggle button on the frontend comment UI.

Closes #2415

### DIFF
--- a/judge/admin/comments.py
+++ b/judge/admin/comments.py
@@ -21,13 +21,13 @@ class CommentForm(ModelForm):
 
 class CommentAdmin(VersionAdmin):
     fieldsets = (
-        (None, {'fields': ('author', 'page', 'parent', 'time', 'score', 'hidden')}),
+        (None, {'fields': ('author', 'page', 'parent', 'time', 'score', 'hidden', 'is_pinned')}),
         (_('Content'), {'fields': ('body',)}),
     )
-    list_display = ['author', 'linked_page', 'time', 'score', 'hidden']
+    list_display = ['author', 'linked_page', 'time', 'score', 'hidden', 'is_pinned']
     search_fields = ['author__user__username', 'page', 'body']
-    actions = ['hide_comment', 'unhide_comment']
-    list_filter = ['hidden']
+    actions = ['hide_comment', 'unhide_comment', 'pin_comment', 'unpin_comment']
+    list_filter = ['hidden', 'is_pinned']
     readonly_fields = ['time', 'score']
     actions_on_top = True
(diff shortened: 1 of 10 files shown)


kiritofeng:
Can you provide pictures of the UI?

judge/admin/comments.py
```diff
@@ -51,6 +51,20 @@ def unhide_comment(self, request, queryset):
                                             '%d comments successfully unhidden.',
                                             count) % count)
 
+    @admin.display(description=_('Pin comments'))
```
kiritofeng, line 54: I don't think mass pinning/unpinning would be required, @quantum5 what are your thoughts?

resources/comments.scss
```diff
@@ -179,3 +179,12 @@ a {
         transform: translatez(0);
     }
 }
+
+.pinned-indicator {
```
kiritofeng, line 183: I don't know how I feel about this: can you add screenshots for what a pinned comment looks like?

judge/jinja2/__init__.py
```diff
@@ -19,7 +19,22 @@
...
+    trees.sort(key=lambda c: (not getattr(c, 'is_pinned', False)))
+    _sort_pinned(trees)
+    return trees
+
+
+def _sort_pinned(nodes):
+    for node in nodes:
+        children = getattr(node, '_cached_children', [])
+        if children:
+            children.sort(key=lambda c: (not getattr(c, 'is_pinned', False)))
```
quantum5, line 33: This is what we call a disgusting, unmaintainable hack inside AI slop. There's no sane way to do this and we don't actually need this feature, so we will not be merging this.

quantum5 (changes requested):

judge/jinja2/__init__.py
AdhamAbouaita: Sorry. Thanks for your time.

PR #2456, opened 2026-04-27
Block non-authenticated users from accessing submission page > 1

Ninjaclasher:
This is untested but I believe.


quantum5 (approved):
Doesn't seem like it can really break that badly, yolo.

PR #2460, opened 2026-05-18
Fix re-rating contests when user is DQ'd; #2414

### LINKED ISSUE #2414: Do not rate contests that have not yet been rated when DQ'ing users
The current behaviour is that when a contestant is DQ'd from a contest, all contests that end after said contest, but before the current point in time, are re-rated.

However, this does not account for the fact that some of these contests have not yet been rated themselves.
This is a bug: and the DQ code should only re-rate contests that have previously been rated, not rate new contests.

weewoo14:
Added the parameter "ended=true" to ensure that it only re-rates contests that have ended (previously been rated), and not new ones.

Closes #2414

### DIFF
--- a/judge/models/contest.py
+++ b/judge/models/contest.py
@@ -487,7 +487,7 @@ class Contest(models.Model):
         with transaction.atomic():
             Rating.objects.filter(contest__end_time__range=(self.end_time, self._now)).delete()
             for contest in Contest.objects.filter(
-                is_rated=True, end_time__range=(self.end_time, self._now),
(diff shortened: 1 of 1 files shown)


kiritofeng:
I don't think this is even valid code: `ended` is a `@cached_property`, not a DB field.
This also doesn't fix the bug: `ended` only checks that the end time is before the current time, which is exactly what the current behaviour is. You want to check that the contests being re-rated have not only ended, but also *been rated already*.

PR #2470, opened 2026-08-23
Django 5.2 upgrade

WallE256:
# Upgrade to Django 5.2 LTS

## Summary

Moves the site from Django 4.2 to **Django 5.2 LTS**, crossing the breaking changes of 5.0 and 5.1 along the way. Changes are organized one commit per class of breakage, ordered so each commit passes the full test suite on the stack it targets (the dependency flip lands last).

## What changed, per breaking-change class

**pytz support removed in Django 5.0** (`fix:` commit)
- `judge/timezone.py`: `TimezoneMiddleware` now returns `ZoneInfo` objects instead of pytz ones, with a fallback to `DEFAULT_USER_TIME_ZONE` if a stored profile timezone fails to resolve (defensive; see data note below).
- `judge/models/choices.py`: timezone picker built from `zoneinfo.available_timezones()`. Every name in `pytz.all_timezones` exists in zoneinfo, so stored `Profile.timezone` values remain valid — no data migration needed.

**`Join.get_joining_columns()` fallback deprecated in 5.0, removed in 6.0** (`refactor:` commit)
- `judge/utils/raw_sql.py`: `FakeJoinField` implements `get_joining_fields()` (pairs of field objects) instead of `get_joining_columns()`. Also drops an `external_aliases` isinstance branch whose comment claimed "Django 3.x" — both 4.2 and 5.x use a dict there.

**Removed settings/docs rot** (`chore:` commit)
- settings.py doc links bumped to the 5.2 series. `SILENCED_SYSTEM_CHECKS` entries re-verified against 5.2 check IDs.

**Dependencies incompatible with Django 5** (`build:` commit)

| Package | Before | After | Why |
|---|---|---|---|
| Django | `>=4.2,<5` | `>=5.2,<5.3` | current LTS |
| social-auth-app-django | `==5.0.0` | `>=5.9,<6` | 5.0 predates Django 5 support; 5.9 declares Django 5.2 and includes fixes up to CVE-2025-61783 |
| social-auth-core | `==4.3.0` | `>=4.8.3,<5` | floor required by app-django 5.9; ceiling keeps us on the core 4 API |

social-auth-app-django **6.x is deliberately avoided**: its login views became POST-only, which would require reworking our GET-based social login links, and it forces a core 5.x major bump.

## Upgrade notes

- Run `manage.py migrate social_django` when deploying. This applies `0011`–`0017`: id-field alters, `extra_data` moved to a native JSON column (**the conversion can be slow on large tables**), and a non-empty constraint on `uid`.
- `pytz` is no longer a dependency.

## Testing

- Full suite green on a 4.2 baseline *before* any changes, and on 5.2 after each commit.
- Suite emits **zero** `RemovedInDjango60Warning`/`RemovedInDjango61Warning` under `-W default`.

## Not in this PR (follow-ups)

- Migrating the two git-fork `jsonfield.JSONField` usages to `models.JSONField`.
- social-auth-app-django 6.x (POST-only login views + core 5.x).
- Revisiting old pins (`mistune<2`, `webauthn<1`, `packaging<22`, `importlib-metadata<5`)

### DIFF
--- a/judge/utils/raw_sql.py
+++ b/judge/utils/raw_sql.py
@@ -1,4 +1,5 @@
 from django.db import connections
+from django.db.models import Field
 from django.db.models.sql.constants import INNER, LOUTER
 from django.db.models.sql.datastructures import Join
 
@@ -22,8 +23,13 @@ class FakeJoinField:
         self.joining_columns = joining_columns
         self.related_model = related_model
 
-    def get_joining_columns(self):
-        return self.joining_columns
+    def get_joining_fields(self):
+        def make_field(column):
+            field = Field()
+            field.column = column
+            return field
+
+        return [(make_field(lhs), make_field(rhs)) for lhs, rhs in self.joining_columns]
 
     def get_extra_restriction(self, alias, remote_alias):
         pass
@@ -35,10 +41,7 @@ def join_sql_subquery(
         parent_alias = parent_model._meta.db_table
     else:
         parent_alias = queryset.query.get_initial_alias()
-    if isinstance(queryset.query.external_aliases, dict):  # Django 3.x
-        queryset.query.external_aliases[alias] = True
-    else:
-        queryset.query.external_aliases.add(alias)
+    queryset.query.external_aliases[alias] = True
     join = RawSQLJoin(subquery, params, parent_alias, alias, join_type, FakeJoinField(join_fields, related_model),
                       join_type == LOUTER)
     queryset.query.join(join)
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-Django>=4.2,<5
+Django>=5.2,<5.3
 django_compressor>=3
 django-mptt>=0.13
 django-registration-redux>=2.10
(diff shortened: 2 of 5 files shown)


requirements.txt
```diff
@@ -28,7 +27,8 @@ qrcode[pil]
 jsonfield @ git+https://github.com/DMOJ/jsonfield.git
 pymoss
 packaging<22
-celery
+# Celery 6.0 removes the legacy setting aliases (e.g. CELERY_RESULT_BACKEND) used by dmoj/celery.py.
```
quantum5, line 30: This should probably be a TODO.

judge/utils/raw_sql.py
```diff
@@ -22,8 +23,13 @@ def __init__(self, joining_columns, related_model):
...
 
-    def get_joining_columns(self):
-        return self.joining_columns
+    def get_joining_fields(self):
+        def make_field(column):
+            field = Field()
+            field.column = column
+            return field
+
+        return [(make_field(lhs), make_field(rhs)) for lhs, rhs in self.joining_columns]
```
quantum5, line 32: This requires testing. What kind of testing have you done?

quantum5: Also you might want to make sure there's a fallback path for 4.2 here in case the migration goes south.

